PR #314, opened 2014-08-05
merged hi2 (https://github.com/errge/hi2).

kuribas:
This will use the modification from hi2.  It highlights indentation columns, and binds shift-tab to indent to the left, instead of backspace.  It's probably a good idea to warn users that the keybindings have changed.

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -1,9 +1,11 @@
 ;;; haskell-indentation.el -- indentation module for Haskell Mode
 
-;; Copyright (C) 2009  Kristof Bastiaensen
+;; Copyright (C) 2013  Kristof Bastiaensen, Gergely Risko
 
 ;; Author: Kristof Bastiaensen <kristof.bastiaensen@vleeuwen.org>
-
+;; Author: Gergely Risko <gergely@risko.hu>
+;; Keywords: indentation haskell
+;; URL: https://github.com/haskell/haskell-mode
 ;; This file is not part of GNU Emacs.
 
 ;; This file is free software; you can redistribute it and/or modify
@@ -32,22 +34,14 @@
 
 ;;; Code:
 
+(require 'hl-line)
 (require 'syntax)
 (with-no-warnings (require 'cl))
 
-(defvar delete-active-region)
+(defvar haskell-indentation-dyn-first-position)
+(defvar haskell-indentation-dyn-last-direction)
+(defvar haskell-indentation-dyn-last-indentations)
 
-;; Dynamically scoped variables.
-(defvar following-token)
-(defvar current-token)
-(defvar left-indent)
-(defvar starter-indent)
-(defvar current-indent)
-(defvar layout-indent)
-(defvar parse-line-number)
-(defvar possible-indentations)
-(defvar indentation-point)
-(defvar haskell-literate)
 
 (defgroup haskell-indentation nil
   "Haskell indentation."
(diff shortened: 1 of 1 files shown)


gracjan:
@kuribas: does this merge replace haskell-indentation with hi2?

In case you did not see results:

    In toplevel form:
    haskell-indentation.el:1190:1:Error: the function `hl2-show-overlays' is not known to be defined.
    make: *** [check-haskell-indentation] Error 1

gracjan:
@purcell: this looks like old version of hi2, not relevant anymore.

This particular pull request can be closed but we should think what to do about hi2 as it is the future of indentation.

kuribas:
It's not an old version of hi2, I merged the changes of hi2 with the current haskell-indentation.el.

gracjan:
@kuribas: Great!

I just saw that this pull request is from 5 Aug 2014 while last changes to nilcons/hi2 were done on 5 Oct 2014.

Do you intent to develop hi2 as separate package or as part of haskell-mode?

kuribas:
@gracjan: no, I think it's better to remove the hi2 fork, and develop haskell-indentation.el as part of haskell-mode.

gracjan:
@kuribas: Am I correct that haskell-indentation was an evolution of haskell-indent in the same sense as hi2 was evolution of haskell-indentation? (This predates my involvement here so I lack knowledge of history).

kuribas:
No, haskell-indentation was written from scratch.

gracjan:
@kuribas: is there anything provided by haskell-indent that is missing in haskell-indentation (hi2)?

kuribas:
Yes, the two are totally different modes, with different features and keybindings.

gracjan:
@kuribas: that is unfortunate, because haskell-indent does not seem to be maintained.

kuribas:
I wrote haskell-indentation because I didn't like editing with haskell-indent, so I cannot help you here, sorry.

gracjan:
@kuribas: yeah, nobody likes haskell-indent. I hoped that haskell-indentation subsumes haskell-indent so we can solve the usual 'which haskell indentation mode should I use' in favor of haskel-indentation.

hvr:
@gracjan @kuribas may I remind you that it's not so clear cut which indentation mode is the single most favorite of everone? See

https://docs.google.com/forms/d/1FzWbzGm6odYWxJZcU3GFHlS3lVFTBOI1-M1c87CjOFg/viewanalytics

where all 3 indentation modes were more or less equally popular.

gracjan:
Preference is on `haskell-indentation`: 43 votes, `haskell-indent`: 27 votes.

Why `hi2` answer is not present in the poll? Wasn't `hi2` available back then?

gracjan:
@hvr: also note that I've worked a lot on `haskell-simple-indent` recently so I know how buggy it used to be. Knowing that people prefered `haskell-simple-indent` to something else is a testimony how many issues did they encounter using other indentation modes.

kuribas:
@gracjan I don't think many people know about hi2, that's why I wanted to push the changes to haskell-indentation.  Most people just use it because it comes with haskell-mode I guess. As @hvr mentioned, I wouldn't recommend on removing haskell-indent.  But I'd use haskell-indentation as default.  Of course I am terribly biased :-)  I know that haskell-indentation sometimes fails on syntax-extensions.  That would be an area of improvement.  I'd rather see people sending bug requests than rejecting haskell-indentation completely.

hvr:
Yes, `hi2` came into existence  shortly after that survey...

gracjan:
@kuribas: I guess a rebase over master would be needed to get a clean compile.

purcell:
Came here to close this as suggested, but if I've misunderstood, feel free to reopen.

gracjan:
Yes, a bit too fast :)

purcell:
Now replaced by #483, I see. Thanks @kuribas

kuribas:
No problem.

PR #325, opened 2014-08-23
add support to indent and align guard and rhs for unicode haskell

m00nlight:
Add support to indent and align for unicode haskell program.


gracjan:
@purcell: Is this one mergable at this point?

purcell:
Yes. Squashed and applied in 1d86c85.

purcell:
(Thanks @m00nlight)

PR #357, opened 2014-10-14
Remove an individual redundant name from an import statement

alexanderkjeldaas:
haskell-mode does not currently remove redundant imports that are individually imported.  For example

``haskell
import Foo (a, b, c)
``

Here `a` will not be removed when ghc complains.  This change fixes that.


gracjan:
@alexanderkjeldaas: Would you care to rebase your commits on top of current haskell/master? It should work much better now.

PR #358, opened 2014-10-14
Add hayoo backend for suggesting imports.

alexanderkjeldaas:
NixOS doesn't seem to have a working hoogle, so hayoo is the better option.  Added a hayoo backend.


gracjan:
@purcell: close?

PR #350, opened 2014-10-01
Added nix-shell-ghci support.

alexanderkjeldaas:
A 5 minute hack, 95% untested, but I get my ghci prompt.

PR #370, opened 2014-11-10
Proposal - Add haskell-process-wrapper-function

ardumont:
Hello,

Following this [PR](https://github.com/haskell/haskell-mode/pull/350) and [its possible improvment](https://github.com/haskell/haskell-mode/pull/350#issuecomment-59434038), I added the `haskell-process-wrapper`.

This permits to launch the `haskell-process-type` command (inside `haskell-process-start` function)  inside a wrapper command (`nix-shell` for example, or others).

For example, using this:

``` elisp
(custom-set-variables '(haskell-process-wrapper '("nix-shell" "default.nix" "--command"))
                      '(haskell-process-type 'cabal-repl)))
```

For this example, this triggers the `cabal repl` command inside `nix-shell` (defined within the [default.nix](https://github.com/ardumont/haskell-lab/blob/master/haskell-lab.nix) expression).

I tested manually - https://github.com/ardumont/haskell-mode/commit/3f6e21d4eecaa39faa4053b2df6bbe31013e4087 with:
- `haskell-process-wrapper` (for nix-shell as above) with `haskell-process-type` in `(cabal-repl cabal-ghci ghci)`.
- `haskell-process-wrapper` nil (the default) and `haskell-process-type`
  in `(cabal-repl cabal-ghci ghci)`.

The code is generic, so this should work for the other commands that @chrisdone mentioned.

_Note_
I added unit tests (to enforce that I did not break anything) and I needed mocks (I added the deps inside the `haskell-process-tests.el` file but this must not be what you want. 
If you have any suggestions on how to improve this, I'm all for it.

### DIFF
--- a/haskell-process.el
+++ b/haskell-process.el
@@ -117,9 +117,9 @@ See `haskell-process-do-cabal' for more details."
 
 If no wrapper is needed, then using 'identify function is sufficient.
 Otherwise, define a function which takes a list of arguments.
-For example: (lambda (argv)
-                (append '(\"nix-shell\" \"haskell-lab.nix\" \"--command\")
-                  (shell-quote-argument argv)))")
+For example:
+  (lambda (argv) (append (list \"nix-shell\" \"default.nix\" \"--command\" )
+                    (list (mapconcat 'identity argv \" \"))))")
 
 (defcustom haskell-process-log
   nil


fatlazycat:
Hi, could some one point me at an emacs config to get haskell mode working with nix-shell. I have tried setting haskell-process-wrapper-function but on a haskell-process-load-or-reload it never finds cabal when setting up the project. The project compiles & runs etc if i enter the shell myself with nix-shell --pure etc.

Thanks

ardumont:
@fatlazycat does none of those work?

https://github.com/haskell/haskell-mode/pull/370#issuecomment-64816757

cocreature:
It was broken for me as well until https://github.com/haskell/haskell-mode/commit/35633bfca637e2b9d667fb6a42328cccae9cebe1, maybe that fixed it for you.

ardumont:
> It was broken for me as well until 35633bf, maybe that fixed it for you.

Ok, indeed, without this code, that would not work :D 
Thanks for the information.

PR #372, opened 2014-11-11
Proposal - Add Cask to deal with test dependencies

ardumont:
Following this [PR](https://github.com/haskell/haskell-mode/pull/370), I needed to add some unit test lib (el-mock) to tests that I did not break anything.

As no dependencies seems necessary until now, I just added a temporary HACK to solve the dependency (https://github.com/haskell/haskell-mode/pull/370#discussion-diff-20121622) at the time.
This comes to solve this HACK (and maybe later ease the installation of new deps).

Cask [seemed a proper solution for this](https://github.com/haskell/haskell-mode/pull/370#discussion_r20136791).

Here is a way to make Cask run the tests (and solve tests dependencies).

The only issue now seems that in the current ci-travis, the `emacs-snapshot` you use is an
emacs 23.3 and Cask only supports emacs >= 24... (that's why the ci break on `emacs-snapshot`)

*Note* If you accept the previous PR, you'll see only the 2 latest commits which add Cask.

Cheers,

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -10,6 +10,7 @@
 ;;          2003      Dave Love <fx@gnu.org>
 ;; Keywords: faces files Haskell
 ;; URL: https://github.com/haskell/haskell-mode
+;; Version: 13.07
 
 ;; This file is not part of GNU Emacs.
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ haskell-mode-autoloads.el
 haskell-mode.info
 haskell-mode.tmp.texi
 dir
+/.cask/
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ env:
     - EMACS=emacs23
     - EMACS=emacs24
     - EMACS=emacs-snapshot
+  global:
+    - PATH=$HOME/.cask/bin:$PATH
 
 matrix:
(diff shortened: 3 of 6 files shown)


gracjan:
@ardumont: In case you did not see this fails with:

    cask upgrade-cask: error: Emacs 24 required, yours is Emacs 23.3.1.
    Cask could not be bootstrapped. Try again later, or report an issue at https://github.com/cask/cask/issues
    make: *** [deps] Error 1

Is there a way to fix this?

ardumont:
> @ardumont: In case you did not see this fails with:

Thanks for raising this up.
I will look into it.

If haskell-mode maintainers are interested, I will rebase to the latest code, squash, remove commits, etc..

haskell-mode.el
```diff
@@ -10,6 +10,7 @@
 ;;          2003      Dave Love <fx@gnu.org>
 ;; Keywords: faces files Haskell
 ;; URL: https://github.com/haskell/haskell-mode
+;; Version: 13.07
```
ardumont: I need a version here otherwise cask is not happy.
What should be the right one here?

ardumont:
I'm working on it :D

ardumont:
I think we can also close this one too as this was split in 2:
- #491 
- #492

Cheers,

PR #373, opened 2014-11-11
Proposal - Add EVM (emacs version manager) to deal with emacs versions on travis

ardumont:
Again, it's a proposal (this eases the .travis.yml's readability).

Also, this reduces the use of [Damien Cassou's PPA which is no longer
maintained](https://launchpad.net/~cassou/+archive/ubuntu/emacs).

What are the targeted versions for emacs?

At the moment, it's ok with emacs-24.3 and emacs-24.4 but otherwise, there are limitations:
- emacs-23.4 -> limit from cask not supporting before emacs 24
- emacs 24.1 -> pb in ghc-core.el regarding cl-lib
- emacs 24.2 -> the same

Cheers,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,31 @@
 language: emacs-lisp
 
 env:
+  global:
+    - PATH=$HOME/.cask/bin:$HOME/.evm/bin:$PATH
   matrix:
-    - EMACS=emacs23
-    - EMACS=emacs24
-    - EMACS=emacs-snapshot
+    #   # Cask does not support emacs < 24
+    # - EVM_EMACS=emacs-23.4-bin
+    #   # dependencies break in ghc-core.el about cl-lib
+    # - EVM_EMACS=emacs-24.1-bin
+    #   # dependencies break in ghc-core.el about cl-lib
+    # - EVM_EMACS=emacs-24.2-bin
+    - EVM_EMACS=emacs-24.3-bin
+    - EVM_EMACS=emacs-24.4-bin
+  # allow_failures:
+  #   - EVM_EMACS=emacs-23.4-bin
+  #   - EVM_EMACS=emacs-24.1-bin
+  #   - EVM_EMACS=emacs-24.2-bin
 
-matrix:
-  allow_failures:
-    - env:
-        - EMACS=emacs-snapshot
-
-install:
-  - if [ "$EMACS" = "emacs23" ]; then
-        sudo apt-get update -qq &&
-        sudo apt-get install -qq emacs23-gtk emacs23-el;
-    fi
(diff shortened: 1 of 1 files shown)


gracjan:
@ardumont: I have a bit of trouble following what you did here. There are 5 commits, most of them seem to unrealed to EVM... Can you somehow explain what is going on?

ardumont:
Hello,

> @ardumont: I have a bit of trouble following what you did here. There are 5 commits, most of them seem to unrealed to EVM...

It's difficult to follow because it's now out of context.

>  Can you somehow explain what is going on?

I proposed in order 3 PR with separate concerns (but depending on each other):
- #370 to permit the use of `haskell-process-wrapper-function` (use of nix-shell, docker, ...).
This introduced some unit tests because I wanted to be certain not to break this awesome mode.
But for this, I add to make 1 quick hack to permit the triggering of tests.

This PR has been closed but the main functionality has been merged by @chrisdone (via cherry-picking) because it was on a major refactoring on the side.

I believe the tests have not been integrated.

- other PR #372 to fix the hack that triggered the tests using cask (a better solution than the previous hack).

- this one which permits the use of evm.

-- 

I believe that this can: 
- either be closed if haskell-mode maintainers are not interested by this.
- either you are interested, thus, I need to rebase the code, remove the needless commits and see what's wrong.

Does this make better sense to you?

Cheers,

ardumont:
I remembered right, the tests have not been merged: https://github.com/haskell/haskell-mode/pull/370#issuecomment-64982771

gracjan:
@purcell: I refer to you as my Emacs knowledge is not good enough to see what is needed here.

purcell:
@ardumont Where does `~/.evm` come from? 0f4bd9f adds references to it, but I don't understand how that all comes together.

ardumont:
> @ardumont Where does ~/.evm come from? 0f4bd9f adds references to it, but I don't understand how that all comes together.

It's the installed here https://github.com/haskell/haskell-mode/commit/0f4bd9fdb4a608cc55e0219897978298102b58b3#diff-354f30a63fb0907d4ad57269548329e3R21

(Oh and I need to update this PR if you are still interested by this.)

ardumont:
Just update this PR with now only one commit.
To keep up with what I just said @purcell, this https://github.com/haskell/haskell-mode/pull/373/files#diff-354f30a63fb0907d4ad57269548329e3R21 is what install evm.

ardumont:
@gracjan's PR https://github.com/haskell/haskell-mode/pull/477 seems to do the same with less limitations than this one.

@purcell I think we can close this one, what do you think?

purcell:
Yes, thanks. #477 is now applied.

Issue #187, opened 2013-07-20
Clean up module dependency cycles introduced by `haskell-{process,session,interactive-mode}`


hvr:
The graph diagram below shows the current module inter-dependencies, and is generated by visualizing  `require` (black edges) and `declare-function` (red dashed edges) statements contained in the `*.el` files (i.e. the nodes):

![depgraph](https://f.cloud.github.com/assets/285533/830411/a79849b6-f14e-11e2-8826-da0305cda994.png)

As can be clearly seen the current integration of the `haskell-{process,session,interactive-mode}` cluster currently destroys an otherwise DAG-like dependency graph.

gracjan:
@hvr @chrisdone: is this one done?

gracjan:
I guess it is done. Reopen if you disagree.

PR #381, opened 2014-11-29
Large re-organization to remove cyclic module dependency graph

chrisdone:
Related to #187. The current `master` module hierarchy evolved organically with much sharing between files, making it a mess of a graph (see @hvr's picture).  This pull request finally corrects that. Opening as a pull request because any changes you guys have committed locally will have to be rebased. So I suggest you push your changes first and I will rebase upon them. However, the longer this goes unmerged it will be harder to do so. I'm going to give it a grace period of a week before merging, although if everyone is happy to go ahead I'll merge it sooner.

I wrote a trivial script that would rm all `.elc` files and byte-compile just the file I wanted to test. Using that, I ran it on every file in the project to see that every file had the proper `require` dependencies. Naturally, the HIM modules were all an interloping mess, so I've significantly moved code around so that the hierachy fits. Here is the new hierarchy:

deps

As you can see, at the top is the `haskell` module. Given that the `haskell-mode.el` module contains the mode, it didn't make sense to have everything being pulled into there, especially given that many things depend on `haskell-mode.el` itself. So we end up with three 'big' modules that are depended on a lot: `haskell-mode.el`, `haskell-interactive-mode.el` and `haskell-process.el`. I removed all the specialized commands and things from haskell-session, haskell-process and haskell-interactive-mode into haskell-load or haskell-commands. It makes those modules nice and small (relatively speaking). I would like to do further breaking down of that `haskell.el` file until it's nearly just a list of imports. 

I didn't actually rename any functions, though I would've liked to (using `make-obsolete`), but that would be far more problematic. The point is to move things in the right files. It's not entirely finished, nor am I sold on the particular module naming, but I consider this a big step in the right direction. After this we can incrementally improve.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ ELFILES = \
 	inf-haskell.el
 
 ELCFILES = $(ELFILES:.el=.elc)
-HASKELL = haskell.el
+AUTOLOADS = haskell-mode-autoloads.el
 
 PKG_DIST_FILES = $(ELFILES) logo.svg NEWS haskell-mode.info dir
 PKG_TAR = haskell-mode-$(VERSION).tar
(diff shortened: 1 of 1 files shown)


gracjan:
@chrisdone: Is your module-refac branch still needed or can we remove it?

PR #385, opened 2014-11-30
Use haskell-check-command for flymake-init

### LINKED ISSUE #384: flymake-init uses `nil` haskell-saved-check-command
Using raw flymake in haskell-mode, without additional functionality enabled (e.g. ghc-mod), `flymake-init` uses `haskell-saved-check-command`, which defaults to `nil` leaving flymake broken. It appears to pre-suppose that the use has already run `haskell-check` which will set `haskell-saved-check-command`. But if flymake mode is enabled by default then it will immediately run on finding the file and fail henceforth. 

haskell-mode version 20141127.28 from melpa
emacs 24.4.1 on debian testing.

andrewsw:
Instead of the default nil of haskell-saved-check-command, uses
haskell-check-command to ensure we have something reasonable to run.

Fixes #384

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -708,6 +708,13 @@ is asked to show extra info for the items matching QUERY.."
                  (const "ghc -fno-code")
                  (string :tag "Other command")))
 
+(defcustom haskell-flymake-check-command "hlint"
+  "*Command used by flymake for on-the-fly syntax checking of a Haskell file."
+  :group 'haskell
+  :type '(choice (const "hlint")
+                 (const "ghc -fno-code")
+                 (string :tag "Other command")))
+
 (defcustom haskell-completing-read-function 'ido-completing-read
   "Default function to use for completion."
   :group 'haskell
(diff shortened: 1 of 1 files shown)


gracjan:
@chrisdone ? @purcell ? Can/should this pull by @andrewsw be merged?

purcell:
Not in its current form, I'd say. This change removes `haskell-saved-check-command` from the picture, but that var is still set from elsewhere, and some people will likely be accustomed to running the checker explicitly first in order to set it.

I'd suggest that when running the check manually, the default command be taken from this new `haskell-flymake-check-command` variable, and that `haskell-saved-check-command` should still be used (if set) as the `flymake` command.

gracjan:
@andrewsw: Are you able to change your commit so that it conforms to what @purcell says?

andrewsw-janrain:
@gracjan , yes probably. Sorry, I've been heavily distracted at work and hadn't gotten back to it. I'll try this weekend, but won't object if someone beats me to it.

gracjan:
It seems that this one is stalled. If anybody finds time to push this forward do no hesitate to reopen or create a new pull request.

PR #386, opened 2014-12-01
Highlight error, undefined with font-lock-warning-face.

ikirill:
Both are very special symbols; this is similar to emacs-lisp-mode, for
example.

### DIFF
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -415,6 +415,9 @@ Returns keywords suitable for `font-lock-keywords'."
              (2 haskell-keyword-face nil lax)
              (3 haskell-keyword-face nil lax))
 
+            ;; Highlight 'error' and 'undefined', like 'error' in emacs-lisp-mode
+            ("\\_<\\(error\\|undefined\\)\\_>" 0 'font-lock-warning-face)
+
             ;; Toplevel Declarations.
             ;; Place them *before* generic id-and-op highlighting.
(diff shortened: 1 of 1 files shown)


gracjan:
I am partial to this one. I understand that it is good to highlight keywords that should not appear in production code but then `error` and `undefined` aren't real keywords and also there are plenty of other functions that should not appear in production code like `head` or `fromJust`.

Meanwhile there is also `throw`, `throwIO` and friends that signal local errors in code and those would also be worth highlighting.

ivan-m:
I think `undefined` is worth highlighting as a warning; I'm not so sure about `error` but it would probably be worth it.

I think adding every single partial function to be marked like this might be a bit excessive though.

But if this is meant to be a face for warnings (i.e. "are you *really* sure you want to have this here?") then `throw` and `throwIO` probably don't belong (as they are valid to use within user code, and preferable to using `error`).

gracjan:
We are working on #724, so your simple fix will be obsolete. We need a more principled approach.

themattchan:
Why have these definitions been removed from the master?

gracjan:
I sense you would like to have these definitions back, if so please create a PR and include test cases. Thanks.

mrBliss:
@themattchan you can add the following lines to your config to revert these changes:

```emacs-lisp
(font-lock-add-keywords
  'haskell-mode
  '(("\\_<\\(error\\|undefined\\)\\_>" . 'font-lock-warning-face)))
```

Issue #407, opened 2014-12-29
haskell-mode prompts for a new session every emacs session


mk0x9:
Everytime I restart emacs and fire up haskell REPL haskell-mode guides me through session creation process. Starts a new project, asks then for Cabal dir, for current working directory, for build target.

Is there any way to record session configuration into project directory?

PR #409, opened 2014-12-29
Customizable settings for skipping questions

ikirill:
Creating a new session in a simple project asks many questions for which
the default answers may be absolute fine.  This change adds four
customizable variables, and setting them (in file-local and dir-local
variables) would allow questions to be skipped during session creation.

See #407 

For example (to use file-local variables; dir-local variables would work too; here `t` means "default"):

```
-- Local Variables:
-- haskell-session-ask-new-project: nil
-- haskell-cabal-use-directory: t
-- haskell-session-current-directory: t
-- haskell-session-use-default-target: t
-- End:
```

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -50,6 +50,7 @@
 
 (require 'cl-lib)
 (require 'haskell-utils)
+(require 'haskell-customize)
 
 (defconst haskell-cabal-general-fields
   ;; Extracted with (haskell-cabal-extract-fields-from-doc "general-fields")
@@ -183,11 +184,15 @@
 (defun haskell-cabal-get-dir ()
   "Get the Cabal dir for a new project. Various ways of figuring this out,
    and indeed just prompting the user. Do them all."
-  (let* ((file (haskell-cabal-find-file))
-         (dir (when file (file-name-directory file))))
-    (haskell-utils-read-directory-name
-     (format "Cabal dir%s: " (if file (format " (guessed from %s)" (file-relative-name file)) ""))
-     dir)))
+  (let* ((file (or (haskell-cabal-find-file) (buffer-file-name)))
+         (dir (file-name-directory file)))
+    (cond
+     ((and dir (eq haskell-cabal-use-directory t)) dir)
+     ((stringp haskell-cabal-use-directory) haskell-cabal-use-directory)
+     (t
+      (haskell-utils-read-directory-name
+       (format "Cabal dir%s: " (if file (format " (guessed from %s)" (file-relative-name file)) ""))
+       dir)))))
 
 (defun haskell-cabal-compute-checksum (dir)
   "Compute MD5 checksum of package description file in DIR.
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -22,6 +22,7 @@
 (require 'haskell-font-lock)
 (require 'haskell-interactive-mode)
 (require 'haskell-session)
+(require 'haskell-customize)
 (require 'highlight-uses-mode)
 
 ;;;###autoload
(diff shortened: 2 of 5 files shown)


gracjan:
As I read comments here I see that there is general consensus of not pulling these 4 vars in and rethinking interaction with user again, am I right?

(Do not hesitate to reopen this issue if I misunderstood the intention).

vlatkoB:
I've made an ugly workaround for this, but the proper way, IMHO, would be to adapt ` haskell-process-load-file` function (or create another one) to accept an argument "to ask or not to ask". So we can bind both to different keys, and have everyone happy.
Would that be acceptable?

Workaround:
```elisp
;; Start Haskell Interactive mode without prompts
(defun haskell-session-new-assume-from-cabal-no-prompt ()
  (let ((name (haskell-session-default-name)))
    (unless (haskell-session-lookup name)
      (haskell-session-make name))))
(defun haskell-utils-read-directory-name-no-prompt (prompt default)
  (let ((filename (file-truename default)))
    (concat (replace-regexp-in-string "/$" "" filename) "/")))
(defun haskell-session-target-no-prompt (s)
  (let* ((maybe-target (haskell-session-get s 'target))
    (target (if maybe-target maybe-target (haskell-session-set-target s ""))))
    (if (not (string= target "")) target nil)))

;; Auto-accept default values on interactive mode
(fset 'haskell-session-new-assume-from-cabal 'haskell-session-new-assume-from-cabal-no-prompt)
(fset 'haskell-utils-read-directory-name     'haskell-utils-read-directory-name-no-prompt)
(fset 'haskell-session-target                'haskell-session-target-no-prompt)
```

gracjan:
If with prefix argument (C-u) it should ask. Without prefix it should not
ask.

vlatkoB:
Should behave something like this:
```elisp
(defun haskell-process-load-or-reload-ad (&optional arg)
  "If ARG is set (C-u), starts session with prompts.
If ARG is not set, accepts defaults without prompting user."
  (interactive "P")
  (if arg
      (haskell-process-load-or-reload)
    (progn
      (advice-add 'haskell-session-new-assume-from-cabal
        :override #'haskell-session-new-assume-from-cabal-no-prompt)
      (advice-add 'haskell-utils-read-directory-name
	    :override #'haskell-utils-read-directory-name-no-prompt)
      (advice-add 'haskell-session-target
	    :override #'haskell-session-target-no-prompt)
      (haskell-process-load-or-reload)
      (advice-remove 'haskell-session-new-assume-from-cabal
		            #'haskell-session-new-assume-from-cabal-no-prompt)
      (advice-remove 'haskell-utils-read-directory-name
		            #'haskell-utils-read-directory-name-no-prompt)
      (advice-remove 'haskell-session-target
		            #'haskell-session-target-no-prompt))))
```
This is, of course, just for testing purpose. This depends heavily on the three advised functions above. Those functions are quite scattered and passing an ARG around would not be the best solution. I'm very fresh in emacs, so I do not know if a temporary global would be acceptable.
```elisp
(defvar do-not-ask nil)
(defun haskell-process-load-or-reload-ad ()
  (interactive)
  (ignore-errors
    (progn
      (setq do-not-ask t)
      (haskell-process-load-or-reload)))
  (setq do-not-ask nil))
```
and rewrite those three functions to check for `do-not-ask`.

purcell:
Sorry, no time to comment fully here, but I wanted to point out that the new-style advice functions aren't available in all the Emacs versions we currently support.

vlatkoB:
I made a [solution](https://github.com/vlatkoB/haskell-mode) with a variable `haskell-process-load-or-reload-p` to enable/disable prompts on starting REPL. Default value is `t`, so no changes in current flow. If user changes it to `nil`, no prompts appear and defaults are accepted.

The three above mentioned functions are slightly changed just to check the value of `haskell-process-load-or-reload-p`, and accordingly prompt for a value or accept defaults. No advices.

If you think it's useful, I can make PR.

gracjan:
Yes please.

vlatkoB:
Done.

Issue #410, opened 2014-12-30
⧺(U+29FA) not highlighted as operator


eigengrau:
I’m using `⧺ (double plus, U+29FA)` as an alias to `++`, but I’ve noticed that this glyph is not highlighted like the other operators, despite U+29FA being a non-alphanumeric character as per the Unicode character data.

This is with Haskell-mode v. 20141213.848.

Thanks for making Haskell-mode!

eigengrau:
Hello! Thank you kindly for the fix!

I presume the most general approach would be to use property values defined in the official Unicode character database. I would have expected that Emacs could query these, but I could only find [functions](http://www.gnu.org/software/emacs/manual/html_node/elisp/Character-Properties.html) which can query the Unicode «general category» of a character, not the individual properties.

It might be that this is enough, since the category will say whether a character is a letter or a number (cf. [here](http://www.wikiwand.com/en/Unicode_character_property#/General_Category)). I guess you could either pull a list based on such a category, or define it programmatically based on the category.

Maybe one could just use a regexp character class (`[^[:alpha:][:digit:]]`?), but I previously noticed quirks in the way Emacs handles character classes (e.g., the whitespace class only matches ASCII whitespace).

eigengrau:
Ah, I just noticed you already added some block ranges. I guess that should work. I’m not sure whether the blocks not explicitly mentioned include some operators, though, or vice versa, whether some of these ranges include characters classified as alphanumeric by the Unicode data specs.

eigengrau:
E.g., the «enclosed alphanumerics» block includes characters officially classified as numeric characters. I’m not sure how Haskell syntax is supposed to handle these, though.

purcell:
@eigengrau See the comments in #412 -- specifically:

> I added enclosed alphanumeric characters, they are not alphanumeric (http://www.fileformat.info/info/unicode/char/2460/index.htm)

eigengrau:
Thanks! Some of the chars from the enclosed numbers block are numeric (general category `other number`), but at least with these, the GHC lexer doesn’t like them at all, so it shouldn’t be a problem to define them as operators for the time being. This seems to be a bug in GHC, though, since other `other number` characters, e.g. superscript 2 `²`, are currently valid in identifiers, and the GHC source says `other number` characters are generally okay.

ikirill:
[...]

PR #412, opened 2014-12-31
Syntax table entries for non-alphanumeric chars

ikirill:
See #410 

List of characters generated with the unicode-fonts package
```
(mapc
 (lambda (block)
   (when (and (string-match-p "operators\\|punctuation\\|symbol\\|arrow"
                              (car block))
              (not (string-match-p "alpha\\|letterlike" (car block))))
     (insert
      (format "\n%-45s ; %s"
              (prin1-to-string
               `(modify-syntax-entry (cons ,(nth 1 block) ,(nth 2 block)) "_" table))
              (car block)))))
 unicode-fonts-blocks)
```

I hope I didn't miss any important characters.


purcell:
Great! Merged.

PR #413, opened 2015-01-01
More precise syntax table entries.

ikirill:
The list of characters was generated procedurally based on
get-char-code-property, corresponding to ghc's Lexeme.hs's
implementation of which categories are valid.

```cl
(defun compress-list (list)
  (when list
    (let (sofar start end)
      (setq start (pop list)
            end start)
      (dolist (c list)
        (if (= c (1+ end))
            (incf end)
          (push (cons start end) sofar)
          (setq start c
                end c)))
      (push (cons start end) sofar)
      (setq sofar (mapcar (lambda (x) (if (= (car x) (cdr x)) (car x) x)) sofar)
            sofar (nreverse sofar))
      sofar)))


(let (haskell-sym-list haskell-var-list)
  (dolist (block unicode-fonts-blocks)
    (let (varid
          symid
          (varok '(Lu Lt Lo Ll Nd No))
          (symok '(Pc Pd Po Sm Sc Sk So)))
      (dolist (char (number-sequence (nth 1 block) (nth 2 block)))
        (let ((cat (get-char-code-property char 'general-category)))
          (cond
           ((< char #x80) nil)
           ((memq cat varok) (push char varid))
           ((memq cat symok) (push char symid)))))
      (setq varid (nreverse varid)
            symid (nreverse symid)
            haskell-sym-list (append haskell-sym-list (compress-list symid))
            haskell-var-list (append haskell-var-list (compress-list varid))
            )))
  ;; haskell-sym-list
  (setq haskell-sym-list (sort haskell-sym-list (lambda (a b) (< (or (car-safe a) a) (or (car-safe b) b)))))
  (setq haskell-var-list (sort haskell-var-list (lambda (a b) (< (or (car-safe a) a) (or (car-safe b) b)))))
  haskell-var-list
  )
```


ikirill:
@purcell  It would make sense to split the lists of characters into a separate file, so that haskell-mode.el doesn't get polluted, but I'm not sure if project definition files need updating and what not.

purcell:
Nice. I think this is fine inline for now, and I agree that it's good to make the syntax table as accurate as possible.

PR #419, opened 2015-01-09
Mark some variables as safe.

gracjan:
Mark haskell-indent-offset, haskell-indent-literate-Bird-default-offset,
haskell-indent-rhs-align-column, haskell-indent-look-past-empty-line and
haskell-indent-thenelse as safe file local variables.


gracjan:
@chrisdone ? @purcell ?

purcell:
Seems reasonable.

PR #421, opened 2015-01-09
Allow _ (underscore) in module names

gracjan:


gracjan:
@purcell? @chrisdone?

PR #424, opened 2015-01-16
Add more tests for simple-indent

gracjan:


gracjan:
@purcell ?

This branch introduces more tests. They might be failing as it builds on functionality in two other branches.

purcell:
Which PRs does this depend on?

PR #426, opened 2015-01-16
Refactor haskell-simple-indent.

gracjan:


gracjan:
@purcell 

Added some comments, used a bit more idiomatic elisp.

purcell:
Can I assume these are covered by tests in the other PR?

purcell:
P.S. Thanks for all these fixes.

gracjan:
Yep. There is a test-only PR that covers testing of this one.

PR #427, opened 2015-01-16
Pr haskell info fixes

### LINKED ISSUE #262: info chapter numbers {7,8} swapped
When viewing the haskell-mode info manual in emacs, chapters 7 and 8 have their numbers swapped. In the index node, this is the order of the 6th,7th and 8th entries:

* Indentation::
* Compilation::
* haskell-decl-scan-mode::

But when i navigate to "Compilation", I see "8 Compilation" instead of "7 Compilation" in the title. Likewise i see "7 `haskell-decl-scan-mode'" in the next node's title.
So either the numbers {7,8} are wrong, or the order of the index entries is.

gracjan:
Fixes  #262.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,16 @@ dir: haskell-mode.info
 	$(INSTALL_INFO) --dir=$@ $<
 
 haskell-mode.tmp.texi: haskell-mode.texi
+	@sed -n -e '/@chapter/ s/@code{\(.*\)}/\1/' \
+                -e 's/@chapter \(.*\)$$/* \1::/p' \
+                -e 's/@unnumbered \(.*\)$$/* \1::/p' \
+               haskell-mode.texi > haskell-mode-menu-order.txt
+	@sed -e '1,/@menu/ d' \
+            -e '/end menu/,$$ d' \
+            haskell-mode.texi > haskell-mode-content-order.txt
+	diff -C 1 haskell-mode-menu-order.txt haskell-mode-content-order.txt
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -95,6 +95,16 @@ dir: haskell-mode.info
...
+	@sed -n -e '/@chapter/ s/@code{\(.*\)}/\1/' \
+                -e 's/@chapter \(.*\)$$/* \1::/p' \
+                -e 's/@unnumbered \(.*\)$$/* \1::/p' \
+               haskell-mode.texi > haskell-mode-menu-order.txt
+	@sed -e '1,/@menu/ d' \
+            -e '/end menu/,$$ d' \
+            haskell-mode.texi > haskell-mode-content-order.txt
+	diff -C 1 haskell-mode-menu-order.txt haskell-mode-content-order.txt
+	@rm haskell-mode-menu-order.txt haskell-mode-content-order.txt
+
```
purcell: I may not have had enough coffee yet this morning, but my reading of this code suggests that the two `-order.txt` files are created and then immediately removed, such that the code has no effect. Please correct me if I'm wrong!

gracjan:
This line:

    diff -C 1 haskell-mode-menu-order.txt haskell-mode-content-order.txt

will exit with non-zero status when those differ thus stopping `make` and giving enough info about what is wrong (chapters not matching what was put in menu).

purcell:
Makes sense, thanks.

PR #428, opened 2015-01-16
Generate tags from *.hs, *.lhs and *.hsc only.

### LINKED ISSUE #199: TAGS file points to hs~ files
It looks like emacs backup files with the "~" at the end end up included in TAGS file.

This is very annoying for code navigation. Instead of going to source file i end up going to its backup file. Is there a way to configure TAGS generation to not include emacs backup files?

gracjan:
Added all hasktags supported formats: *.hs, *.lhs and *.hsc.

It used to take also emacs backup files like "*.hs~". This is no longer
the case.

Fixes  #199.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -667,17 +667,16 @@ command from GHCi."
             (if (eq system-type 'windows-nt)
                 (haskell-process-send-string
                  (car state)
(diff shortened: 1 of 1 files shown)


purcell:
It's a nitpick, but could you please use `-print0` for `find` and `-0` for `xargs` so that spaces in filenames are not a problem.

gracjan:
Good idea. Fixed.

PR #429, opened 2015-01-17
Support loading a view buffer into GHCi

acowley:
This creates a temporary file to back a buffer that has no file name so
that it may be loaded into a REPL session. The previous behavior was an
elisp error when trying to save the buffer being loaded.

My use case is this: In org mode, you can use `org-babel-expand-src-block`
to resolve noweb links and give a nice preview of a source block. This
patch lets you load such a block into a REPL session. The expanded view
is a bit tricky as it is read-only, so any changes you make to it are
not synchronized back to the source document. However, the ability to
load a tangled block of code into the REPL is quite useful.

### DIFF
--- a/haskell.el
+++ b/haskell.el
@@ -309,6 +309,11 @@
 (defun haskell-process-load-file ()
   "Load the current buffer file."
   (interactive)
+  (when (not (buffer-file-name))
+    (setq buffer-file-name (make-temp-file "haskell-temp" nil ".hs"))
+    (setq buffer-read-only nil)
+    (set-buffer-modified-p t)
+    (message "haskell-process loading a buffer backed by a temporary file"))
   (save-buffer)
   (haskell-interactive-mode-reset-error (haskell-session))
   (haskell-process-file-loadish (format "load \"%s\"" (replace-regexp-in-string


purcell:
The intention makes sense, but it seems very undesirable to suddenly attach a file to the current buffer, and have to mess around with `buffer-modified-p` etc. From the description in the PR, I expected that the code would create a copy of the buffer in a separate temp file/buffer, and then explicitly tell the haskell process to load that instead of `buffer-file-name`. Is that an approach you tried and encountered problems with?

acowley:
No, that sounds good. I opened the PR in part for just that kind of feedback. Thinking about it, I'll just try to write the file without messing with the buffer so much.

chrisdone:
Interesting. I've been wanting something similar to this for a while.

Creating a uniquely named and buried buffer sounds like a decent solution. You could save in a buffer-local variable the newly created buffer, so that subsequent loads on this nameless buffer will re-use that temp buffer and filename.

Using a temporary file and deleting it after loading is more work due to the asynchronous nature of loading, we'd perhaps have to make `haskell-process-file-loadish` accept a function to run on completion.

acowley:
I had thought that leaving tmp directory cleanup to the OS was okay. You'd rather have the file somewhere less temporary?

chrisdone:
I just meant re-using the same temp file is better than making a bunch of temp files, even if they're all cleaned up once you reboot (which I never do). 

In my second paragraph I was referring to Emacs's functionality like `(with-temp-file )` or something like that, which does the clean-up too. But it's more effort to use that with the haskell process.

gracjan:
@acowley: Did you manage to proceed on this PR?

acowley:
No, I figured maybe I should have a custom load function in my emacs configuration. There are reasonable questions about temporary files above that I don't have a great answer for. I'm fine accumulating temp files over a few months between reboots, but I understand how that might be objectionable.

gracjan:
I think this is stalled. If anybody feels like pushing this forward do not hesitate to reopen this pr or create a new one.

PR #430, opened 2015-01-17
Keep point in haskell-simple-indent.

gracjan:
Keep point in sane positions after haskell-simple-indent. This commit
solves multiple problems:

- messing line content when indent offset was not found
- getting to the biggining of line in couple of cases
- failing to indent second line in the buffer in some cases

Make indentation more predictable to users.

### DIFF
--- a/haskell-simple-indent.el
+++ b/haskell-simple-indent.el
@@ -156,18 +156,28 @@ point beyond the current column, position given by
 (defun haskell-simple-indent-backtab ()
   "Indent backwards.  Dual to `haskell-simple-indent'."
   (interactive)
-  (back-to-indentation)
-  (let ((saved-column (current-column))
+  (let ((saved-column (or (save-excursion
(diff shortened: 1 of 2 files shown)


gracjan:
@purcell, build is fixed.

Question: is there an official minimal version of Emacs that should be supported?

purcell:
> Question: is there an official minimal version of Emacs that should be supported?

README says 23 and up, but we have no Travis build for Emacs < 24, and I think we're all being careful to maintain 23.x compatibility these days. @chrisdone - please correct me if I'm wrong. Perhaps we should start assuming >= 24.1.

gracjan:
Well, my recent patches did not have Emacs 23 in mind.

Without a working Travis build for Emacs 23 it will be hard to keep supporting it in clean conscious.

gracjan:
@purcell? This commit really improves usability!

Issue #432, opened 2015-01-17
Some function types don't seem to be shown by default


ikirill:
Consider the following file; emacs 25.0.50 running with "-Q" and haskell-mode from git repository.

    module A where

    import Data.List
    import Data.Ord

    something :: [(Int, Int)]
    something = sortBy (comparing fst) [(k, -k) | k <- [1..10]]

Start a ghci session, with no special settings. Then placing point on `sortBy` shows its type, but placing point on `comparing` shows nothing, which is disappointing. 

I tried to set `haskell-doc-use-inf-haskell` (which I found in `haskell-doc-sym-doc` to t, but it did something rather strange to the ghci session, probably because I had enabled `interactive-haskell-mode` instead of `inferior-haskell-mode`.

Maybe I'm missing some configuration or something (I don't think so), but why doesn't haskell-mode (with interactive-haskell-mode) display types of fairly common functions, by default? How do I get it to do it?

P.S. Would it be interesting if haskell-mode could show the type of a selected region? That is, if you selected "comparing fst", it would show you the type `comparing fst :: Ord a => (a, b) -> (a, b) -> Ordering`.

I have patches, but is this actually a bug or am I misunderstanding something? I can't tell from documentation.

gracjan:
@ikirill: I tried to reproduce the problem and what I found is:

- ghci session by itself does not show types unless you do `C-c C-t`
- what shows types is `haskell-doc-mode`
- my version of `haskell-doc-mode` shows types in minibuffer both for `sortBy` and for `comparing`

Can you check out newest git version?

gracjan:
@ikirill: Did your patch fix everything? I see a lot of edits on it... not sure if this is done or not yet...?

ikirill:
@gracjan It was some time ago, but I think this was fixed in #433.

PR #433, opened 2015-01-18
Haskell eldoc support with ghci in haskell-interaction-mode.

ikirill:
See #432 for what this is supposed to do. I'm not sure if I missed some corner cases, or if I broke something that I don't use myself. Is `haskell-process` the right function to check if there's a running process available?

I don't want to make synchronous calls to ghci, so the way it's written now it only works for eldoc because other functions that get type information seem to be synchronous.

Why is all this in haskell-doc anyway? Isn't a lot of it duplicating eldoc-mode?

Some of the commented-out code in haskell-doc.el and todo-items are from 2001, btw :)

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -336,15 +336,24 @@ If PROMPT-VALUE is non-nil, request identifier via mini-buffer."
   (interactive "P")
   (if insert-value
       (haskell-process-insert-type)
-    (haskell-process-do-simple-echo
-     (let ((ident (haskell-ident-at-point)))
-       ;; TODO: Generalize all these `string-match' of ident calls into
-       ;; one function.
-       (format (if (string-match "^[_[:lower:][:upper:]]" ident)
-                   ":type %s"
-                 ":type (%s)")
-               ident))
-     'haskell-mode)))
+    (let* ((expr
+            (if (use-region-p)
+                (buffer-substring-no-properties (region-beginning) (region-end))
+              (haskell-ident-at-point)))
+           (expr-okay (and expr
+                         (not (string-match-p "\\`[[:space:]]*\\'" expr))
+                         (not (string-match-p "\n" expr)))))
+      ;; No newlines in expressions, and surround with parens if it
+      ;; might be a slice expression
+      (when expr-okay
+        (haskell-process-do-simple-echo
+         (format
+          (if (or (string-match-p "\\`(" expr)
+                 (string-match-p "\\`[_[:alpha:]]" expr))
+              ":type %s"
+            ":type (%s)")
+          expr)
+         'haskell-mode)))))
 
 ;;;###autoload
 (defun haskell-mode-jump-to-def-or-tag (&optional next-p)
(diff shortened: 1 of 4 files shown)


ikirill:
Okay, I can change this avoid lexical-binding.

But is this at all interesting for haskell-mode? There's a lot of different stuff in haskell-mode, so I can never really tell whether something is missing by design, missing because it's not implemented yet, or it's there but I don't know about it.

purcell:
> But is this at all interesting for haskell-mode? There's a lot of different stuff in haskell-mode, so I can never really tell whether something is missing by design, missing because it's not implemented yet, or it's there but I don't know about it. For example, does this PR reimplement an existing feature, or not?

Answering this falls in @chrisdone's corner, I think. :-)

chrisdone:
haskell-doc predates my involvement with haskell-mode and I don't think anyone's touched it for a long time.  I don't have time to review the patch in detail. Is it just showing the type or info of something at point?

ikirill:
> Is it just showing the type or info of something at point?

Pretty much. I use eldoc-mode, which is a built-in mode that shows you some information about the thing at point in the minibuffer (I'm not sure if haskell-doc duplicates this functionality or not). Then haskell-mode specifies a function that eldoc-mode should call to get the message to be displayed (https://github.com/ikirill/haskell-mode/blob/2c289a77b82812d7c9d29285ab8ab04610bd5899/haskell-doc.el#L1524). It used to be that, with haskell interactive mode, that function would not check the ghci interactive session for possible type information and give a message only if the symbol at point is in its index or a precomputed index (which is nice for Prelude functions and haskell syntax).

This patch adds another function that issues a command to interactive haskell session with ":type", and shows the output to the user.

To test it, enable eldoc-mode, maybe disable haskell-doc-mode (I'm not too sure if they interact), and place point on a symbol like "comparing" from Data.Ord. Ordinarily haskell-mode wouldn't show its type (see the linked issue I submitted), but with this patch it does, by checking ghci.

You can also select a region like "comparing fst", and the type of "comparing fst" should be shown in the minibuffer. If eldoc isn't doing this, try invoking haskell-doc-current-info, I think eldoc doesn't trigger when you select a new region.

Doesn't work on regions containing newlines because I couldn't find out how to pass them to ghci's :type.

chrisdone:
What if you use `haskell-session-maybe` and then `haskell-session-process` if there is a session?

ikirill:
Right, thank you. I see what I missed, that should be fixed now.

ikirill:
@purcell Is lexical-let okay?

purcell:
> Is lexical-let okay?

Yes, but you'd have to suppress the byte-compilation warnings about `'cl`, otherwise the build will fail. We recently reworked the code to remove `cl`, so if you can avoid using `lexical-let` that would be preferable.

ikirill:
I used `(eval-when-compile (require 'cl))`, so there's no cl at runtime (and lexical-let is a macro, so expanded at compile-time). I don't see the build warning: https://travis-ci.org/haskell/haskell-mode/builds/47850113. Isn't the warning for *runtime* uses?

purcell:
Yes, I think you're right. I think it warns at runtime about runtime usage.

gracjan:
@purcell: is this good to merge?

purcell:
Yes, let's go with this and see how we get on. We should certainly prefer using the `eldoc` machinery to a hand-rolled equivalent. Thanks @ikirill -- sorry for the slow handling of this.

PR #435, opened 2015-01-18
Use original shnippet in place of haskell-yas.

### LINKED ISSUE #361: haskell-yas.el should be a separate package
As per https://github.com/haskell/haskell-mode/pull/356#issuecomment-58934768. The bugtracker is usually the best place to track TODOs like this.

gracjan:
Fixes #361.

Original 'shippet' is well maintained at https://github.com/LukeHoersten/shnippet. Someday it should be added as a git submodule. For now users are advised to refer to the original version directly.

Note that haskell-yas was a simple rename of shnippet and does not offer any additional functionality.


purcell:
Have changes been made to the snippets following the import and renaming? `shnippet` hasn't been touched for a year.

I like the idea of keeping this out of `haskell-mode`, but until `shnippet` can be installed via a package, I'd be wary of removing it from the tree.

purcell:
It's not completely unheard of for major modes to bundle yasnippets, even without declaring a package dependency on `yasnippet`, though, so if the other maintainers would prefer to keep this here then I don't feel strongly either way.

gracjan:
Imported snippets have not been changed since import. There also have been zero issue reports about snippets. Both strongly suggest that users do not care about shnippets inside haskell-mode.

Packaging shnippet properly is a very good idea that should be brought up in original repo: https://github.com/LukeHoersten/shnippet.

gracjan:
As per #305 haskell-mode snippets are *activelly maintained* here: https://github.com/AndreaCrotti/yasnippet-snippets, haskell snippets are in subdirectory:

https://github.com/AndreaCrotti/yasnippet-snippets/tree/master/haskell-mode

purcell:
That looks like a user fork with personal snippets. Not sure if it's usable. Best option is still to either keep the status quo, or package `shnippet` then merge this PR.

LukeHoersten:
I still actively maintain `shnippet` - I'm not sure about AndreaCrotti fork since I've received no PRs. We should still come up with a better way to package shnippets into haskell-mode. I like the subrepo idea.

LukeHoersten:
Oops ment to open the issue not the PR.

PR #442, opened 2015-01-21
Operator `M.v` font lock.

### LINKED ISSUE #438: Identifiers within backticks not highlighted when prefixed with module name
While functions enclosed within backticks are normally highlighted using the operator face by default, I have noticed that this doesn’t happen when the function has a module prefix in front. Is this intended?

E. g., in the following program, `Module.function` will have `haskell-default-face`, wheras it has `haskell-operator-face` when the module prefix is left out.

```
x = y `Module.function` z
```

This is with Melpa ver. 20150101.1151.

Best regards!

gracjan:
Font lock for `M.v` should be operator face same as `v`.

Fixes #438.

### DIFF
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -429,9 +429,12 @@ Returns keywords suitable for `font-lock-keywords'."
             ("(\\(,*\\|->\\))" 0 haskell-constructor-face)
             ("\\[\\]" 0 haskell-constructor-face)
             ;; Expensive.
+            (,(concat "`" varid "`") 0 haskell-operator-face)
+            (,(concat "`" conid "`") 0 haskell-operator-face)
+            (,(concat "`" qvarid "`") 0 haskell-operator-face)
+            (,(concat "`" qconid "`") 0 haskell-operator-face)
             (,qvarid 0 haskell-default-face)
             (,qconid 0 haskell-constructor-face)
-            (,(concat "\`" varid "\`") 0 haskell-operator-face)
             ;; Expensive.
             (,conid 0 haskell-constructor-face)
 


purcell:
Not sure why the line order changed here. Also `qvarid` appears to be "qualified varid", so you should probably use that instead of the hand-rolled `[[:alnum:]]` etc.

gracjan:
Well, font-lock syntax for haskell is in a need for refactoring. To answer your question:

qvarid is only module prefixed version and does not apply to both. What we need is a qualified-or-plain-varid. This is not defined so that is why I went with just inline regex.

Order of rules needed to be changed because otherwise a line later that matches conid and qconid would match `conid` and `qconid`. So my line had to be before.

Leftovers like ;; Expensive or Emacs 21 support needs to be cleared up.

Note that the line

    (,reservedsym 1 haskell-operator-face)

appears three times for no reason. That is the best testimony to the quality of this fragment of the code.

Does anybody have pointers to an example how to write unit tests for font-lock or have some more interactive way of working with syntax definitions?

purcell:
> That is the best testimony to the quality of this fragment of the code.

:-)

> Does anybody have pointers to an example how to write unit tests for font-lock or have some more interactive way of working with syntax definitions?

There's https://github.com/Lindydancer/font-lock-studio, which might help.

gracjan:
Is this merged?

purcell:
> Is this merged?

Not yet, no. Isn't there a way to combine `qvarid` and `varid` instead of repeating the `[[:alnum:]]` incantation? Something like `(concat "\\(?:" varid "\\|" qvarid "\\)")` would perhaps do the trick.

gracjan:
Sadly `modid` is defined using normal group not the shy group, so it does not compose:

    (modid (concat "\\b" conid "\\(\\." conid "\\)*\\b"))

So using `qvarid` and `qconid` might bring trouble in places. So I went with the obvious four-liner that you see in the commit (the new one).

gracjan:
@purcell: what about this?

PR #443, opened 2015-01-21
Allow module name completion in import statements

ikirill:
I noticed that haskell-mode wasn't doing this, because ghci needs to be passed the "import " string to make it complete the module name after.

I also changed the behaviour so that it would not forcefully start a new process if there isn't one already running. I think this is standard behaviour in other modes, and it can be intrusive because it is the first thing that happens when I open a new file and it asks several questions about the new session before I can do anything else.


chrisdone:
Nice!

PR #444, opened 2015-01-22
OPTIONS and LANGUAGE completion using ghci

ikirill:
Similar to a previous PR, this adds completion for LANGUAGE and OPTIONS ghc pragmas inside comments, using ghci's `:complete repl ":set ..."`, which understands all language/option flags.

It's not spectacularly useful because normally (?) you'd need to type the first three (?) characters yourself before completion kicks in.

I also made a small change to some previous code to avoid some unnecessary string -> symbol -> string conversion.

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -333,6 +333,18 @@ when Data.Map is the candidate.
   :group 'shm
   :type '(repeat 'string))
 
+(defcustom haskell-ghc-supported-languages
+  (split-string (shell-command-to-string "ghc --supported-extensions"))
+  "List of language pragmas supported by the installed version of GHC."
+  :group 'haskell
+  :type '(repeat string))
+
+(defcustom haskell-ghc-supported-options
+  (split-string (shell-command-to-string "ghc --show-options"))
+  "List of options supported by the installed version of GHC."
+  :group 'haskell
+  :type '(repeat string))
+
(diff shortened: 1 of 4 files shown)


hvr:
The allowed `LANGUAGE` and `OPTIONS` values are static information  (and can hence be cached index by `ghc`-version). 

IIRC, those lists already collected elsewhere at startup by quering `ghc --supported-languages` and `ghc --show-options` respectively for use by Haskell snippets (which is an alternative, and in some way more efficient way to input and complete `LANGUAGE`/`OPTIONS_GHC` pragmas IMHO).

ikirill:
I see, I found it: it's in haskell-yas, and you have to type "lang" by itself, not inside a comment, and it will expand to a LANGUAGE pragma inside a proper comment and offer completions. I missed that, I didn't look in the yasnippet menu carefully. 

I can't find an OPTIONS snippet, I think it's not there.

I would still like for ordinary non-snippet completion to work for these too.

I changed it to precompute the list of completions properly.

gracjan:
I vote to have @ikirill s version of completion instead of the yas snippet. See #435 for reference.

ikirill:
@gracjan I didn't remove the snippet, I added completion but it doesn't replace the snippet.

hvr:
@gracjan those two facilities are not mutually exclusive, I just wanted to hint at that there's already infrastructure that could be shared

LukeHoersten:
I just committed a change renaming `haskell-ghc-supported-languages` to `haskell-ghc-supported-extensions` to make it a bit more clear. The config variable is renamed and redocumented as well.

PR #445, opened 2015-01-22
Resurrect Emacs23 support

gracjan:


gracjan:
Ugh, I give up. I find no way to run ERT under Emacs23 and that rules out support for Emacs23.

gracjan:
Well, at least ERT works now. But tests do not look good at all.

gracjan:
It wasn't that bad after all. I'll clean this up a little and then we can see if we want emacs23 or not.

purcell:
Nice work! Ping me when you consider this ready to merge.

gracjan:
Clean to merge!

purcell:
Thanks!

PR #451, opened 2015-01-24
Pr font lock 1

gracjan:
Some rather obvious improvements to haskell-font-lock.


purcell:
defvar-local is Emacs 24.3 and greater only. Please combine `defvar` and `make-variable-buffer-local` instead.

gracjan:
True. Fixed.

PR #452, opened 2015-01-24
Cleanup following #449

Fanael:


purcell:
:+1:

gracjan:
Nice. I missed haskell-compat.el.

PR #453, opened 2015-01-27
Differentiate Haddock and regular comments better

karljs:
Because this regular expression accepted many spaces it would lead to the problem that in the following example, the first line would be highlighted as a regular comment and the second as a Haddock comment.

    -- fn (Pattern x y) | fn2 x = True
    --                  | otherwise = fn y

Intuitively I think one space or tab should be fine for Haddock comments to keep working, but I would appreciate other people thinking about that independently before merging this.


waern:
Yes, (double dash) Haddock comments must have the prefix "-- ", so one space is allowed only (and no tabs).
See: https://github.com/ghc/ghc/blob/master/compiler/parser/Lexer.x#L358

gracjan:
@karljs: can you remove the \t as it is not legal there either. Thanks!

gracjan:
@waern: thanks.

Summary:

For double-dash: double-dash plus exactly one space plus $docsym
For brace-dash: brace-dash plus optional one space plus $docsym
And docsym = [\| \^ \* \$]

is this correct?

waern:
Correct.

karljs:
Sorry, I pushed too early and don't want to squash a bunch of nonconsecutive commits.  I'll reopen a clean pull request for this as soon as possible.

PR #455, opened 2015-01-31
haskell-cabal indent-relative on empty

tonyday567:
When on an empty line in a .cabal file, tab does an indent-relative (which was the previous behaviour).  Currently, it does nothing, so you have to hit space etc.


gracjan:
Looks like improvement. Thanks!

PR #456, opened 2015-02-03
Use haskell-process-compute-process-log-and-command in haskell-process-start

cocreature:
If you oppose to the indentation changes, I'll take those out. I just thought those lines were way too long. Also I'm a really bad lisper, so I hope I didn't break anything.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -47,58 +47,14 @@
     (haskell-process-set-session process session)
     (haskell-process-set-cmd process nil)
     (haskell-process-set (haskell-session-process session) 'is-restarting nil)
-    (let ((default-directory (haskell-session-cabal-dir session)))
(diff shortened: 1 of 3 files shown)


chrisdone:
Can you explain the point of this so when I get time to merge it I'll know why? Cheers

cocreature:
During refactoring somebody removed the process-log-and-compute function call which resulted in haskell-process-wrapper-function no longer being used. This commit uses the still existing process-log-and-compute function again so that haskell-process-wrapper-function is actually used.

chrisdone:
Oh, `haskell-process-compute-process-log-and-command`. I see. I'll test it in a bit. Thanks

cocreature:
Anything wrong with it or did you just not have time to look at it?

chrisdone:
Looks okay. Tested it, didn't break anything. Merged.

Issue #315, opened 2014-08-06
Allow search for previous expressions in interactive-mode


rpglover64:
Basically, I would like `C-r` in haskell-interactive-mode to behave like `C-r` in ghci.

This is particuarly relevant after a `C-c C-k`.

TikhonJelvis:
I would love a reverse command search bound to `M-r` like comint and the old inferior haskell mode.

gracjan:
@rpglover64: Common sentiment is that the GHCi buffer should build on comint mode, effectively merging functionality of interactive and inferior modes. There is a lot to get done before this happens. Can you help with this?

rpglover64:
@gracjan Probably not. :frowning: I'm not very knowledgeable in elisp and I have a project for work taking all my time.

gracjan:
In that case we need to wait for a volunteer.

PR #457, opened 2015-02-04
Add haskell-interactive-copy-to-prompt.

cmears:
This change lets you press Enter on a line in a haskell-interactive session to copy that line to the current prompt. It is a small change to make the lack of history-search (e.g. #315) slightly less painful.

In #315 Ivan suggested: "If nothing else, inf-haskell let you C-r to do a backward search in the buffer, and then hitting Enter would copy that line to the current prompt line." That's essentially what I've done.

I'm only an inexpert Emacs Lisper, so any corrections or improvements are welcome!

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -178,6 +178,21 @@ be nil.")
       (switch-to-buffer-other-window haskell-interactive-previous-buffer)
     (message "No previous buffer.")))
 
+(defun haskell-interactive-copy-to-prompt ()
+  "Copy the current line to the prompt, overwriting the current
+prompt."
+  (interactive)
+  (let ((l (buffer-substring-no-properties (line-beginning-position)
+                                           (line-end-position))))
+    ;; If it looks like the prompt is at the start of the line, chop
+    ;; it off.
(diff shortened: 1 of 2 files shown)


gracjan:
@purcell? @chrisdone? @hvr? Any opinion about this one?

purcell:
Looks reasonable to me.

gracjan:
I do not know what to think about this one. Is there a prior art with this kind of RET behavior? I've never seen anything working like this.

purcell:
@gracjan Guess the best solution would be proper history search. Without that, a command like this is kinda reasonable, but perhaps not bound to RET.

ivan-m:
The old inferior haskell setup has something like this, and I for one have missed it (I've ended up copy/pasting lines that are too far back in the history to keep traversing through).

gracjan:
I'm not qualified enough to decide on this one. Steve, if you want to click
the merge button I have nothing against (but nothing for this either).

hvr:
@gracjan Is this now behaving similar to `M-x ielm`?

gracjan:
[ielm](https://www.emacswiki.org/emacs/InferiorEmacsLispMode)

Issue #317, opened 2014-08-09
`interactive-haskell-mode` should not bind C-c c


ivarru:
This now calls `haskell-process-cabal`, but the elisp manual clearly states (in section D.2):
- Don't define `C-c LETTER' as a key in Lisp programs.  Sequences
   consisting of`C-c' and a letter (either upper or lower case) are
   reserved for users; they are the _only_ sequences reserved for
   users, so do not block them.

It seems this was introduced in a0dd7405627975d99e7b8d35811377dd713181ad.

PR #461, opened 2015-02-08
Follow Emacs key binding convention

iquiw:
Not to use C-c letter bindings.

The 2nd commit is just cosmetic change, not essential.


hvr:
c.f. https://github.com/haskell/haskell-mode/issues/317#issuecomment-51699063

gracjan:
@purcell: Referenced #317 is closed but this is not. What does this mean?

PR #465, opened 2015-02-12
Fix #459: 'find-tag-marker-ring' is an obsolete variable (as of 25.1)

### LINKED ISSUE #459: 'find-tag-marker-ring' is an obsolete variable (as of 25.1)
Hi guys!

Once Haskell mode was compiling, I got error
> 'find-tag-marker-ring' is an obsolete variable (as of 25.1); use 'xref-push-marker-stack' or 'xref-pop-marker-stack' instead.

So, I prepared the following patches to fix the issue:
```diff
--- haskell-commands.el	2015-02-06 21:26:36.000000000 +0300
+++ haskell-commands-patched.el	2015-02-06 21:56:57.000000000 +0300
@@ -18,6 +18,7 @@
 ;;; Code:
 
 (require 'cl-lib)
+(require 'xref)
 (require 'haskell-process)
 (require 'haskell-font-lock)
 (require 'haskell-interactive-mode)
@@ -355,9 +356,9 @@
 it will always be accurate, in contrast to tags, which always
 work but are not always accurate.
 
-If the definition or tag is found, the location from which you
-jumped will be pushed onto `find-tag-marker-ring', so you can
-return to that position with `pop-tag-mark'."
+If the definition or tag is found, the location from which you jumped
+will be pushed onto `xref--marker-ring', so you can return to that
+position with `xref-pop-marker-stack'."
   (interactive "P")
   (let ((initial-loc (point-marker))
         (loc (haskell-mode-find-def (haskell-ident-at-point))))
@@ -365,8 +366,11 @@
         (haskell-mode-handle-generic-loc loc)
       (call-interactively 'haskell-mode-tag-find))
     (unless (equal initial-loc (point-marker))
-      ;; Store position for return with `pop-tag-mark'
-      (ring-insert find-tag-marker-ring initial-loc))))
+      (save-excursion
+        (goto-char initial-loc)
+        (set-mark-command nil)
+        ;; Store position for return with `xref-pop-marker-stack'
+        (xref-push-marker-stack)))))
 
 ;;;###autoload
 (defun haskell-mode-goto-loc ()
@@ -380,7 +384,7 @@
 (defun haskell-mode-goto-span (span)
   "Jump to the span, whatever file and line and column it needs
 to to get there."
-  (ring-insert find-tag-marker-ring (point-marker))
+  (xref-push-marker-stack)
   (find-file (expand-file-name (plist-get span :path)
                                (haskell-session-cabal-dir (haskell-interactive-session))))
   (goto-char (point-min))
```

```diff
--- inf-haskell.el	2015-02-06 22:12:08.000000000 +0300
+++ inf-haskell-patched.el	2015-02-06 22:16:06.000000000 +0300
@@ -33,6 +33,7 @@
 (require 'comint)
 (require 'shell)             ; For directory tracking.
 (require 'compile)
+(require 'xref)
 (require 'haskell-mode)
 (require 'haskell-decl-scan)
 (require 'haskell-cabal)
@@ -568,7 +569,7 @@
             (setq file (expand-file-name file)))
           ;; Push current location marker on the ring used by `find-tag'
           (require 'etags)
-          (ring-insert find-tag-marker-ring (point-marker))
+          (xref-push-marker-stack)
           (pop-to-buffer (find-file-noselect file))
           (when line
             (goto-char (point-min))
```
These patches should work in principle. Maybe you will find a more suitable place for `(require 'xref)`.

Cheers.

z0rch:
I have tried @vonavi's patch, proposed in #459, and it does seem to fix the described problem.

For your convenience I'm converting the patch into PR, with a hope that it can be reviewed and merged faster.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -18,6 +18,7 @@
 ;;; Code:
 
 (require 'cl-lib)
+(require 'haskell-compat)
 (require 'xref)
 (require 'haskell-process)
 (require 'haskell-font-lock)
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -18,6 +18,7 @@
 ;;; Code:
 
 (require 'cl-lib)
+(require 'xref)
 (require 'haskell-process)
 (require 'haskell-font-lock)
 (require 'haskell-interactive-mode)
@@ -354,10 +355,9 @@ jump to the tag.
 Remember: If GHCi is busy doing something, this will delay, but
 it will always be accurate, in contrast to tags, which always
 work but are not always accurate.
-
-If the definition or tag is found, the location from which you
-jumped will be pushed onto `find-tag-marker-ring', so you can
-return to that position with `pop-tag-mark'."
+If the definition or tag is found, the location from which you jumped
+will be pushed onto `xref--marker-ring', so you can return to that
+position with `xref-pop-marker-stack'."
   (interactive "P")
   (let ((initial-loc (point-marker))
         (loc (haskell-mode-find-def (haskell-ident-at-point))))
@@ -365,8 +365,11 @@ return to that position with `pop-tag-mark'."
         (haskell-mode-handle-generic-loc loc)
       (call-interactively 'haskell-mode-tag-find))
     (unless (equal initial-loc (point-marker))
-      ;; Store position for return with `pop-tag-mark'
-      (ring-insert find-tag-marker-ring initial-loc))))
+      (save-excursion
+        (goto-char initial-loc)
+        (set-mark-command nil)
+        ;; Store position for return with `xref-pop-marker-stack'
+        (xref-push-marker-stack)))))
 
 ;;;###autoload
 (defun haskell-mode-goto-loc ()
@@ -380,7 +383,7 @@ command from GHCi."
 (defun haskell-mode-goto-span (span)
   "Jump to the span, whatever file and line and column it needs
 to to get there."
-  (ring-insert find-tag-marker-ring (point-marker))
+  (xref-push-marker-stack)
   (find-file (expand-file-name (plist-get span :path)
                                (haskell-session-cabal-dir (haskell-interactive-session))))
   (goto-char (point-min))
--- a/haskell-compat.el
+++ b/haskell-compat.el
@@ -21,6 +21,8 @@
 ;;; Commentary:
 
 ;;; Code:
+(require 'xref nil 'noerror)
+(require 'ring)
 
 ;; Missing in Emacs23, stolen from Emacs24's `subr.el'
 (unless (fboundp 'process-live-p)
@@ -33,14 +35,12 @@ A process is considered alive if its status is `run', `open',
 
 ;; Cross-referencing commands have been replaced since Emacs 25.1.
 ;; These aliases are required to provide backward compatibility.
-(unless (require 'xref nil 'noerror)
+(unless (fboundp 'xref-push-marker-stack)
   (defalias 'xref-pop-marker-stack 'pop-tag-mark)
 
   (defun xref-push-marker-stack ()
     "Add point to the marker stack."
-    (require 'ring)
     (defvar find-tag-marker-ring)
-
     (ring-insert find-tag-marker-ring (point-marker)))
 
   (provide 'xref))
--- a/haskell-compat.el
+++ b/haskell-compat.el
@@ -31,6 +31,20 @@ A process is considered alive if its status is `run', `open',
     (memq (process-status process)
           '(run open listen connect stop))))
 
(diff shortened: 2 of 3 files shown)


gracjan:
@z0rch: In case you missed the build failure:

    In toplevel form:
    ghci-script-mode.el:20:1:Error: Cannot open load file: no such file or directory, xref
    make: *** [ghci-script-mode.elc] Error 1

z0rch:
I'll look into it later this week.

z0rch:
@gracjan, I can see that `xref.el` [was introduced on 25 Dec 2014](https://lists.gnu.org/archive/html/emacs-diffs/2014-12/msg00434.html), so this change applies only for the newest Emacs versions.

Since you support Emacs back to v23, you clearly don't want to merge this patch right away.

Do you have any examples in mind, how are you dealing with such situations? I'm sure it is not the first breaking change in Emacs that you see. Then I could improve this patch according to the pattern you use for this project.

gracjan:
@z0rch: we are starting to take backward compatibility seriously. As of now there is `haskell-compat.el` that is a general bin for compatibility. We also ship our version of cl-lib (used for tests only for now when no other cl-lib is available).

`xref` stuff seems to be super-new in Emacs. Isn't there a way to accomplish same funcitonality using somewhat more stable method? Does `xref` provide a migration guide?

z0rch:
> `xref` stuff seems to be super-new in Emacs. Isn't there a way to accomplish same functionality using somewhat more stable method? Does `xref` provide a migration guide?

@gracjan, unfortunately my knowledge of elisp is pretty limited, I can't tell if a third "better" option exist.

However based on what you described, I believe the right approach is to use new functions, but provide aliases for them in `haskell-compat.el` that replicate the old behavior. This will ensure that newer Emacs versions use the most recent API directly and don't require any wrappers, and at the same time backwards compatibility is preserved for older Emacs versions.

Especially since in this particular case, only two aliases are needed: `xref-pop-marker-stack` and `xref-push-marker-stack`.

Would you agree?

----

Now to a more technical level, I'm having issues with aliasing the latter function. Could you please have a look at `haskell-compat.el` file?

If I just put the call to `ring-insert` function, I get a Travis error: reference to free variable `find-tag-marker-ring`. This is the first bit that I don't understand, i.e. why it happens and why it was not an issue before (it was used in both `haskell-commands.el` and `inf-haskell.el`). <br>I added the dummy `defvar` as suggested [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Warning-Tips.html), but still looks suspicious.

The second problem is that I get another Travis error: the function `ring-insert` is not known to be defined. Moreover, `(require 'ring)` just before the usage doesn't seem to help. 

I'm sure these are newby questions, but I would appreciate if you could give me a quick hint on those so that I could complete this pull request as appropriate.

gracjan:
@z0rch: Your thinking is correct. `haskell-compat.el` should be the place to put backward compatibility shims. The issues should go away when you find which `require` statements should go on top of `haskell-compat.el` file.

Remember to `require haskell-compat` too.

gracjan:
@z0rch: quesiton: how do you test Emacs25? I cannot find a prebuild version that could be put in .travis.yml file. Is there anything like this available?

z0rch:
Then I'll look more and try to find out which `require` I am missing, thanks.

I use the [following ppa](https://launchpad.net/~ubuntu-elisp/+archive/ubuntu/ppa) to get pre-built Emacs 25, have a look if that would suit you.

gracjan:
@z0rch: Hmm, we seem to be using emacs-snapshot, like this:

    if [ "$EMACS" = "emacs-snapshot" ]; then
        sudo add-apt-repository -y ppa:cassou/emacs &&
        sudo apt-get update -qq &&
        sudo apt-get install -qq emacs-snapshot &&
        sudo apt-get install -qq emacs-snapshot-el emacs-snapshot-gtk;
    fi

I'm not sure which version of emacs we are getting here.

gracjan:
I see we are using old ppa for emacs-snapshot and this sources GNU Emacs 24.3.50.1, so not exactly the newest snapshot.

z0rch:
> move (require 'ring) to the top, use usual magic that disables errors.

That was the thing, with this change all errors are gone. Thanks, @gracjan!

There is no need to ignore `(require 'ring)` error, since this file exists in all Emacs versions. I have addressed all the others comments though. 

Travis is hanging today, but I installed Emacs 24 locally and ran the `make EMACS=emacs24 check`, it passed. Looking forward to see the Travis results, in order to eventually get this and #477 merged :smile:

gracjan:
I took the liberty of cleaning up the history a little (preserving your name) and using your work in #477.

gracjan:
@purcell: this pr is subsumed by #477 and can be closed.

PR #466, opened 2015-02-13
Add customizeable blank module template

### LINKED ISSUE #462: Customizable module template
I wanted to see if you'd take a PR to change `haskell-auto-insert-module-template` to use a customizeable var. At my work we use a different indentation for module headings so while I use this feature, I always have to manually reformat. I was thinking the var could be in `format` style: "--|\nimport %s where\n". Here's the relevant code.

https://github.com/haskell/haskell-mode/blob/af34d4e5afe11061bd98d2757d9d2bb93006faec/haskell-mode.el#L1036

If you would take a PR, should I keep the same function name, or make it `haskell-apply-auto-insert-template`, or if not, what would you like the var named?

MichaelXavier:
This resolves #462

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -1152,27 +1152,19 @@ given a prefix arg."
                              collect (replace-regexp-in-string "\\.l?hs$" "" part))))
     (mapconcat 'identity (reverse components) ".")))
 
+(defvar haskell-auto-insert-module-format-string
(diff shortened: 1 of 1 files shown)


MichaelXavier:
Ahem. Yep nothing wrong here. Definitely didn't leave the string "teehee" in a PR.

chrisdone:
\o/

PR #469, opened 2015-02-15
Make haskell-process-generate-tags really output TAGS

cjmazey:
The `find' invocation was only generating any output for *.hsc files.


chrisdone:
Thanks!

cjmazey:
You're welcome.

PR #470, opened 2015-02-17
Clean up examples directory

gracjan:


gracjan:
@purcell? Those files are more confusing than good.

purcell:
I can see how they might be helpful for testing, so perhaps an alternative would be to rename `examples` to `test`.

gracjan:
@purcell: I tried to use those in my recent endavours in `haskell-simple-indent` and `font-lock` and, trust me, they are not useful.

purcell:
Okay then. :-)

Issue #415, opened 2015-01-04
Assertion failure in haskell-indent-in-comment


ikirill:
With the following file:

```hs
{-# OPTIONS -w
    #-}

module Test where

-- {{{ A header
-- !

import Data.List

-- }}}
```

place point right after the exclamation mark and press enter with electric-indent-mode turned on. (With electric indent turned off, press enter, then indent with tab.)

I get:
```
Debugger entered--Lisp error: (cl-assertion-failed (<= start end))
  signal(cl-assertion-failed ((<= start end)))
  (or (<= start end) (signal (quote cl-assertion-failed) (list (quote (<= start end)))))
  haskell-indent-in-comment(60 37)
  haskell-indent-indentation-info(60)
  haskell-indent-virtual-indentation(60)
  haskell-indent-after-keyword-column(("where" 0 0) 60)
  haskell-indent-indentation-info()
  haskell-indent-cycle()
  indent-for-tab-command(nil)
  funcall-interactively(indent-for-tab-command nil)
  call-interactively(indent-for-tab-command nil nil)
  command-execute(indent-for-tab-command)
```

Maybe it should just return nil instead of raising an error?

gracjan:
That was easy to confirm and reproduce. I've added a test that should be a nice jump start for a solution.

gracjan:
We have just merged new and improved `haskell-indentation` that should solve issues like this and more.

I'm closing this particular issue. Do not hessitate to report similar issues related to `haskell-indentation`.

PR #471, opened 2015-02-17
Put Haskell symbol chars in punctuation char class.

### LINKED ISSUE #311: symbol includes trailing @ sign in as-patterns
Emacs 24.3.91.1 (built from src on ubuntu)
haskell-mode  20140726.1852 (from melpa)

When I use `find-tag` or `highlight-symbol-at-point` on an as-pattern, emacs incorrectly thinks the at-sign is part of the symbol.

For example, in the following code, if point is on the `xxs` on the left-hand side of line 3, and you type `M-.` to `find-tag`, the default symbol to search for will be `xxs@`

```
tails                   :: [a] -> [[a]]
tails []                =  [[]]
tails xxs@(_:xs)        =  xxs : tails xs
```

It's also observable in this screenshot of [this code](https://github.com/dbp/visualdb/blob/master/src/FileStore.hs) `store@` is highlighted but not `store` 

store

gracjan:
Emacs defines 'w' class as word chars and '_' as set that together with
'w' can form identifiers. Note that by this operators in programming
languages fall under punctuation '.' class and are treated as such by
for example c-mode, java-mode, etc.

It is confusing that characters defined in Haskell Report as 'symbol'
are NOT the same as 'symbol' character class defined in Emacs.

Fixes #311.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -465,20 +465,9 @@ Run M-x describe-variable haskell-mode-hook for a list of such modes."))
     (modify-syntax-entry ?\[  "(]" table)
     (modify-syntax-entry ?\]  ")[" table)
 
-    (cond ((featurep 'xemacs)
-           ;; I don't know whether this is equivalent to the below
-           ;; (modulo nesting).  -- fx
-           (modify-syntax-entry ?{  "(}5" table)
-           (modify-syntax-entry ?}  "){8" table)
-           (modify-syntax-entry ?-  "_ 1267" table))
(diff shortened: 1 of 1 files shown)


gracjan:
Fixes also #415.

gracjan:
@purcell?

purcell:
Let's give this a go... :-)

mattraibert:
Thanks for working on this!

gracjan:
You are welcome!

PR #472, opened 2015-02-17
Drop XEmacs, Emacs20 and Emacs21 support.

gracjan:


ivan-m:
Any particular reason to do this?  How do we know there isn't someone still using haskell-mode with XEmacs?

gracjan:
I should have been more clear: XEmacs supports same set of modifiers as plain Emacs and does not need special care in context of syntax tables.

PR #473, opened 2015-02-17
Pr fix string suffix prefix

gracjan:
Emacs has string-prefix-p and string-suffix-p. Use them when needed.


gracjan:
@purcell? This seems non-controversial.

Issue #475, opened 2015-02-22
haskell-mode Version header


holomorph:
Quoting the elisp manual, emphasis my own:
> ‘Version’
>      If you wish to record version numbers for the individual Lisp
>      program, put them in this line.  Lisp files distributed with Emacs
>      generally do not have a ‘Version’ header, since the version number
>      of Emacs itself serves the same purpose.  _If you are distributing a
>      collection of multiple files, we recommend not writing the version
>      in every file, but only the main one._

haskell-indent.el seems to be the only file in the project with a Version header. This should instead be haskell-mode.el, because I imagine that is the "main" one.

This became apparent when I used the shiny `package-install-file`[0] in development Emacs, and I ended up with haskell-indent in my package list. Otherwise the install was flawless. Neat!

Anyways, fixing the headers is a small change and I can make a PR if you want.

[0] http://endlessparentheses.com/new-in-emacs-25-1-easily-install-multifile-package-from-a-directory.html

gracjan:
@holomorph: Great! PR would be appreciated!

Issue #394, opened 2014-12-11
Tag a more recent version to melpa-stable


joe9:
Hello,

Just want to check if you could tag a more recent version to make it available to melpa-stable.

Thanks

gracjan:
@chrisdone, @hvr, @purcell, can you clarify release/tagging procedure?

MELPA has 20150120.1233 as current version at the time of this writing. Is this automatic?
How is melpa-stable done? I tried to find out about marmelade but failed to find out what is going on.

chrisdone:
We tend to release as dates, if ever: https://github.com/haskell/haskell-mode/releases Any other kind of versioning is meaningless.

purcell:
So shall we just arbitrarily tag current head as v15.01?

chrisdone:
Maybe once the code stops changing so much?

gracjan:
Duplicate of #226, as this is part of release procedure (or at least should be connected).

gracjan:
Melpa-stable has haskell-mode 13.10: http://stable.melpa.org/#/haskell-mode
Marmelade has 13.7: https://marmalade-repo.org/packages/haskell-mode

Was there a decision to stop updating marmelade?

gracjan:
Note also that haskell-mode-pkg.el needs update:

    (define-package "haskell-mode" "13.10" "A Haskell editing mode" 'nil)

gracjan:
This is a release workflow issue and we are working on release procedure that will also cover releasing to melpa.

Issue #226, opened 2014-01-02
Update marmalade


svenpanne:
http://marmalade-repo.org/packages/haskell-mode is quite a bit behind the code here on GitHub. It would be nice if marmalade was updated to something more recent, it is IMHO the most convenient way of installing haskell-mode.

gracjan:
@chrisdone, @hvr, can you clarify release/tagging procedure?

MELPA has 20150120.1233 as current version at the time of this writing. Is this automatic?
I tried to find out about marmelade but failed to find out what is going on.

bbatsov:
> MELPA has 20150120.1233 as current version at the time of this writing. Is this automatic?
> I tried to find out about marmelade but failed to find out what is going on.

It's automatic - MELPA builds the package from the master after each commit. MELPA Stable builds the package from the newest tag. Packages on Marmalade are updated manually.

gracjan:
Duplicate of #394, as this is part of release procedure (or at least should be connected).

purcell:
Yep, will close this and take care of updating Marmalade as part of #394.

PR #476, opened 2015-02-22
Put version header only in haskell-mode.el

holomorph:
#475

Over a year since a release? :open_mouth:


gracjan:
@purcell ?

purcell:
> Over a year since a release?

See #394 & #226.

PR #477, opened 2015-02-23
Update emacs-snapshot source to ppa:ubuntu-elisp/ppa.

gracjan:


gracjan:
And we need fix for `xref` so that it compiles. @z0rch is working on this in #465.

gracjan:
Hmm, for Emacs25:

    Running 3 tests (2015-02-26 16:01:25+0000)
    Debugger entered: ((cl-assertion-failed (<= start end) nil))
    cl--assertion-failed((<= start end))
    haskell-indent-in-comment(38 13)

This test is expected to fail due to `cl--assertion-failed`, but that exception should be caught by `ert`. It should not trigger debugger.

How to make `ert` catch `cl-assert` as it used to?

z0rch:
@gracjan, yesterday I run `make EMACS=emacs-snapshot check` on my branch in #465 (`z0rch:459-find-tag-marker-ring-is-obsolete`), and I got `checks passed!`.

I can have another look later today, perhaps it is the very recent snapshot that is failing.

z0rch:
Ah, no, perhaps more realistic idea is that my branch in #465 is based on earlier version of `master` rather than your branch here. Anyway, I'll have a look.

gracjan:
Internets are silent why cl-assert causes 'Debugger entered' even in batch mode. ERT should be able to catch it just like it did in Emacs 24.

gracjan:
@purcell?

@z0rch: thanks for help! Works now!

gracjan:
Now is much better, kludges removed.

purcell:
All looks good to me. Thanks guys.

PR #478, opened 2015-02-24
Use tags-table-list instead of tags-file-name

davidsiefert:
This way I can jump not only in the file I'm in, but to the library sources as well...

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -635,7 +635,7 @@ command from GHCi."
                        "xargs -0 hasktags -e -x"))))
       :complete (lambda (state response)
                   (when (cdr state)
-                    (let ((tags-file-name
+                    (let ((add-to-list 'tags-table-list
                            (haskell-session-tags-filename
                             (haskell-process-session (car state)))))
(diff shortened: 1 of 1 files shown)


chrisdone:
Neat!

davidsiefert:
Sorry man, looks like I broke the build.... will figure it out...

gracjan:
@davidsiefert: for your convenience:

    In toplevel form:
    haskell-commands.el:636:32:Error: malformed let binding: 
        `(add-to-list (quote tags-table-list) (haskell-session-tags-filename (haskell-process-session (car state))))'
    make: *** [haskell-commands.elc] Error 1

gracjan:
@davidsiefert: is it working better now?

gracjan:
@davidsiefert: is it working better now?

LukeHoersten:
Fixed in #592

davidsiefert:
Sorry for my inactivity--I never got a chance to look into this.  Thanks for taking ownership and adding this feature!  You guys rock!

PR #482, opened 2015-02-26
DEL now deletes a region, consistent with other emacs modes

MrBones118:
Hitting delete/backspace previously deleted only a single character, even if the region was active. This is not consistent with how other modes work. The cause was a requirement for delete-selection-mode to be active in order to delete the region. Removed this check for delete-selection-mode


gracjan:
New `haskell-indentation` does not use `delete-section-mode` at all so this issue seems not relevant anymore. Reopen it this is still a problem.

PR #483, opened 2015-02-27
merged hi2 (https://github.com/errge/hi2).

kuribas:
Merge hi2.el changes with haskell-indentation

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -1,9 +1,11 @@
 ;;; haskell-indentation.el -- indentation module for Haskell Mode
 
-;; Copyright (C) 2009  Kristof Bastiaensen
+;; Copyright (C) 2013  Kristof Bastiaensen, Gergely Risko
 
 ;; Author: Kristof Bastiaensen <kristof.bastiaensen@vleeuwen.org>
-
+;; Author: Gergely Risko <errge@nilcons.com>
+;; Keywords: indentation haskell
+;; URL: https://github.com/haskell/haskell-mode
 ;; This file is not part of GNU Emacs.
 
 ;; This file is free software; you can redistribute it and/or modify
@@ -32,22 +34,14 @@
 
 ;;; Code:
 
+(require 'hl-line)
 (require 'syntax)
 (require 'cl-lib)
 
-(defvar delete-active-region)
+(defvar haskell-indentation-dyn-first-position)
+(defvar haskell-indentation-dyn-last-direction)
+(defvar haskell-indentation-dyn-last-indentations)
 
-;; Dynamically scoped variables.
-(defvar following-token)
-(defvar current-token)
-(defvar left-indent)
-(defvar starter-indent)
-(defvar current-indent)
-(defvar layout-indent)
-(defvar parse-line-number)
-(defvar possible-indentations)
-(defvar indentation-point)
-(defvar haskell-literate)
 
 (defgroup haskell-indentation nil
   "Haskell indentation."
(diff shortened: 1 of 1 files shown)


kuribas:
Why does it complain about CL functions?

gracjan:
@kuribas: use cl-find-if, cl-remove-if and friends. It was decided (long time ago) to use `cl-lib.el` instead of `cl.el` proper so cl functions get the prefix. I guess there is no need to fight this particular decision.

purcell:
> use cl-find-if, cl-remove-if and friends.

Yes please.

Also, can we take this opportunity to prefix the unprefixed top-level symbols such as `following-token`?

Apart from that, as a hi2 user I'm very much +1 on merging this, and will do so when the above changes are done unless I hear otherwise from the other maintainers.

gracjan:
@purcell: Can we MELPA stable before merging, as this is potentially shockingly great change but a bit involved one?

purcell:
@gracjan What do you mean by "can we MELPA stable"?

gracjan:
@purcell: Tag a stable release before merging hi2.

purcell:
Perhaps. When it's been so long since the last numbered release, I doubt that this would be the change that would cause the most breakage if people suddenly start getting a new version with much more recent code: the whole `haskell-process`/`haskell-interactive-mode` stuff would be the big issue. In the meantime, we should feel free to apply to `master` reasonable changes, and be prepared to quickly fix any breakage.

gracjan:
@kuribas, @purcell: this hi2 is super cool stuff!

I fixed the missing cl-xx and then it works as a charm.

purcell:
> this hi2 is super cool stuff!

Great, isn't it? :-)

purcell:
Am I right in thinking this now needs to be rebased against `master`.

gracjan:
I merges cleanly so rebase should be no problem.

2015-03-01 10:28 GMT+01:00 Steve Purcell <notifications@github.com>:

purcell:
@gracjan Where are your fixes for `remove-if` etc?

gracjan:
Local only.

errge:
Hey Guys,

I just wanted to say that I'm so grateful and happy that @kuribas took this up and is converting my messy hi2 to something that can be contributed back to the original codebase!  You rock!

It was obviously not my intention to fork haskell-mode, just was annoyed with the indentation stuff and had the idea of having some visual feedback and wanted to practice my elisp skills on a hackathon.  After that, reality checked in and I have never found the time to properly contribute back my changes to the community.

So a big thank you for everyone involved and especially to @kuribas.  My work would not have been possible without his haskell-indentation parser and also would have never been finished without his finishing touches!

Once this is merged, please feel free to send a pull request against hi2 that warns the user that my hack is not needed anymore!

Cheers,
Gergely

haskell-indentation.el
```diff
@@ -1,9 +1,11 @@
 ;;; haskell-indentation.el -- indentation module for Haskell Mode
 
-;; Copyright (C) 2009  Kristof Bastiaensen
+;; Copyright (C) 2013  Kristof Bastiaensen, Gergely Risko
 
 ;; Author: Kristof Bastiaensen <kristof.bastiaensen@vleeuwen.org>
-
+;; Author: Gergely Risko <gergely@risko.hu>
```
errge: Can you please use errge@nilcons.com with this merge?  I know that I've used gergely@risko.hu myself, but I actually prefer errge@nilcons.com for haskell projects nowadays!

kuribas:
@errge thanks, but it wasn't that much work to integrate it, and I really like the changes you made.

gracjan:
I gave it a couple of days of usage and I think the plan to getting hi2 merged in should be:

- [ ] create a branch here so that we can work on this together
- [ ] create a test suite so that from the get-go we care about quality
- [ ] fix a couple of remaining surprises (probably caused by merge)
- [ ] fix a couple of 'debugger entered' situation

kuribas:
@gracjan If you tell me the bugs, or file an issue on my repository, I'll try to fix it.  A test suite is a good idea.  I have been thinking to run it automatically on a large haskell source file (ghc maybe?), preferably containing lots of syntax extensions.  It could then report where the parser fails.

gracjan:
I'll try to create a test suite out of this, for now as a teaser for you issues I found:

http://pastebin.com/ypr4SajS

(Note that this pertains to your merge rebased over current master).

kuribas:
I made some comments: http://pastebin.com/gU8pr6Tu

kuribas:
Note that I consider it a bug only when:
  - it throws a parseerror on valid haskell
  - it fails to show a valid indentation (or a semantically equivalent one).
  - it has inconsistent indentation (debatable in some cases)

Everything else is a feature :-)

In theory I'd like it to support all GHC syntax extensions, but in practice people need to file a bug report when they need support for the extension.

gracjan:
@kuribas:

1. Underscore bugs are due to change of interpretation of underscore, this is now 'symbol char' when it used to be 'word char'. c-mode/java-mode define it like this and we should be compatible.
2. You have seen the bugs, they were mostly the annoying ones.
3. val :: String <- return "sdfa" is an extension, I forgot which one was it.

I guess we need 1. and 2. before merging. Other stuff is lower priority.

Debugger jumps on me when there is something wrong with the code a screen or two above (invisible). It is pretty annoying. I think that indentation should do something predictable given any garbage as input. That would also ensure that possible future haskell extensions are compatible with haskell-indentation.

kuribas:
I don't think there is any need to support garbage, but I agree that giving meaningful feedback when the parser fails would be a good idea.

kuribas:
Btw, the parser looks at the code before point to see how to indent it.  It considers all code from the closest line before point that has zero indentation.  So if there is any garbage there, it will crash with a parse error.

gracjan:
@kuribas: 'Support' in the minimal sense of not doing anything super surprising, like messing up whole buffer, hanging or popping up debugger. TAB should not cause damage. TAB should also not cause interactive features.

kuribas:
@gracjan Does it do anything like that?  It shouldn't even enter the debugger for a parser error, since it uses condition-case.

gracjan:
@kuribas: does for me.

I do have `(debug-on-error . t)` but then it should not matter as docs say:

> Does not apply to errors handled by `condition-case' or those matched by `debug-ignored-errors'.

Aquamacs 3.2 GNU Emacs 24.4.51.2 (x86_64-apple-darwin14.0.0, NS apple-appkit-1343.14) of 2014-11-07 (Aquamacs-3.2) on watson.local

gracjan:
I've created a branch called 'wip/hi2' that is rebased on master, has cl-xx functions fixed and documents one failing test case (for now). @kuribas: can you have a look at it?

gracjan:
@kuribas: I've reported an issue https://github.com/nilcons/hi2/issues/15 in @errge repo.

I think that when those two are fixed we can merge!

gracjan:
@kuribas: So I fixed those myself.

gracjan:
Surprises fixed on wip/hi2, and the 'debugger entered' issues was because of my `(debug-on-error t)`.

gracjan:
Closing in favor of #502.

gracjan:
@kuribas: does hi2 reparse current function at every cursor movement? Including left-right? It is slow.

kuribas:
Yes, because it uses pre-command and post-command hooks.

PR #484, opened 2015-02-27
Fix Haddock comment regexp

karljs:
Do to some idiocy on my part, I am reopening a pull request to fix the Haddock comment regexp.  Please see the now-closed pull request #453 for some discussion and an example.

It should now only allow the following to be recognized as starting Haddock comments:

    -- docsym
    {-docsym
    {- docsym


karljs:
@gracjan Thanks for the tips.  Here's an updated attempt.

gracjan:
Thanks!

PR #485, opened 2015-02-28
Make turn-on-xxx and turn-off-xxx obsolete

gracjan:
turn-on-xxx and turn-off-xxx are a very old convention. This branch modernizes usage of modes and minor modes.


gracjan:
not good, closing

LukeHoersten:
What was not good about this?

gracjan:
Those need to be moved to separate file, documentation needs to be updated, wiki pages all over the internets need to be updated and generally phasing out needs a process.

We aren't ready for that yet.

PR #491, opened 2015-03-03
Add tests to haskell-process-wrapper-function

ardumont:
As asked https://github.com/ardumont/haskell-mode/commit/ec29dfbcde3682d429cdc3ae737e889b7ee42de4#commitcomment-10007144

Cheers,


cocreature:
As this is no longer related to cask, you should probably change the commit message. What is the reason for changing the gitignore? Is that no longer needed?

ardumont:
lol.

Indeed, I focused so much on the content of the commit, I forgot the message!

For the .gitignore,

ardumont:
Rebased!

Thanks!a

tests/haskell-process-tests.el
```diff
@@ -0,0 +1,113 @@
...
+    (with-current-buffer (url-retrieve-synchronously "http://git.savannah.gnu.org/gitweb/?p=emacs.git;a=blob_plain;hb=ba08b24186711eaeb3748f3d1f23e2c2d9ed0d09;f=lisp/emacs-lisp/package.el")
+      (save-excursion
+        (goto-char (point-min))
+        (kill-line 8)
+        (eval-buffer))))
+  (require 'package)
+  (package-initialize)
+  (add-to-list 'package-archives '("melpa" . "http://melpa.milkbox.net/packages/"))
+  (package-refresh-contents)
+  (package-install 'el-mock))
```
ardumont: As usual, beware the ugly hack!

Uglier because of emacs-23 compat' issue on package.

gracjan:
@ardumont: There is a directory `tests/compat` that already has `cl-lib.el` and `ert.el`. Just put `el-mock.el` in there and you will not need to download it from anywhere.

(el-mock.el is under GPL and has license in itself, so this copy is ok license wise).

ardumont:
Neat, way better!

Thanks for the information, will update this then.

Cheers,

gracjan:
Having said that can't we get away without el-mock.el? For example this fragment:

    (mocklet (((haskell-session-name "dummy-session") => "dumses1"))
        (haskell-process-compute-process-log-and-command "dummy-session" 'ghci)))

isn't it equivalent to:

    (let ((haskell-session-name (lambda (x) "dumses1")))
        (haskell-process-compute-process-log-and-command "dummy-session" 'ghci)))

?

Emacs is dynamically bound (usually). This has plenty of downsides but a couple of upsides, too :)

gracjan:
Also `cl` is frowned upon in favor of `cl-lib`.

ardumont:
> (let ((haskell-session-name (lambda (x) "dumses1")))
    (haskell-process-compute-process-log-and-command "dummy-session" 'ghci)))

I'm not sure.
Also, I'm not sure that the different emacs version does work the same way with this.

> Emacs is dynamically bound (usually). This has plenty of downsides but a couple of upsides, too :)

Indeed.

> Also cl is frowned upon in favor of cl-lib.

Right. I admin, I'm kind of overwhelmed with this cl thing. 
I never remember which one to use :D

gracjan:
Digging inside el-mock.el reveals that it does some advanced redefinition magic and does not use dynamic variables. It that case it makes sense to pull this dependency in.

ardumont:
> Digging inside el-mock.el reveals that it does some advanced redefinition magic and does not use dynamic variables. It that case it makes sense to pull this dependency in.

I'm at work right now.
This evening, I will rework this PR to add `el-mock.el` in the `contrib/` folder.

As a consequence, I'll also rework #492 and remove the dependency on this PR (that is by removing the common commit, this one).

Cheers,

ardumont:
Nope, that won't work.

The `tests/compat` folder is for emacs version < 24.
We need elmock for all versions.

What about adding another folder `tests/deps` (or a better name) and adding el-mock.el there.
Then updating the [EFLAG](https://github.com/haskell/haskell-mode/blob/master/Makefile#L6) entry to add this folder as dependencies?

gracjan:
`test/compat` should be added as the last element to `load-path` so that whatever is not found in standard locations is found there.

And it should be added for all Emacs versions.

ardumont:
> Nope, that won't work.

And indeed, it does not.

> What about adding another folder tests/deps (or a better name) and adding el-mock.el there.
Then updating the EFLAG entry to add this folder as dependencies?

This does :D 

> test/compat should be added as the last element to load-path so that whatever is not found in standard locations is found there.
> And it should be added for all Emacs versions.

Are you suggesting I update the `EFLAG` according to this and integrate this in this PR?

ardumont:
>> test/compat should be added as the last element to load-path so that whatever is not found in standard locations is found there.
>> And it should be added for all Emacs versions.

> Are you suggesting I update the EFLAG according to this and integrate this in this PR?

It also works (at least on my emacs-24-4).

gracjan:
@ardumont: looks very good! I'll check this out soon as time permits.

gracjan:
I've reworked your commits for better history and merged them. Thanks!

gracjan:
@ardumont: note that there are plenty of other functions in `haskell-process` that could use unit testing. You have `el-mock.el` in the repo, how about you write some more tests?

ardumont:
> I've reworked your commits for better history and merged them. Thanks!

Great!

> @ardumont: note that there are plenty of other functions in haskell-process that could use unit testing. You have el-mock.el in the repo, how about you write some more tests?

Right. I'll see what I can do.

Cheers,

ardumont:
> I'll see what I can do.

WIP https://github.com/ardumont/haskell-mode/tree/add-tests-to-haskell-process-namespace

I'm working (slowly) on it.

PR #492, opened 2015-03-03
Introduce cask

ardumont:
As asked ardumont@ec29dfb#commitcomment-10007144

Depends on #491 to explicit the problem this solves.

I suppose the original PR #372 can be closed.

Cheers,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ haskell-mode-autoloads.el
 haskell-mode.info
 haskell-mode.tmp.texi
 dir
+/.cask/
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ env:
     - EMACS=emacs23
     - EMACS=emacs24
     - EMACS=emacs-snapshot
+  global:
+    - PATH=$HOME/.cask/bin:$PATH
 
 matrix:
   allow_failures:
@@ -27,6 +29,9 @@ install:
         sudo apt-get install -qq emacs-snapshot &&
         sudo apt-get install -qq emacs-snapshot-el;
     fi
+  - make deps
+  - cask --version
+  - cask
 
 script:
   lsb_release -a && $EMACS --version && make EMACS=$EMACS check
--- a/Cask
+++ b/Cask
@@ -1,0 +1,11 @@
+(source gnu)
+(source melpa)
+
+(package-file "haskell-mode.el")
+
+(files "*.el")
+
+(development
+ (depends-on "ert")
+ (depends-on "ert-expectations")
+ (depends-on "el-mock"))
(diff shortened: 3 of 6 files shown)


cocreature:
This still contains the wrapper-process tests, not sure if that is intended. Also you are breaking emacs 23 compatibilty and sadly I don't see an easy way around that as you use package.el and cask and both work only from 24 upwards. I'm not one of the maintainers so you need to wait for them to comment but afaik they want to still support emacs 23 (hi debian stable users or who is still using 23 :)).

ardumont:
> This still contains the wrapper-process tests, not sure if that is intended. 

Yes, it is intended
This way, we can see why cask is needed (to deal with dev and/or runtime dependencies).

If the previous PR is merged (I suppose I must fix the [emacs-23 package.el dependency first](http://git.savannah.gnu.org/gitweb/?p=emacs.git;a=blob_plain;hb=ba08b24186711eaeb3748f3d1f23e2c2d9ed0d09;f=lisp/emacs-lisp/package.el)), we will see only the cask introduction as asked. 

> Also you are breaking emacs 23 compatibilty and sadly I don't see an easy way around that as you use package.el 

I suppose this [package.el](http://git.savannah.gnu.org/gitweb/?p=emacs.git;a=blob_plain;hb=ba08b24186711eaeb3748f3d1f23e2c2d9ed0d09;f=lisp/emacs-lisp/package.el)) can help.

> and cask and both work only from 24 upwards. I'm not one of the maintainers so you need to wait for them to comment but afaik they want to still support emacs 23 (hi debian stable users or who is still using 23 :)).

Yes, I wait :D 

emacs-23 is breaking because of an [unfortunate cask limitation](https://github.com/cask/cask/blob/8e81eeb1d546fb5c63121d59ff21f00732bf65a8/doc/guide/installation.rst#prerequisites).

@gracjan asks me to have a dedicated PR for introducing cask so I did.
But I don't think this will be merged because of its limitation.

*Note*
Yes, some debian user still uses emacs-23.
I suppose they are using el-get to deal with their installation reproductibility.

Anyway, thanks @cocreature for your interest :D

Cheers,

ardumont:
Rebased and removed the dependencies on #491.

I think that since the emacs-23 compatibility is a goal, this may not be the right approach right now (at least during the unknown period of emacs-23 compability for haskell-mode and no compatibility for Cask).

Also, some new discussion on #491 made me realize there is a way to deal with external dependencies which may not be perfect (duplication of existing code). 
But this works!

So, my 2 cents, I think you can close it :D 

Cheers,

gracjan:
Yep. We will have your code waiting for times when it is really needed! Closing for now!

PR #498, opened 2015-03-09
Add missing autoload cookies

petere:
Some auto-mode-alist settings didn't have an autoload cookie applied
(correctly), making them effectively useless.  Add the missing cookies
to make the behavior consistent.


gracjan:
Merged! Thanks!

gracjan:
@petere: Are there any other obvious missing things in haskell-mode that you could easily fix?

petere:
Not at the moment, but I'll be back when I find something. ;-)

PR #499, opened 2015-03-10
Fix font-lock-warning-face regular expression

bryangarza:
The regular expression was incorrectly matching `=======` without a
space afterwards, so headings were being incorrectly colored as
warnings.

### DIFF
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -372,7 +372,7 @@ Returns keywords suitable for `font-lock-keywords'."
           `(;; NOTICE the ordering below is significant
             ;;
             ("^<<<<<<< .*$" 0 'font-lock-warning-face t)
-            ("^=======" 0 'font-lock-warning-face t)
+            ("^======= .*$" 0 'font-lock-warning-face t)
             ("^>>>>>>> .*$" 0 'font-lock-warning-face t)
             ("^#.*$" 0 'font-lock-preprocessor-face t)
 


gracjan:
Merged! Thanks!

@bryangarza: Do you see any other obvious mistakes you could help fixing in `haskell-mode`?

bryangarza:
Not yet, but I'll keep an eye out for more fixes 👍

ivan-m:
This is going to cause problems with git and darcs error hunks (which is why I believe it was added, along with the lines above and below).

gracjan:
@ivan-m: good catch, for reference `simple-merge-mode` uses these regexp markers:

    (defvar simple-merge-conflict-begin-mark "^<<<<<<< .*\n"
      "String used to mark the start of a conflict region.")
    (defvar simple-merge-conflict-end-mark "^>>>>>>> .*\n"
      "String used to mark the end of a conflict region.")
    (defvar simple-merge-conflict-base-mark "^||||||| .*\n"
      "String used to mark the beginning of the base version in a conflict region.")
    (defvar simple-merge-conflict-their-mark "^=======\n"
      "String used to mark the beginning of their version in a conflict region.")

Official git definition is here: http://git-scm.com/docs/git-merge.

@bryangarza: Are you able to create a patch so that it matches the above? Please use '$' instead of '\n' as this is more correct.

gracjan:
Hmm, still the base maker regexp does not look correct when looking at git-merge site... @bryangarza: can you find what is the real definition of a git marker?

bryangarza:
@gracjan here's the patch. From what I can tell, the `|||||||` and `=======` markers will always be by themselves on a line. I can create another if PR that's preferred.

```patch
From 44865d11f64c795f39cd2a915e359f3f9d39acf1 Mon Sep 17 00:00:00 2001
From: Bryan Garza <brygarza@gmail.com>
Date: Wed, 11 Mar 2015 11:01:51 -0700
Subject: [PATCH] Correct marker regexes

`>>>>>>>` and `<<<<<<<` have text after, but `|||||||` and `=======`
do not.
---
 haskell-font-lock.el | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/haskell-font-lock.el b/haskell-font-lock.el
index 2df19b8..932b833 100644
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -372,7 +372,8 @@ Returns keywords suitable for `font-lock-keywords'."
           `(;; NOTICE the ordering below is significant
             ;;
             ("^<<<<<<< .*$" 0 'font-lock-warning-face t)
-            ("^======= .*$" 0 'font-lock-warning-face t)
+            ("^|||||||$" 0 'font-lock-warning-face t) ; "diff3" style
+            ("^=======$" 0 'font-lock-warning-face t)
             ("^>>>>>>> .*$" 0 'font-lock-warning-face t)
             ("^#.*$" 0 'font-lock-preprocessor-face t)

--
2.3.2
```

gracjan:
@bryangarza: it is easier for me to handle proper PR and you get to record your name in the history forever!

bryangarza:
Sounds good, here's the new PR: https://github.com/haskell/haskell-mode/pull/501.

PR #502, opened 2015-03-11
Merge hi2

gracjan:


gracjan:
@purcell, @chrisdone, I've been using this branch for a while and I see major improvement over old code. I think we should merge what is here. Opinions?

purcell:
:+1:

purcell:
@gracjan Please go ahead and merge this if you're satisfied it's ready.

gracjan:
I'll do a little more testing by using.

kuribas:
Maybe it would make sense to only parse when point changes to another line?

kuribas:
For example add to post command-hook:
  - if last-point-line != current-line
       parse and display overline
       last-point-line <- current-line
  - else do_nothing

gracjan:
I've fixed this already in wip/hi2 branch. It parses after a timer right
now and is not getting in the way any more.

kuribas:
A timer is rather arbitrary.  The method I described will only parse when necessary.

kuribas:
Or use both...

gracjan:
`post-command-hook` is invoked after any command, be it cursor movement, a change or unrelated. I do not know any way to know that a file got changed by a command, so we need to take the worst case scenario in account, that is that `post-command-hook` is called after a command that changed the file and did not move the cursor.

And times is also good because it moves expensive computation out of interactive loop.

kuribas:
There are not many problematic cases where a command changes the buffer but leaves the current line the same.  The only one I can think of is changing or editing a keyword as the first token on the current line.  This will change the indentations.  That could be easily checked for.  
A replace-string might change indentations, but in that case it is likely that it has messed up other indentations, so the user would need to clean that up anyway.
Having a lengthy computation in a timer doesn't feel right, and it may still run too often for low-end machines.  There were already people complaining that the parser was to slow (in the original mode).
From the emacs-lisp manual: "Indeed, for a timer to call a function that takes substantial time to run is likely to be annoying."

PR #504, opened 2015-03-12
Operators should be fontified. Fix #503.

### LINKED ISSUE #503: Operators not getting face applied
I've noticed recently that operators (`$`, `*`, `+`, etc) are no longer getting a face applied to them.

screen shot 2015-03-12 at 11 38 35

In that screen shot, I expect `*` and `+` to have the same colour as the `=` and `::`

gracjan:
Fix #503.


chrisdone:
Cheers!

PR #506, opened 2015-03-13
Buffer's directory must exist when finding cabal file

ivan-m:
I didn't think about the case that `find-file` was called to create a new file in an as-yet-nonexistent directory.

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -180,7 +180,8 @@ there is no corresponding setting with that name in the .cabal
 file), then this function returns nil."
   (interactive)
   (when (and name buffer-file-name)
-    (let ((cabal-file (haskell-cabal-find-file (file-name-directory buffer-file-name))))
+    (let* ((dir (file-name-directory buffer-file-name))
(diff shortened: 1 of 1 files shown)


ivan-m:
Actually, that probably isn't a bad idea.

What I was thinking of doing was adding in support of being able to automatically traverse non-existent path components (I use `haskell-cabal-guess-setting` to create a module skeleton, and the skeleton is inserted before I can call `make-directory`, hence how I found this issue).  The problem with that is that `haskell-cabal-find-file` currently tries to ensure that the cabal file is owned by the same user as the directory that starting directory, but the starting directory doesn't actually exist yet.

Any suggestions on how to do that?

gracjan:
(Can you change the name of `haskell-guess-setting ` to `haskell-cabal-guess-setting ` while you are at it?)

I know that `haskell-cabal-find-file` should return `nil` if the .cabal file is not available for some reason instead of causing issues.

I do not think I understood what you try to achieve so I cannot comment on that, sorry.

PR #508, opened 2015-03-13
Clean up haskell-string.el utiliy module.

gracjan:


gracjan:

purcell:
Also good - thanks.

PR #509, opened 2015-03-13
Merge hsc-mode into haskell-mode.

gracjan:


gracjan:

purcell:
What's the reason for this one?

gracjan:
Cleanup. haskell-c gave exactly nothing as haskell-mode also did font-lock `#if..#else..#endif` and other preprocessor directives.

There was nothing more in there.

purcell:
Makes sense, then. Thanks.

PR #510, opened 2015-03-13
Modernize haskell-font-lock-symbols.

gracjan:


gracjan:

purcell:
You should probably explicitly set the coding for the file now that it contains the unicode chars.

I noticed the old code used `char-displayable-p` -- what case did that cover, and is it not needed any more?

gracjan:
`char-displayable-p` seems to come from times when Emacs wasn't Unicode and font handling was subpar. Now it seems that it can display all glyphs in all practical situations.

gracjan:
This file used to contain Unicode chars before my change. Still setting encoding explicitly seems like a good idea.

purcell:
Done in e191f50

PR #511, opened 2015-03-14
Added haskell-hoogle-url

tonyday567:
Wasn't exactly sure what was idiomatic for defcustom variables in haskell-mode.  I find the fp-complete site better.


ivan-m:
Maybe add a custom option as well in case people have their own server (for internal builds, etc.)?

gracjan:
@ivan-m: as I see the custom option is already there, far to the right on the screen, isn't it?

ivan-m:
@gracjan oh, the `string` option? Missed that :/

PR #513, opened 2015-03-14
Cleanup haddock font lock

gracjan:
Improve haddock font lock propagation as this was a potentially expensive piece of code. Now it does a more local processing (previous line at most).


gracjan:
@purcell: review and merge appreciated.

purcell:
This removes the ability to suppress special fontification of haddock comments. That seems reasonable to me, but we should be prepared to restore the custom var if anyone really wants it: it might have been added for reasons other than performance, though the 2007 commit doesn't say.

PR #514, opened 2015-03-14
Cleanup pragma font lock {-# .. #-}

gracjan:
Fontify pragmas while fontifying comments. 	

Change default of haskell-pragma-face to font-lock-preprocessor-face as
this is closer to semantic meaning of a similar #pragma construct in cpp.


gracjan:
@purcell: review and merge appreciated.

PR #516, opened 2015-03-14
Added haskell-hayoo-url as a custom variable

tonyday567:
It looked like the hard-coded url was being redirected to a new url, and went with the new one as the default.  Added url-hexify-string to query.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -867,6 +867,14 @@ is asked to show extra info for the items matching QUERY.."
           (hoogle-start-server)
         (error "hoogle is not installed")))))
 
+(defcustom haskell-hayoo-url "http://hayoo.fh-wedel.de/?query=%s"
+  "Default value for hayoo web site.
+"
+  :group 'haskell
+  :type '(choice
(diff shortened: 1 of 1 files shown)


gracjan:
I'm a bit lost, does this pull request work together with #511 or instead of #511?

tonyday567:
@gracjan, it's completely independent to #511.  I just saw that the hayoo url was hardcoded and thought I may as well generalise it the same way as the hoogle edit.

gracjan:
I see. Someday we need to generalize hayoo and hoogle handling as the code looks very similar for both.

tonyday567:
@gracjan,, yes, or rationalise the hayoo and hoogle libraries even ;)

gracjan:
and by 'rationalize' you mean...?

tonyday567:
I meant that they do the same thing, and thus it would be beneficial to users of both libraries to work towards a common API, and perhaps look at commonalities in code.  Just my idle opinion.

gracjan:
Do you see obvious commonalities in code? You have changed that part of the code recently, is there something obvious that you could also improve before you forget?

tonyday567:
No, nothing comes to mind.

PR #518, opened 2015-03-15
Wrap session target

cocreature:
The session target needs to be wrapped too otherwise it gets passed as an argument to the wrapping command.

### DIFF
--- a/haskell-process.el
+++ b/haskell-process.el
@@ -93,9 +93,11 @@ HPTYPE is the result of calling `'haskell-process-type`' function."
                      nil)
                (apply haskell-process-wrapper-function
                       (list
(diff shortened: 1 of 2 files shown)


gracjan:
FAILED test-haskell-process--with-wrapper-compute-process-log-and-command-cabal-repl

Locally you can run this by:

    make check EMACS=/path/to/your/emacs

cocreature:
Ah sorry, didn't see that there are tests. I changed the test as this is exactly what this pull request is trying to accomplish.

gracjan:
Merged. @cocreature, do you see other issues that could use some improvement?

cocreature:
Nope, when I'll find issues I'll report and/or try to fix them.

Issue #460, opened 2015-02-07
Document how to get :complete support


unhammer:
Perhaps `(error "GHCi lacks ':complete' support")` in haskell-process.el:268 could be changed to say `"GHCi lacks ':complete' support (try installing 7.8 or ghci-ng)"` – when I searched for "ghc complete support" I got no hits, only found out by accident that I could just `cabal install ghci-ng` to get it

gracjan:
@unhammer: So true!

Sound like a good task. Are you able and willing to get it done and submit a pull request?

Issue #522, opened 2015-03-16
C-c C-l gives uneasy suggestion


gracjan:
`C-c C-l` says:

    You tried to do an interaction command, but an interaction mode has not been enabled yet.
    Run M-x describe-variable haskell-mode-hook for a list of such modes.

(this does not fit in minibuffer because it is two lines long, but disregarding that)

`M-x describe-variable haskell-mode-hook` tell the usual Emacs blurb and this:

    Documentation:
    Hook run after entering Haskell mode.

`M-x customize-option RET haskell-mode-hook` suggests a lot but neither `interactive-haskell-mode` nor `inf-haskell-mode` is in the list.

Then there is a link to "See also haskell-mode and (haskell-mode)haskell-mode-hook." and that leads to:

    @vindex haskell-mode-hook
    One of the important setting you should customize is the @code{haskell-mode-hook} variable  
    (@pxref{Hooks,,,emacs}) which gets run right after the @code{haskell-mode} major mode is
    initialized for a buffer. You can customize @code{haskell-mode-hook} by 
    @kbd{M-x customize-variable @key{RET} haskell-mode-hook}. It's highly recommended you
    set up indentation to match your preferences; @xref{Indentation}, for more details about
    the indentation modes included with Haskell Mode.

So still nothing about `interactive-haskell-mode`. 

There is wiki: https://github.com/haskell/haskell-mode/wiki/Haskell-Interactive-Mode-Setup , where we read:

    Run C-` to make a REPL open, this will create a session, start GHCi, and open the REPL.

Meanwhile `C-backtick` (`haskell-interactive-bring`) is not bound at all.

gracjan:
Note that more documentation for `haskell-mode-hook` is there, just it is not shown via `describe-variable`.

gracjan:
Seems Emacs just does not show documentation for hook beyond the generic `Hook run after entering XXX mode.`. So helpful!

unhammer:
Odd, when I do `C-h v haskell-mode-hook RET`, I get the full documentation (same as when looking at the elisp).

(By the way, what's the use of `haskell-interactive-bring` instead of `haskell-interactive-switch`? The only difference I can see is that -bring switches to the interactive buffer if and only if it's not visible; I find that a bit inconsistent.)

gracjan:
Interesting, I use Aquamacs, that might be the reason.

No idea about -bring and -switch. Is it sensibly fixable?

unhammer:
[...]

PR #524, opened 2015-03-17
let error say how to get complete support (#460)

unhammer:
https://github.com/haskell/haskell-mode/issues/460


gracjan:
Thanks! There are other places that could use documentation improvement, for example see #522.

PR #533, opened 2015-03-19
Define haskell-cabal-mode-map only once.

gracjan:


iquiw:
This is my mistake. Thank you for the fix!

gracjan:
@iquiw: You are forgiven! :)

PR #537, opened 2015-03-19
Load haskell-mode-map normally.

gracjan:


ivan-m:
Why not autoload?

gracjan:
Why should autoload `haskell-mode-map`? Why not autoload every other `xxxx-mode-map`?

PR #542, opened 2015-03-21
Implement 'gap' rule

### LINKED ISSUE #497: Multiline string literal not terminated by syntax highlighter
Going through the RWH book, I ended up having the following 2 definitions:

```haskell
raw = "\
      \#define X_FOOBAR   /* a foobar */\n\
      \#define X_QUUX     /* a quux */\n\
      \#define Y_BLAH     /* a blah */\n\
      \#define X_BLAH     /* a blah */\n\
      \"

defs = map sndWord . (lines >>> filter ("#define X_" `isPrefixOf`))
  where sndWord = (!! 1) . words
```

(coincidentally, GitHub's markdown has the same bug :))

screen shot 2015-03-08 at 11 51 13

Inserting a space between the `\` and `"` results in correct behavior (but obviously also a different string value stored in `raw`):

screen shot 2015-03-08 at 11 52 47

Note that _both_ snippets compile, i.e. `\"` as the last line of a multiline string is perfectly legal and not the same as ending a single-line string with a backslach.

gracjan:
Fix #497.

Implemente rule about two backslashes separated by spaces inside of strings.

While at it fix also backslashes not being escape characters outside of strings.

### DIFF
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -459,19 +459,19 @@ that should be commented under LaTeX-style literate scripts."
                               nil)
                              (t ".")))) ; other symbol sequence
 
-    ;; If we find a backslash inside a string that is considered to be
-    ;; escape we need to go back a little (over whitespace) and see if
-    ;; there is another backslash that is escape. In such situation
-    ;; the second one should have escape bit removed. This implements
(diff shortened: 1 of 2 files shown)


gracjan:
This was fixed by other pull request.

PR #555, opened 2015-03-27
Make haskell-indentation the default.

gracjan:
Add also deactivation mechanism if somebody likes any of the other modes
they will transparently turn haskell indentation off and not cause any
trouble.


gracjan:
World is not ready for this yet.

PR #557, opened 2015-03-27
Dim stylish-haskell errors.

### LINKED ISSUE #401: Prevent stylish-haskell errors
With `haskell-stylish-on-save` set to `t` and something that `stylish-haskell` can't parse (via `haskell-src-exts`) in a buffer, e.g.

```
{-# LANGUAGE ConstraintKinds #-}
{-# LANGUAGE MultiParamTypeClasses #-}

class Foo c a where

  bar :: (c a) => a -> a
```

every time the buffer is saved the `*Warnings*` buffer pops up (typically overriding another buffer I had open, thus forcing me to change it back) with a parse error from stylish-haskell.

Is there a way to prevent this?  It seems that this is because of the explicit call to `display-warnings` in https://github.com/haskell/haskell-mode/blob/a165795af8ab0347bdde8a877a88fc5ff34385db/haskell-commands.el#L760

gracjan:
Fixes #401.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -749,15 +749,17 @@ remains unchanged."
             (buffer-substring-no-properties (point-min) (point-max)))))
     (if (string= "" stderr-output)
         (if (string= "" stdout-output)
-            (funcall errout
-                     "Error: %s produced no output, leaving buffer alone" cmd)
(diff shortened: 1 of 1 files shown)


purcell:
I haven't tested this, but it looks like the right way to do this.

PR #562, opened 2015-03-29
Better test cases for haskell-indentation.

gracjan:

Issue #366, opened 2014-10-29
Support shallow indentation


mkscrg:
I find myself writing Haskell with mostly 4-space indents from the prior line. That applies in a lot of cases, e.g.
```haskell
foobar :: Foo -> Bar
foobar foo = doSomething $
    withALong subexpression (thatDoesnt fitOn) thePrevious $ line foo
```
 or
```haskell
data Foo = Foo
    { fooBar :: Bar
    , fooBaz :: Baz
    , fooQux :: Qux
    } deriving (Eq, Show, Ord)
```

In fact, I find myself  using 4-space indentation much more often than "align under start of this expression", which seems to be the `haskell-mode` approach.

Can I use `haskell-mode` with its indentation features disabled? 4-space indents are really easy to maintain by hand, and I find `haskell-mode`'s other features very useful!

gracjan:
@mkscrg: Me too! This is a reallly good idea as one of possible options for `haskell-indentation`.

gracjan:
So I worked a little on this and `haskell-indentation` is pretty close to supporting what we want here.

PR #571, opened 2015-04-04
add (+ tab-width) indentation proposals

s9gf4ult:
This adds indentation proposals on `(+ tab-indent)` 

for example, if you write

```haskell
newtype SessionId = SessionId
```

then default indentation proposals would be 

```haskell
newtype SessionId = SessionId
{ getSessionId :: Text
```

or 

```haskell
newtype SessionId = SessionId
                    { getSessionId :: Text
```

this patch adds also 

```haskell
newtype SessionId = SessionId
    { getSessionId :: Text
```

and 

```haskell
newtype SessionId = SessionId
                        { getSessionId :: Text
```

variants to indent.

In fact, it just duplicates `(+ tab-width)` for each found indentation proposal.

Unfortunately my code is hackish and this feature might be done better.

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -1102,10 +1102,11 @@ the current buffer."
    (+ left-indent haskell-indentation-left-offset)))
 
 (defun haskell-indentation-push-indentation (indent)
-  (when (or (null possible-indentations)
-            (< indent (car possible-indentations)))
-    (setq possible-indentations
-          (cons indent possible-indentations))))
+  (loop for ind in (list (+ indent tab-width) indent) do
+        (when (or (null possible-indentations)
+              (< ind (car possible-indentations)))
+          (setq possible-indentations
+                (cons ind possible-indentations)))))
 
 (defun haskell-indentation-token-test ()
   (let ((current-token nil)


gracjan:
Related #366.

gracjan:
Thanks for tackling this one. This is important feature.

Having said that I think it should be done a `indentation style` setting. Either tab-width is important or alignment is important, not both.

Rationale:

Indentation in Haskell is mostly deterministic except in a couple of places when a layout block ends. Keeping it as deterministic as possible will enable us to have a sane `indent-region`.

Your commit as it is now would increase non-determinism, which is a very sad trade off to take.

s9gf4ult:
This patch is very stupid and code in `haskell-indentation.el` is too complicated to understand it in measurable time. I think this patch is the best I can offer.

gracjan:
Your patch is an important vote for shallow indentation. Thanks.

There is a place in the code where indentation offset for the `{` is calculated. Can you find it?

s9gf4ult:
Well, there is many places where `{` pattern occures in code. `haskell-indentation.el` has very very complex code. To do this patch I just grepped `possible-indentations` defined in `haskell-indentation-parse-to-indentations` assuming it is a list of posible indentation positions. I found one such place, which is function `haskell-indentation-push-indentation`, then I modified this function and trick worked. 

Code in `haskell-indentation.el` is not very functional-style, everything mutates everything, it is very complex to understand. We need someone who knows how all this works to make this feature right.

gracjan:
That is why current project focus is lowering the entry barrier for new users, new issue reporters and first time contributors and the goal is 100 haskell-mode contributors.

We need to leave after us a maintainable piece of software. I've started doing this by writing some unit tests like in #562.

While you tried to hack on `haskell-indentation` you have learned a lot already. Can you document your knowledge in the form ofo unit tests, comments or refactorings?

s9gf4ult:
Well, I will try, but can not promise anything usefull.

gracjan:
Great! Visit us on irc if you need some help!

gracjan:
I'm closing this PR with strong conviction that #366 is the way to go.

PR #573, opened 2015-04-07
Fix haskell-docs invocation.

### LINKED ISSUE #535: Import suggester always fails
```
Identifier `lift' not in scope, choose module to import? (y or n) y
Haskell process command errored with: (wrong-type-argument listp t)
```

This lead @chrisdone to the discovery of a sub-problem where `M-! haskell-docks <$> RET` fails because of the way <$> is being escaped. This is not an issue for functions like `lift` though.

gracjan:
Fixes #535.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -198,9 +198,18 @@ PROCESS."
 
 (defun haskell-process-haskell-docs-ident (ident)
   "Search with haskell-docs for IDENT, returns a list of modules."
-  (cl-remove-if-not (lambda (a) (string-match "^[A-Z][A-Za-b0-9_'.]+$" a))
-                    (split-string (shell-command-to-string (concat "haskell-docs --modules " ident))
-                                  "\n")))
+  (cl-remove-if-not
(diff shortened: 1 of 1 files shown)


gracjan:
@LukeHoersten: can you review and merge?

LukeHoersten:
Tested with:

```haskell
test = toUpper <$> Just "hey"
```
and got the following failure (including `*Messages*` context):

```
Not in scope: ‘toUpper’
Identifier `toUpper' not in scope, choose module to import? (y or n) y
Invalid face reference: ido-first-match-face [106 times]
Haskell process command errored with: (wrong-type-argument listp t)
```

I don't think this patch fixes the issue.

gracjan:
Ugh. But we are one step forward. What are these 'face' errors? Do you have
some special ido setup?

LukeHoersten:
I only have default ido. `ido-first-match-face` isn't listed as a valid function or variable on my emacs.

LukeHoersten:
I tested with a clean emacs config and everything works fine. I tried and it works fine with `ido` in a clean config as well. Something else I have in my config is conflicting. We'll have to track it down but the `<$>` escape problem is fixed for sure.

PR #576, opened 2015-04-12
Fix buffer in the marker pushed by xref-push-marker-stack

Yuhta:
`(goto-char)` does not switch buffer; thus the buffer of `initial-loc` did not get pushed into `xref--marker-ring`/`find-tag-marker-ring`.


gracjan:
Thanks!

PR #580, opened 2015-04-16
Use more efficient looking-at-p instead of looking-at in most places

sergv:
Change `looking-at` calls into `looking-at-p`. The latter is more efficient as it does not clobber match-data. While the former is redundant in places where match-data is not used after call (through e.g. match-{beginning,end,string,string-no-properties} functions).


gracjan:
Cool! Let me review this properly.

gracjan:
Thanks for these changes, efficiency is very important. There are other places in the code that could use similar efficiency changes. Can you do some more?

sergv:
This was mostly a brief skim through the code base. I made changes whenever I was confident it won't break anything. I'll try to do some more in the future.

gracjan:
So this one didn't go very well, including a hanging emacs session. For example `haskell-decl-scan.el` is full of code fragments like:

```el
(if (looking-at "[ \t]*\\(\\sw+\\)")
    (progn
      (setq name (match-string-no-properties 1))
      (setq name-pos (match-beginning 1))
      (setq type 'class))))
```
So this change isn't as obvious as we thought.

sergv:
Yes. It seems I missed something, unfortunately.

PR #583, opened 2015-04-18
Use faces directly.

gracjan:
Use faces directy as symbols not as variables. As per font-lock
documentation:

Note that in new code, in the vast majority of cases there is no need to
create variables that specify face names.  Simply using faces directly
is enough.  Font-lock is not a template to be followed in this area.


gracjan:
@LukeHoersten: can you review this?

PR #588, opened 2015-04-21
Typo in haskell-cabal-mode manual.

mk0x9:
There is small typo in haskell-cabal-mode manual section, ``eval-after-load`` FORM must be quoted.


gracjan:
Thanks.

Issue #589, opened 2015-04-21
Prettify symbol troubles with dot operator


hmgibson23:
Great fix for the issues with ```>>=```. Thanks for the super fast response. The fix however has unfortunately introduced another bug. Now in imports ```.``` is transformed into &#8728; so you get the fairly unpleasant import Data&#8728;ByteString&#8728;Char8 instead of ```Data.ByteString.Char8```.

Sorry I feel like the scrooge of prettify-symbols-mode!

gracjan:
@hmgibson23: If you look through the code you will see that "." requires special handling.

Are you able to look at this yourself and create a pull request?

hmgibson23:
Yes will have a look

gracjan:
@hmgibson23: should be pretty easy: just check if a dot has space before and space after.

hmgibson23:
Yeah good call.

ivan-m:
@gracjan it can't just be a matter of requiring a space before/after: that means that `(.)` wouldn't be pretty-fied.

Then again, how do you tell if `Foo.Bar` is subsequently applying two constructors or just applying one qualified constructor?

gracjan:
See here:

https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-180002.4

specifically this table:

This 	Lexes as this

    f.g 	f . g (three tokens)
    F.g 	F.g (qualified ‘g’)
    f.. 	f .. (two tokens)
    F.. 	F.. (qualified ‘.’)
    F. 		F . (two tokens)

hmgibson23:
I'm not that familiar with Elisp, so would something like 

```
    (or (re-search-backward "\\<forall\\>[^.\"]*\\="
                            (line-beginning-position) t)
        (re-search-backward "[^ ]\.[^ ]"
                            (line-beginning-position) t))))
```

work?

Which makes me think the ```forall``` regex is actually redundant.

gracjan:
Close, but no cigar. I guess using `char-before` and `char-after` would be better idea. You can read about those here:

http://www.gnu.org/software/emacs/manual/html_node/elisp/Near-Point.html

PR #591, opened 2015-04-22
Dot composition detection

hmgibson23:
As per https://github.com/haskell/haskell-mode/issues/589

Seems to work pretty well but I'm welcome to alternatives!

I removed the ```forall``` check since it seems redundant now.

### DIFF
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -82,9 +82,11 @@ be disabled at that position.")
 This is the case if the \".\" is part of a \"forall <tvar> . <type>\"."
   (save-excursion
     (goto-char start)
-    (not (or
-      (string= " " (string (char-after start)))
-      (string= " " (string (char-before start)))))))
+    (or (re-search-backward "\\<forall\\>[^.\"]*\\="
+                            (line-beginning-position) t)
+        (not (or
+              (string= " " (string (char-after start)))
+              (string= " " (string (char-before start))))))))
 
 (defface haskell-keyword-face
   '((t :inherit 'font-lock-keyword-face))
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -82,8 +82,9 @@ be disabled at that position.")
 This is the case if the \".\" is part of a \"forall <tvar> . <type>\"."
   (save-excursion
     (goto-char start)
-    (re-search-backward "\\<forall\\>[^.\"]*\\="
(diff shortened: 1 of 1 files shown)


hmgibson23:
Ah okay. Sure will try those.

hmgibson23:
@gracjan not sure I understand the issue with the lonely ```.``` since it will return nil - isn't that expected?

gracjan:
It should not crash.

hmgibson23:
It doesn't crash for me. What version of Emacs are you using?

gracjan:
I haven't checked it, I just asked after seeing code.

hmgibson23:
Okay well it seems to work fine. I've put back the ```forall``` code and all seems to work well.

gracjan:
I'll check it out soon.

deviant-logic:
Tried it.  It's better than the current state of things, so if I have a vote, I vote merge it (or to be more consistent merge a change that looks like:

    (defun haskell-font-lock-dot-is-not-composition (start)
      "Return non-nil if the \".\" at START is not a composition operator.
    This is the case if the \".\" is part of a \"forall <tvar> . <type>\"."
      (save-excursion
        (goto-char start)
        (or (re-search-backward "\\<forall\\>[^.\"]*\\="
                                (line-beginning-position) t)
            (not (or (eq ?\  (char-before start))
                     (eq ?\  (char-after  start)))))))

Which uses `eq` and `?\ ` instead of turning `(char-before start)` into a string to compare it to `" "`.  I don't know if that actually matters, but it's how the rest of the file seems to go about things.

This change still misses the case where `(.)` is used as composition without spaces around it, which is common in lens-style code like `_1._2._3`.  For that matter, it also misses `(.)`.  I conjecture that the rule we're looking for is closer to: "if the character before `start` is is a word character, and the identifier before `start` starts with a capital letter, and we're not dealing with the `forall` case, then dot isn't composition"?

gracjan:
What you really want is:

    (?:MODID\.)*([allsymbolchars]+)

and then use prettyfication only when (`match-string 1)` is exactly equal to a single dot.

Current prettyfication engine is not ready for such logic.

gracjan:
@deviant-logic: You solutions is slightly better but it is still a hack. I'll merge the version by @hmgibson23 because it is good enough for the time being.

Meanwhile we wait for a skilled volunteer to actually fix the whole mess.

PR #592, opened 2015-04-22
Use tags-table-list instead of tags-file-name

geraldus:
Tested:

+ (haskell-process-generate-tags) ⇒ "Tags generated."
+ (haskell-process-generate-tags "SomeTag") ⇒ jumps to tag definition

_Sorry, I can't get rid of first commit (without delete and re-fork)._

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -658,10 +658,12 @@ command from GHCi."
                        "xargs -0 hasktags -e -x"))))
       :complete (lambda (state response)
                   (when (cdr state)
-                    (let ((tags-file-name
-                           (haskell-session-tags-filename
-                            (haskell-process-session (car state)))))
-                      (find-tag (cdr state))))
+                    (let ((session-tags
+                          (haskell-session-tags-filename
+                           (haskell-process-session (car state)))))
+                      (add-to-list 'tags-table-list session-tags)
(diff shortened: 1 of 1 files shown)


gracjan:
Do:

    git rebase -i HEAD~2

then squash those commit into one commit (read on the internet how to do this)

then force-push new version:

    git push origin +tags-table

gracjan:
In magit this is `magit-log`, got to the commit line, `Shift-E`, then `s`, then `C-c C-c`, then edit commit message so it is nice, then `Shift-P` then `- f`, then `Shift-P`.

geraldus:
Phew, looks like I've done! Thank you, Grecjan!

gracjan:
@LukeHoersten: can you also have a look at this? How tags work is a bit of mystery to me... Thanks!

LukeHoersten:
What's the difference between `tags-table-list` and `tags-file-name`? Why is `tags-table-list` better?

gracjan:
It allows to use multiple TAGS files at the same time. Think: program +
library.

geraldus:
@LukeHoersten I had similar question, original PR #478 message was:

> This way I can jump not only in the file I'm in, but to the library sources as well...

Issue #334, opened 2014-09-12
haskell-ident-at-point returns empty identifier


alexanderkjeldaas:
and haskell-ident-pos-at-point returns a tuple with the same number when no identifier is found

This is not according the the docstring and confuses for example 

```elisp
(let ((sym (haskell-ident-at-point)))
    (when sym
       ...))
```

gracjan:
@alexanderkjeldaas, thanks for the report. Are you able/willing to create a patch for this one? Note that 'haskell-ident-at-point' has nice test suite in 'tests/haskell-mode-tests.el' and that should make it easy for you to get going. Just do:

    make EMACS=/path/to/your/emacs check-haskell-mode

alexanderkjeldaas:
I had a patch somewhere, but I've changed my setup since.  I'll see if I can find it.

gracjan:
@alexanderkjeldaas: were you able to find your patch?

antalsz:
@gracjan: Given #603, should this be closed now?

PR #601, opened 2015-04-24
Guard `haskell-process-do-try-info` from non-strings

deviant-logic:
`haskell-process-do-try-info` will break process interaction if it isn't passed
a string.  This can happen when using `haskell-mode-contextual-space` and
typing a space after `::`.  At least, it happened to me when I did it.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -572,21 +572,22 @@ command from GHCi."
 
 (defun haskell-process-do-try-info (sym)
   "Get info of `sym' and echo in the minibuffer."
-  (let ((process (haskell-interactive-process)))
-    (haskell-process-queue-command
-     process
-     (make-haskell-command
-      :state (cons process sym)
-      :go (lambda (state)
-            (haskell-process-send-string
-             (car state)
-             (if (string-match "^[A-Za-z_]" (cdr state))
-                 (format ":info %s" (cdr state))
-               (format ":info (%s)" (cdr state)))))
-      :complete (lambda (state response)
-                  (unless (or (string-match "^Top level" response)
-                              (string-match "^<interactive>" response))
-                    (haskell-mode-message-line response)))))))
+  (when (stringp sym)
+    (let ((process (haskell-interactive-process)))
+             (haskell-process-queue-command
+              process
+              (make-haskell-command
+               :state (cons process sym)
+               :go (lambda (state)
+                     (haskell-process-send-string
+                      (car state)
+                      (if (string-match "^[A-Za-z_]" (cdr state))
+                          (format ":info %s" (cdr state))
+                        (format ":info (%s)" (cdr state)))))
+               :complete (lambda (state response)
+                           (unless (or (string-match "^Top level" response)
+                                       (string-match "^<interactive>" response))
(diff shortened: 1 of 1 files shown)


gracjan:
@deviant-logic: In elisp the usual protocol agreement is that functions are *not called* with arguments falling outside of their domain.

So the proper solution is to fix `haskell-mode-contextual-space` or even better is to fix the source of this problem that is described in issue #334.

So I would gladly merge changes to either `haskell-mode-contextual-space` or `haskell-ident-at-point`.

Can you create such a pull request?

deviant-logic:
@gracjan: I can fix `haskell-mode-contextual-space` tomorrow; I'm up way
past my bedtime at the moment.  Having `haskell-process-do-try-info` be
defensive may be justified here regardless, though, as its failure-mode
breaks process interaction.  I'm ambivalent.  Since
`haskell-ident-at-point` is documented to return `nil` when it can't find
an identifier, mucking with that function strikes me as inappropriate here.

gracjan:
We should try to follow the usual practice of emacs lisp and that is
callers are responsible for providing argument of correct type.

Please fix haskell-mode-contextual-space when you have time. And please
look at haskell-ident-at-point as we had reports that it does not follow
its own documentation.

antalsz:
@gracjan: `haskell-ident-at-point` now follows its own documentation (it returns `nil` when there isn't an identifier), but `haskell-ident-pos-at-point` does not (it still returns a pair of the form `(n . n)`) – I can fix that.  Perversely, however, there seems to be plenty of code which assumes that `haskell-ident-at-point` does *not* follow its own documentation and returns `""` (including possibly the root cause of the `haskell-process-do-try-info` bug).

I can try to fix this, but I'm not sure where I should put that commit – a new PR, here, on #602?

Also, a +1 for fixing the crash bug in `haskell-process-do-try-{info,type}` regardless (either @deviant-logic's fix or mine), because it renders the GHCi buffer completely unusable unless you kill the session.  (But yes, fixing the root cause is better.)

gracjan:
I would prefer another pull request that fixes everything about mishandling of haskell-ident-at-point and haskell-ident-pos-at-point.

After that is done we will look again on #601 and #602.

gracjan:
Also note that `haskell-ident-at-point` has a testsuite and tests for `haskell-ident-pos-at-point` should also get a mention in the test suite.

antalsz:
@gracjan, @deviant-logic: I submitted pull request #603, which fixes the behavior around `haskell-ident-at-point`, thereby fixing `haskell-contextual-space` and the problem here.

Thinking again about whether `haskell-do-try-info` ought to be defensive: perhaps the real issue is that an error inside the `:go` callback of `make-haskell-command` completely fouls up process interaction permanently.  Maybe that's what really ought to be fixed?

gracjan:
If I understand correctly this was subsumed by #603, right?

gracjan:
@antalsz: About `:go` being unprotected. This is something we should definitely look into. Chris rolled about home grown async/await kind of functionality probably because there was no such thing available when he did it. I guess that there should be something like that now and we should find something working out-of-the-box.

PR #602, opened 2015-04-25
Check `string-match` argument type in `haskell-do-try-info` and `haskell-do-try-type`

antalsz:
In `haskell-do-try-info` and `haskell-do-try-type`, there is code of the form `(string-match "^[A-Za-z_]" (cdr state))` that didn't check `(stringp (cdr state))`.  This messed up my Emacs's connection to the `*haskell*` buffer when hitting space would sometimes cause an error.  Checking that `string-match` can apply to `(cdr state)` seems to have fixed this, though.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -581,7 +581,7 @@ command from GHCi."
             (haskell-process-send-string
              (car state)
              (if (and (stringp (cdr state))
-                      (string-match "^[A-Za-z_]" (cdr state))
+                      (string-match "^[A-Za-z_]" (cdr state)))
                  (format ":info %s" (cdr state))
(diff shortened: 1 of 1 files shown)


gracjan:
@antalsz: Are you trying to solve the same issue as #601?

If so please read the comments down there and try to fix the core of the problem.

antalsz:
@gracjan: Oops, sorry I missed that.  (And deviant-logic balanced their parentheses correctly the first time!)  Yes, it's the same bugfix, just in two functions and with a more localized check.  I can take a look, although deep elisp hackery isn't my strong suit.

gracjan:
If I understand correctly this was subsumed by #603, right?

PR #603, opened 2015-04-25
Fix `haskell-ident-at-point` uses and `haskell-ident-pos-at-point` behavior

antalsz:
The documentation mandates that `haskell-ident-at-point` and `haskell-ident-pos-at-point` return `nil` if there is no identifier at point.  However:

1. Apparently, at some point `haskell-ident-at-point` instead returned `""` in that case.
2. Currently, `haskell-ident-pos-at-point` instead returns a cons cell of the form `(n . n)` in that case.

This pull request fixes two things:

1. Some code assumed that `haskell-ident-at-point` always returned a string; these assumptions have all been fixed.  (Commit f8521d90e7c200c07a84328c8d19e2993aad15ac.)
2. `haskell-ident-pos-at-point` now returns `nil` correctly, and any code relying on the old behavior was fixed.  (Commit 6a3257c54cf77084d106483090c393e18331cc4d.)

(Fixing `haskell-ident-pos-at-point` indirectly fixes `haskell-spanable-pos-at-point` in the same way.)

This directly fixes issue #334.  It also indirectly solves the problem in pull requests #601 and #602: `haskell-contextual-space` was comparing `haskell-ident-at-point` to `""` instead of `nil` at one point.

### DIFF
--- a/haskell.el
+++ b/haskell.el
@@ -312,7 +312,7 @@
   (let ((ident (haskell-ident-at-point))
         (tags-file-name (haskell-session-tags-filename (haskell-session)))
         (tags-revert-without-query t))
-    (when (and (stringp ident) (not (string= "" (haskell-string-trim ident))))
+    (when (and ident (not (string= "" (haskell-string-trim ident))))
       (cond ((file-exists-p tags-file-name)
              (find-tag ident next-p))
             (t (haskell-process-generate-tags ident))))))
--- a/haskell.el
+++ b/haskell.el
@@ -299,11 +299,11 @@
                           (insert (cdr mapping)))
                  (insert module)))
              (haskell-mode-format-imports)))
-          ((not (string= "" (save-excursion (forward-char -1) (haskell-ident-at-point))))
+          (t
            (let ((ident (save-excursion (forward-char -1) (haskell-ident-at-point))))
              (insert " ")
-             (haskell-process-do-try-info ident)))
-          (t (insert " ")))))
+             (when ident
+               (haskell-process-do-try-info ident)))))))
 
 ;;;###autoload
 (defun haskell-mode-jump-to-tag (&optional next-p)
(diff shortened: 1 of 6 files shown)


geraldus:
Awesome! (:

gracjan:
Awesome!

One more thing: there is a `tests/haskell-mode-tests.el` that has some unit tests for `haskell-ident-at-point`. Can you:

- [ ] add a test case for `haskell-ident-at-point` really returning `nil` when it should
- [ ] add a test case for `haskell-ident-pos-at-point` returning non-nil
- [ ] add a test case for `haskell-ident-pos-at-point` returning nil
- [ ] add a test case for `haskell-spanable-pos-at-point` returning non-nil
- [ ] add a test case for `haskell-spanable-pos-at-point` returning nil

 ?

haskell.el
```diff
@@ -312,7 +312,7 @@
   (let ((ident (haskell-ident-at-point))
         (tags-file-name (haskell-session-tags-filename (haskell-session)))
         (tags-revert-without-query t))
-    (when (not (string= "" (haskell-string-trim ident)))
+    (when (and (stringp ident) (not (string= "" (haskell-string-trim ident))))
```
gracjan: Change:

```el
(when (and (stringp ident) (not (string= "" (haskell-string-trim ident)))
```
into

```el
(when (and ident (not (string= "" (haskell-string-trim ident)))
```

`ident` should be either `nil` or `string` at this point. I would prefer it to crash if it is something else rather than silently swallow wrong type.

antalsz:
@gracjan: Done.  I could add even more tests if you felt that was important (see the commit message for c4cbd61a1ed015e76840415a90972c955f111d2a), but I don't think testing `haskell-{ident,spanable}-pos-at-point` on the same input as `haskell-ident-at-point` is deeply valuable.

PR #605, opened 2015-04-25
Make regular expressions use character classes

antalsz:
Instead of using `A-Z`, `a-z`, `A-Za-z`, and `A-Za-z0-9`, regular expressions for matching Haskell identifiers should use `[:upper:]`, `[:lower:]`, `[:alpha:]`, and `[:alnum:]` (`_`, `'`, and `.` can stay, though.)

This pull request doesn't change all such uses – for instance, looking for a `-XExtension` flag is guaranteed to use ASCII, and there are similar output-matching cases – but I hope I got everything syntax-sensitive.

[Review on Reviewable](https://reviewable.io/reviews/haskell/haskell-mode/605)


gracjan:
Here:

https://github.com/gracjan/haskell-mode/commit/7893cc2ab4b8cdc1e37a753b8253ef13cdea8d67

antalsz:
*sigh* Yes, you're right, I do want to create such a file :-)

I don't think the places you highlighted ought to be `alnum`s, though – those are places where the code is trying to match the start of the identifier.  But again, having actual named regular expressions would make that clear :-)

gracjan:
Oh, I thought it matched whole identifier or module name... I remember fixing one such spot that said `[a-zA-B]` (notice problem).

I would really love `haskell-lexeme.el`, `defconst` and matching *whole identifiers*.

Anyway, you have my initial `haskell-lexeme.el` here https://github.com/gracjan/haskell-mode/commit/7893cc2ab4b8cdc1e37a753b8253ef13cdea8d67. I probably does not even compile, I intended to submit it much later. But then you came and I would be glad if you took this over so that I can get some other stuff done.

gracjan:
@antalsz: Any progress on this one?

antalsz:
@gracjan: Sorry, I'm a grad student, and the end of the semester just hit :-)  No real progress worth showing, but it is still on my radar (and once the end of semester detritus winds down, I'll be able to spend some more cycles on it).

gracjan:
Good luck with your exams!

For the lazy that didn't know the word (like me): http://en.wikipedia.org/wiki/Detritus

gracjan:
@antalsz: Reminder that this one is needed very much!

gracjan:
@antalsz: I did #757 so you can base your work on `haskell-lexeme.el`.

gracjan:
@antalsz: ping. Can you help out with this one?

gracjan:
@antalsz: Are you able to return to this topic?

gracjan:
Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[haskell-commands.el, line 417 \[r2\]](https://reviewable.io:443/reviews/haskell/haskell-mode/605#-JyH9YED-AzRh_hVnOb6)** ([raw file](https://github.com/haskell/haskell-mode/blob/b920aec3dd797ffa56c1714c30a6ea8179531359/haskell-commands.el#L417)):</sup>
This regexp is missing apostrophes

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/haskell/haskell-mode/605)

gracjan:
@antalsz: If you will find time to update this feel free to reopen or create a new pull request. This change is needed!

Issue #607, opened 2015-04-26
`haskell-present` does not follow its doc string


geraldus:
This is what doc string of `haskell-present` states:
> Present CODE in a popup buffer suffixed with NAME and set
> SESSION as the current haskell-session.

However, there is nothing related to SESSION arg in function body [1]:
```el
(defun haskell-present (name session code)
  "Present CODE in a popup buffer suffixed with NAME and set
SESSION as the current haskell-session."
  (let* ((name (format "*Haskell Presentation%s*" name))
         (buffer (get-buffer-create name)))
    (with-current-buffer buffer
      (haskell-presentation-mode)
      (if (boundp 'shm-display-quarantine)
          (set (make-local-variable 'shm-display-quarantine) nil))
      (let ((buffer-read-only nil))
        (erase-buffer)
        (insert (propertize "-- Hit `q' to close this window.\n\n"
                            'face
                            'font-lock-comment-face))
        (let ((point (point)))
          (insert code "\n\n")
          (font-lock-fontify-region point (point))
          (goto-char point))))
    (if (eq major-mode 'haskell-presentation-mode)
        (switch-to-buffer buffer)
      (pop-to-buffer buffer))))
```
However, `haskell-process-do-simple-echo` does session assignment, which is not documented  [2]:
```el
(haskell-session-assign
  (haskell-process-session (car state)))
```
Am I missing something?

[1] https://github.com/haskell/haskell-mode/blob/master/haskell-presentation-mode.el#L38-L58
[2] https://github.com/haskell/haskell-mode/blob/master/haskell-interactive-mode.el#L1119-L1120

geraldus:
Looks like functionally of these functions got messed, look: `haskell-present` applies fontification to inserted code `(font-lock-fontify-region point (point))`. But this is not documented.
Now, `haskell-process-do-simple-echo` have optional MODE arg, which is neither documented nor used as well:
```el
(defun haskell-process-do-simple-echo (line &optional mode)
  "Send LINE to the GHCi process and echo the result in some
fashion, such as printing in the minibuffer, or using
haskell-present, depending on configuration."
  (let ((process (haskell-interactive-process)))
    (haskell-process-queue-command
     process
     (make-haskell-command
      :state (list process line mode)
      :go (lambda (state)
            (haskell-process-send-string (car state) (cadr state)))
      :complete (lambda (state response)
                  ;; TODO: TBD: don't do this if
                  ;; `haskell-process-use-presentation-mode' is t.
[...]

PR #613, opened 2015-04-28
Tidy up `haskell-present` and `haskell-process-do-simple-echo` according to #607

geraldus:
Simplify `haskell-present`:

+ do not apply text properties to hint and fontification to buffer,
  because  presentaions are fontified according to haskell mode
  automatically
+ properly assing presentation buffer to haskell session
+ make buffer readonly

Rename `haskell-process-do-simple-echo` to
`haskell-process-show-repl-response` and remove redundant mode argument

Update doc strings


geraldus:
@gracjan updated according to last comments and fixed Emacs 23 build failure

PR #627, opened 2015-05-04
Removes references to Marmalade in documentation and build process. Closes #626.

### LINKED ISSUE #626: Remove all mentions of Marmelade
Marmelade is as good as dead now, see here: https://github.com/nicferrier/elmarmalade/issues/92.

Remove all mentions of Marmelade from:

- [x] README.md
- [x] Makefile (package generation part)
- [x] [haskell-mode/wiki](https://github.com/haskell/haskell-mode/wiki)

Find out also about haskell-mode on marmelade mentions in external wikis or tutorials, contact respective authors, tell them to update.

jhenahan:
`make all` proceeds as normal, save for not generating a tar.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,6 @@
 # - Makefile
 # - haskell-mode.el
 # - haskell-mode.texi
-# - haskell-mode-pkg.el (used for marmelade)
 #
 # We should have a script that changes it everywhere it is needed and
 # syncs it with current git tag.
@@ -78,7 +77,6 @@ ELCFILES = $(ELFILES:.el=.elc)
 AUTOLOADS = haskell-mode-autoloads.el
 
 PKG_DIST_FILES = $(ELFILES) logo.svg NEWS haskell-mode.info dir
-PKG_TAR = haskell-mode-$(VERSION).tar
 ELCHECKS=$(addprefix check-, $(ELFILES:.el=))
 
 %.elc: %.el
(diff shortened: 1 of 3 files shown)


gracjan:
Can you check if `haskell-mode-pkg.el` is really unused?

Here is what MELPA does with this file:

https://github.com/milkypostman/melpa/blob/aa6427ca25596e81906cdcd79a6616f557fdf3e5/package-build.el#L745

I'm not sure if it sources same information from Commentary sections or from where. But we would like to keep the dependency information about `cl-lib`.

So can you make sure that `haskell-mode-pkg.el` is really unused?

jhenahan:
Sorry, my mistake. I didn't realize that MELPA used it.

gracjan:
I'm not sure if it REALLY uses it.

jhenahan:
According to [the API info] (https://github.com/milkypostman/melpa/blob/4a227a6d44f1981e8a3f73b253d2c33eb18ef72f/README.md#api) in the README, it does, for exactly the purpose you mentioned.

gracjan:
OK, in that case `haskell-mode-pkg.el` needs to stay.

jhenahan:
Already restored, so all should be well.

gracjan:
Can you squash https://github.com/jhenahan/haskell-mode/commit/ecd3fe7d16a1a067f6d536a24dee83dd9d04393b with https://github.com/jhenahan/haskell-mode/commit/10ef4df2e199c87d72a880a843c3c28300a728ac and then force push to the branch like this:

    git push origin +feature/no-marmalade-please

(we need clean history on `haskell-mode/master`).

jhenahan:
Should be all set.

PR #636, opened 2015-05-08
Disallow leading spaces in error regexp in interactive mode.

### LINKED ISSUE #635: Interactive mode recognizes some lines as compiler errors incorrectly
In this error message, `haskell-interactive-next-error-function` incorrectly picks up the line 

         program :: IO () ....

 as a "FILE:LINE:COL-COL: ", and then fails to find the corresponding file, causing it to fail when the filename ends up being nil.

Here's an example:

    src/EnvelopesGLFW.hs:144:3-18: Couldn't match expected type ‘Program’ with actual type ‘IO ()’ …
        In the pattern: program :: IO ()
        In a stmt of a 'do' block:
          program :: IO () <- makeProgram (ProgramSpec {..})
        In the expression:
          do { program :: IO () <- makeProgram (ProgramSpec {..});
               vao :: VertexArrayObject <- genObjectName;
               programCoords <- uniformCoords program "coords";
               bufPos <- setupBuffer vao ArrayBuffer
                         $ do { locPos <- attribLocation' program "position" FloatVec2;
                                vertexAttribArray locPos $= Enabled;
                                .... };
               .... }

ikirill:
Disallow leading spaces in error regexp in interactive mode.

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -444,7 +444,7 @@ SESSION, otherwise operate on the current buffer.
                                 (not visibility)))))))
 
 (defconst haskell-interactive-mode-error-regexp
-  "^\\([A-Z]?:?[^\r\n:]+\\):\\([0-9()-:]+\\):?")
+  "^\\(\\(?:[A-Z]:\\)?[^ \r\n:][^\r\n:]*\\):\\([0-9()-:]+\\):?")
 
 (defun haskell-interactive-at-compile-message ()
   "Am I on a compile message?"


gracjan:
Fixes #635.

Note that we should really have a test suite with examples of texts that should and should not be matched.

ikirill:
> The old regexp was really bad, but the new one of yours isn't that much better either.

`haskell-compilation-error-regexp-alist` and `haskell-process-parse-error` seem to do the actual work of parsing the error message. I just meant to fix the thing that I thought was immediately wrong with it.

What about this? It matches the filename greedily. Also, I don't think filenames are really restriced to alnum characters, even though module names might be.

    (rx line-start
        (group
         (+ (not (any control))))
        ?:
        (group
         (+ (any digit))
         ?:
         (+ (any digit))
         (? ?-
            (+ (any digit))))
        ?:
        )
    "^\\([^[:cntrl:]]+\\):\\([[:digit:]]+:[[:digit:]]+\\(?:-[[:digit:]]+\\)?\\):"

gracjan:
Does GHC do some quoting/escaping should a file name contain unfriendly
character like on of `:'"/?

gracjan:
Can you please create some ert tests for this function? I feel uneasy accepting this commit without any tests whatsoever.

Please create a new file `tests/interactive-haskell-mode-tests.el`, copy paste a test from one of the other files in `tests` and add some lines that you want matches and some other that you want to not match.

ikirill:
> Does GHC do some quoting/escaping should a file name contain unfriendly character like on of `:'"/?

Filenames are very unlikely to contain special characters like `:/`, especially when those can be directory separators. On the other hand, quotes are not a problem:

    test"`'.hs:2:8-9: Couldn't match expected type ‘IO ()’ with actual type ‘()’ …

gracjan:
@ikirill: Can we have some ert tests for this? It is a very useful functionality, but we must have tests before we merge.

gracjan:
@ikirill: What's up with this one? Have we achieved consensus here?

Do we have tests?

gracjan:
@ikirill ?

gracjan:
@bergey: Daniel, if you had time to take this one over it would be beyond great. We need tests.

bergey:
OK, I'll add tests for this.

PR #644, opened 2015-05-12
Better doc strings in haskell-commands.el

geraldus:


haskell-commands.el
```diff
@@ -41,6 +41,7 @@
   (haskell-process-start (haskell-interactive-session)))
 
 (defun haskell-process-start (session)
+  ;; FIXME Doc stirng missing SESSION appearance.
```
gracjan: Add info how to create a SESSION. Link to `haskell-session-make` would be good.

gracjan:
One fix and we can merge!

geraldus:
@gracjan done!

Issue #655, opened 2015-05-19
`find-tag' is an obsolete function (as of 25.1); use `xref-find-definitions' instead


gracjan:
`find-tag' is an obsolete function (as of 25.1); use `xref-find-definitions' instead

deviant-logic:
This was observed in a travis build failure at: https://travis-ci.org/haskell/haskell-mode/jobs/63166721#L237

PR #657, opened 2015-05-19
use `xref-find-definitons' instead of `find-tag'

deviant-logic:
Emacs 25 deprecates `find-tag'.  See issue #655.


gracjan:
Lets see if it builds!

gracjan:
@deviant-logic: thanks!

Issue #323, opened 2014-08-20
Parsing broken on quasi quotes


s9gf4ult:
for example, if I enter something like this 

```haskell
func = do 
    r <- withPG $ \con -> query con [sql|
select codes.user_id, aps.client_id
from authorization_codes as codes inner join applications as aps on codes.application_id = aps.id
where codes.access_token = ? and (codes.expires > now() or codes.expires is null)|]
                             [token]
    doSomething -- Indentation broken from this line with "parser error" message
    doMore
```

quasi quotes must be considered as strings, but they are considered as usual Haskell code. Maybe you can suggest how to solve it? I hope it is not very hard to solve this.

s9gf4ult:
bump

gracjan:
True. This should not be that hard:
- mark opening bracket as generic string
- mark closing bracket as generic string
- fontify as string or maybe something special

s9gf4ult:
Yes, but how? I am not familiar with Emacs's parsing stuff and especially how and where it used in `haskell-mode`

gracjan:
You should look at `haskell-basic-syntactic-keywords` in `haskell-font-lock.el` and understand what is going on there with double-quoted strings and single-quoted character literals. That is the hard part.

Quasi-quotes are basically string literals so you need to put "|" char class on opening bracket and on the closing bracket.

(Note additional issues with multiline, but then if you get the single line working that we can convert it to multiline mode later).

Relevant documentation starts about here: https://www.gnu.org/software/emacs/manual/html_node/elisp/Syntax-Class-Table.html

s9gf4ult:
But operator `||` becomes string starter in this case. 

String must be started at pattern, matching to regexp `\[[a-zA-Z]*\|` and ended at string `|]`. Is there such posibility?

gracjan:
More exactly the char `[` should be marked as _generic string_ and then the last ']' should be also marked as _generic string_.

The char `|` does not change its meaning. The syntax code `|` means _generic string_ and it is given to brackets. Sorry for the confusion.

(Note: some quasiquotes like e, d, t are actually Haskell so it might be good to use _paired delimiters_ `$` syntax class).

s9gf4ult:
If I understand you correctly, you propose make Emacs consider any symbols between `[` and `]` as generic string. This means that usual Haskell lists like 

```haskell 
[1,2,3]
```

and list comprehensions like

```haskell
[(i,j) | i <- [1,2], j <- [1..4]]
```

[...]

PR #660, opened 2015-05-20
Implement QuasiQuotes [quote| xx |] as string face

gracjan:


gracjan:
Fixes #323.

geraldus:
I'm curious is it possible to provide some way to fontify quasi quotes according to its syntax if it is exists like in org mode. For example I have package for Yesod's Shakespearen Templates highlighting and it will be very cool to apply fortification for such quasi quotes as `shamlet`, `hamlet`, `julius`, and etc.

gracjan:
Possible and could be done as a next step. What you need is:

- a mapping `quoter` -> `mode`
- use the mapping to to `fontify-as-mode` on the string inside quasi quote

Pretty simple.

geraldus:
If I'm not mistaken quoter to mode map will raise "reference to void variable" errors in case there is no dependent package installed. How to avoid this?

gracjan:
I guess `(if (fbound `xxx-mode) ...)` is the way to go.

geraldus:
I've thought about it, but I guess true branch of mentioned `if` expression will raise error any way.

Issue #634, opened 2015-05-08
Single buffer for Presentation Mode


geraldus:
Currently presentation mode presents messages in separate buffers. Also, currently the only one function 
uses presentations: it is `haskell-process-do-info`. Documentation for `haskell-process-use-presentation-mode` states:

> Use presentation mode to show things like type info instead of
>   printing to the message area.

and presenting type information is work in progress.

But I've found that current behaviour of presentation mode is a bit overweight. If I infer several type signatures my buffer list is overcrowded by presentation buffers, making mess when I'm switching between them <kbd>C-x b</kbd> (I have fuzzy search enabled and this makes things worst). I also think that no one need to keep all that type information or `:info` output for further work, because it is always faster to ask it again instead of search the existing buffer with that info.

So I propose to make `presentation-mode` always to exploit single buffer with constant name, rather than make new ones every time. We can prepend new output to that buffer. Optionally it is possible to add support of cleaning buffer with single keystroke, for example `c`.

gracjan:
Yes, one buffer. Read-only.

geraldus:
@gracjan can you explain this function and stuff related to it?
`haskell-presentation-present-slot` https://github.com/haskell/haskell-mode/blob/master/haskell-interactive-mode.el#L764

gracjan:
Good question. I have no clue what this os supposed to be doing.

PR #668, opened 2015-05-24
Make Haskell Presentation buffer the only

geraldus:
Related to #634


gracjan:
And general comment, can you please define keymap using `easy-mmode`, like described in #669 so that we have some control over keys?

geraldus:
@gracjan  updated!

PR #671, opened 2015-05-25
Remove menu order checking

gracjan:
This should be replaced with Emacs based validation. Flycheck does it,
we should check how it does it and use same method in Makefile.


geraldus:
Thanks :+1:

PR #672, opened 2015-05-25
Switch from signal to throw/catch in haskell-indentation

gracjan:


gracjan:
@kuribas: Can you review this pull request?

kuribas:
I originally had throw/catch in my implementation.  Someone changed it to signal, I am not sure why?

gracjan:
You can use git blame in haskell-mode repo or git blame in nilcons repo.

kuribas:
Ok, I see I had error originally, and it was changed to parse-error.  Using throw/catch makes sense.  It should not be necessary to prefix the haskell-indentation-parse-error symbol, just parse-error would be safe enough.  Though you can keep it if you like it.

PR #680, opened 2015-05-27
Align -> and => with :: in type declaration

gracjan:


gracjan:
@kuribas: Can you review this one?

Issue #669, opened 2015-05-24
Use easy-mmode-defmap to define all keymaps


gracjan:
`haskell-mode` code mostly predates Emacs 23. Since then (or even earlier) there is `easy-mmode` available and it seems to be standard method to declare keymaps, like this:

    (easy-mmode-defmap xxx-map
      `(("n" . xxx-next)
        ("p" . xxx-prev))
       "The base keymap for `xxx-mode'.")

Use this metod everywhere keymaps are defined.

fice-t:
I don't know, is it the standard? I `zrgrep`'d `/usr/share/emacs/24.5/lisp/` and got 47 results for `let ((map (make-keymap)))` and 17 results for `easy-mmode-defmap`. 3rd party packages seem to use the `let` method as well.

It does look a bit nicer though.

gracjan:
Hmm, I'm not sure at this point. Can you just look at the keymaps and do some cleanup there so that all keymaps are defined in the same way?

gracjan:
Okay, lets just close this as I'm no longer sure this is a good idea.

gracjan:
@fice-t: I've closed this easymenu, wasn't clear if it was a good idea. But there is a good idea to make eldoc-mode subsume haskell-doc-mode: https://github.com/haskell/haskell-mode/issues/1138. Do you think you could take care of it?

fice-t:
Hmm, sorry, I don't think I'll be able to man that. I expect to be busy in the next few months (I put off some stuff I shouldn't have...), so I probably won't have much time for anything bigger.
Maybe if I see something really small myself I might try it, but no requests. :-)

PR #685, opened 2015-05-31
easy-mmode-defmap, still broken

bitemyapp:
Do not merge.

 I wrapped it in that macro, but the stuff that needs to be an immediately-invoke-on-load isn't bound. watdo?

### DIFF
--- a/ghci-script-mode.el
+++ b/ghci-script-mode.el
@@ -52,7 +52,9 @@
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.ghci\\'" . ghci-script-mode))
 
-(define-key ghci-script-mode-map (kbd "C-c C-l") 'ghci-script-mode-load)
+(easy-mmode-defmap ghci-script-mode-map
+  `("C-c C-l" . ghci-script-mode-load)
+  "Keymap for ghci-script-mode")
 
 (defun ghci-script-mode-load ()
   (interactive)
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -116,21 +116,20 @@
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.cabal\\'" . haskell-cabal-mode))
 
-(defvar haskell-cabal-mode-map
-  (let ((map (make-sparse-keymap)))
-    (define-key map (kbd "C-c C-s") 'haskell-cabal-subsection-arrange-lines)
-    (define-key map (kbd "C-M-n") 'haskell-cabal-next-section)
-    (define-key map (kbd "C-M-p") 'haskell-cabal-previous-section)
-    (define-key map (kbd "M-n") 'haskell-cabal-next-subsection)
-    (define-key map (kbd "M-p") 'haskell-cabal-previous-subsection)
-    (define-key map (kbd "C-<down>") 'haskell-cabal-next-subsection)
-    (define-key map (kbd "C-<up>") 'haskell-cabal-previous-subsection)
-    (define-key map (kbd "C-c C-f") 'haskell-cabal-find-or-create-source-file)
-    (define-key map (kbd "M-g l") 'haskell-cabal-goto-library-section)
-    (define-key map (kbd "M-g e") 'haskell-cabal-goto-executable-section)
-    (define-key map (kbd "M-g b") 'haskell-cabal-goto-benchmark-section)
-    (define-key map (kbd "M-g t") 'haskell-cabal-goto-test-suite-section)
-    map))
+(easy-mmode-defmap haskell-cabal-mode-map
+  `(("C-c C-s" . haskell-cabal-subsection-arrange-lines)
+    ("C-M-n" . haskell-cabal-next-section)
+    ("C-M-p" . haskell-cabal-previous-section)
+    ("M-n" . haskell-cabal-next-subsection)
+    ("M-p" . haskell-cabal-previous-subsection)
+    ("C-<down>" . haskell-cabal-next-subsection)
+    ("C-<up>" . haskell-cabal-previous-subsection)
+    ("C-c C-f" . haskell-cabal-find-or-create-source-file)
+    ("M-g l" . haskell-cabal-goto-library-section)
+    ("M-g e" . haskell-cabal-goto-executable-section)
+    ("M-g b" . haskell-cabal-goto-benchmark-section)
+    ("M-g t" . haskell-cabal-goto-test-suite-section))
+  "Keymap for haskell-cabal-mode")
 
 ;;;###autoload
 (define-derived-mode haskell-cabal-mode fundamental-mode "Haskell-Cabal"
--- a/haskell-debug.el
+++ b/haskell-debug.el
@@ -68,16 +68,18 @@
   text-mode "Debug"
   "Major mode for debugging Haskell via GHCi.")
 
-(define-key haskell-debug-mode-map (kbd "g") 'haskell-debug/refresh)
-(define-key haskell-debug-mode-map (kbd "s") 'haskell-debug/step)
-(define-key haskell-debug-mode-map (kbd "t") 'haskell-debug/trace)
-(define-key haskell-debug-mode-map (kbd "d") 'haskell-debug/delete)
-(define-key haskell-debug-mode-map (kbd "b") 'haskell-debug/break-on-function)
-(define-key haskell-debug-mode-map (kbd "a") 'haskell-debug/abandon)
-(define-key haskell-debug-mode-map (kbd "c") 'haskell-debug/continue)
-(define-key haskell-debug-mode-map (kbd "p") 'haskell-debug/previous)
-(define-key haskell-debug-mode-map (kbd "n") 'haskell-debug/next)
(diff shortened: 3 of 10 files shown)


bitemyapp:
#669

ghci-script-mode.el
```diff
@@ -52,7 +52,9 @@
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.ghci\\'" . ghci-script-mode))
 
-(define-key ghci-script-mode-map (kbd "C-c C-l") 'ghci-script-mode-load)
+(easy-mmode-defmap ghci-script-mode-map
+  `("C-c C-l" . ghci-script-mode-load)
```
geraldus: `(kbd "C-c C-l")`?

geraldus: Oh, I suppose it's not enough. Likely it should be:

`((kbd "C-c C-l") . #'ghci-script-mode-load)`.

`#'` explanation: http://endlessparentheses.com/get-in-the-habit-of-using-sharp-quote.html

haskell-cabal.el
```diff
@@ -116,21 +116,20 @@
...
+    ("M-n" . haskell-cabal-next-subsection)
+    ("M-p" . haskell-cabal-previous-subsection)
+    ("C-<down>" . haskell-cabal-next-subsection)
+    ("C-<up>" . haskell-cabal-previous-subsection)
+    ("C-c C-f" . haskell-cabal-find-or-create-source-file)
+    ("M-g l" . haskell-cabal-goto-library-section)
+    ("M-g e" . haskell-cabal-goto-executable-section)
+    ("M-g b" . haskell-cabal-goto-benchmark-section)
+    ("M-g t" . haskell-cabal-goto-test-suite-section))
+  "Keymap for haskell-cabal-mode")
```
geraldus: same here

haskell-debug.el
```diff
@@ -68,16 +68,18 @@
...
+    ("s"   . haskell-debug/step)
+    ("t"   . haskell-debug/trace)
+    ("d"   . haskell-debug/delete)
+    ("b"   . haskell-debug/break-on-function)
+    ("a"   . haskell-debug/abandon)
+    ("c"   . haskell-debug/continue)
+    ("p"   . haskell-debug/previous)
+    ("n"   . haskell-debug/next)
+    ("RET" . haskell-debug/select))
+  "The base keymap for haskell-debug-mode")
```
geraldus: sharp quotes here and so on

bitemyapp:
@geraldus a *lot* of code does it that way unless you mean it addresses the self-call problem.

http://www.emacswiki.org/emacs/diffstat.el

https://github.com/haskell/haskell-mode/issues/669

geraldus:
@bitemyapp first of all this is from your second link ``(("\C-c\C-l" . diffstat-goto-diff)`, so at least you should escape controls in `"C-c"` and etc., and I believe this is what `kbd` function for.
Secondly, [these](https://github.com/haskell/haskell-mode/pull/668/files#diff-826c9dbdf11f0d9007e1655e7169fac1R45) keybindings perfectly work. Sharp quote denoted to compiler that following symbol is defenetely is a function.

geraldus:
Here is documentation:

> (easy-mmode-defmap M BS DOC &rest ARGS)
> Define a constant M whose value is the result of `easy-mmode-define-keymap'.

> (easy-mmode-define-keymap BS &optional NAME M ARGS)
> Return a keymap built from bindings BS.
> BS must be a list of (KEY . BINDING) where
> KEY and BINDINGS are suitable for `define-key'.

Sharp quoted symbol is suitable binding for `define-key`, so I suppose this is why it works for `easy-mmode-defmap` too.

haskell.el
```diff
@@ -35,18 +35,16 @@
...
+(easy-mmode-defmap interactive-haskell-mode-map
+  `(("C-c C-l" . haskell-process-load-or-reload)
+    ("C-c C-t" . haskell-process-do-type)
+    ("C-c C-i" . haskell-process-do-info)
+    ("M-." . haskell-mode-jump-to-def-or-tag)
+    ("C-c C-k" . haskell-interactive-mode-clear)
+    ("C-c C-c" . haskell-process-cabal-build)
+    ("C-c C-x" . haskell-process-cabal)
+    ([?\C-c ?\C-b] . haskell-interactive-switch)
+    ([?\C-c ?\C-z] . haskell-interactive-switch))
```
geraldus: Also, would it be great not to mix different ways of key definition, won't it?

bitemyapp: @geraldus that's why I did the strawman PR.

The haskell-debug-mode bindings, for example, work fine as they are currently defined.

gracjan:
@geraldus, @bitemyapp: We should follow what people in packages distributed inside of Emacs do, for example `smerge` has this definition:

```el
(easy-mmode-defmap smerge-basic-map
  `(("n" . smerge-next)
    ("p" . smerge-prev)
    ("r" . smerge-resolve)
    ("a" . smerge-keep-all)
    ("b" . smerge-keep-base)
    ("o" . smerge-keep-other)
    ("m" . smerge-keep-mine)
    ("E" . smerge-ediff)
    ("C" . smerge-combine-with-next)
    ("R" . smerge-refine)
    ("\C-m" . smerge-keep-current)
    ("=" . ,(make-sparse-keymap "Diff"))
    ("=<" "base-mine" . smerge-diff-base-mine)
    ("=>" "base-other" . smerge-diff-base-other)
    ("==" "mine-other" . smerge-diff-mine-other))
  "The base keymap for `smerge-mode'.")
```

So:
- [ ] no `kbd`
- [ ] no sharp-quotes
- [ ] control chars defined as `"\C-m"`

No sharp-quotes looks also strange to me but this is what people do so lets follow the crowd.

geraldus:
@bitemyapp sorry if I was unhelpful, I'm not an Emacs Lisp expert and thought that my comments could be relevant to your problem. One other thing is looking wrong for me is this line https://github.com/haskell/haskell-mode/pull/685/files#diff-c9062b97a2be4b8029d22242507117e6R84. Travis reports error on line 20 of `ghci-script-mode.el`, but it looks spurious. Maybe line 84 is a culprit?

bitemyapp:
@geraldus It's indeed spurious, that's me trying to get it to "fire" on mode load.

bitemyapp:
It seems the question isn't going to be answered so I'm yanking this.

gracjan:
If you mean this question "I wrapped it in that macro, but the stuff that needs to be an immediately-invoke-on-load isn't bound. watdo?" then I think I do not understand it.

Which parts are "immediately-invoke-on-load"?

And general comment: preferably nothing should be "immediately-invoke-on-load" as a general rule. Can we make it so it not "invoke-on-load"?

(It is better to keep this PR open, otherwise it is too easy to forget or miss it in github interface, so reopening)

haskell-interactive-mode.el
```diff
@@ -62,27 +62,26 @@ interference with prompts that look like haskell expressions."
...
+    ("M-p" . haskell-interactive-mode-history-previous)
+    ("M-n" . haskell-interactive-mode-history-next)
+    ("C-c C-p" . haskell-interactive-mode-prompt-previous)
+    ("C-c C-n" . haskell-interactive-mode-prompt-next)
+    ("C-<up>" . haskell-interactive-mode-history-previous)
+    ("C-<down>" . haskell-interactive-mode-history-next)
+    ("TAB" . haskell-interactive-mode-tab)
+    ("<C-S-backspace>" . haskell-interactive-mode-kill-whole-line))
   "Interactive Haskell mode map.")
+(haskell-interactive-mode-map)
```
geraldus: This line seems to be there by mistake for me.

geraldus:
@bitemyapp oh, why you closed this? You did a great work!
And as for me I also didn't catch your question. Can you clarify a bit or rephrase?

bitemyapp:
@geraldus I opened it to ask one question. I wasn't looking for feedback on anything else except for that one question.

@gracjan could we add a question or strawman issue/PR tag? It's more efficient for me to just PR stuff when I have a question. I don't have time to do Q&A in IRC, so it has to be asynchronous.

gracjan:
Okay, good. Can you point me to the stuff that is 'immediately-invoke-on-load'?

bitemyapp:
@gracjan almost definitely using the wrong terminology, but: https://github.com/haskell/haskell-mode/pull/685/files#diff-fa6c31455261917e156b4ecb38c4e5ebL71

I'm guessing I need to add a reference to the var to fire off the key bindings somewhere/somehow?

gracjan:
The statement:

```el
(define-derived-mode haskell-debug-mode ...)
```

Defined new mode and also a variable called `haskell-debug-mode-map`. Your part fills this var, which is good. This mode-map will be working when `haskell-debug-mode` is enabled. You can artificially enable it for some buffer by issuing `M-x haskell-debug-mode`, functions will fail but at least you will see if keybidings work.

bitemyapp:
@gracjan haskell-debug-mode presently works, I think the interactive stuff did not.

bitemyapp:
Migrating this to a topic branch per `CONTRIBUTING.md`

PR #687, opened 2015-06-02
Migrate to easy-mmode-defmap

bitemyapp:
Still having a problem,

`("RET"      . #'haskell-interactive-mode-return)` doesn't appear to be getting fired

screenshot from 2015-06-01 19 32 10

I tried removing `-mode` from the name based on what derived mode said about keymaps but that didn't seem to do it. What am I doing wrong?

/cc @gracjan @geraldus


geraldus:
Hmm... What `C-h c RET` shows you in Haskell REPL buffer?

gracjan:
Use "\C-m"

bitemyapp:
@gracjan 

`Use "\C-m"`

What? Where?

gracjan:
Look at sample keymap:

```el
(easy-mmode-defmap smerge-basic-map
  `(("n" . smerge-next)
    ...
    ("\C-m" . smerge-keep-current)
    ...
    ("=" . ,(make-sparse-keymap "Diff")))
  "The base keymap for `smerge-mode'.")
```

See?

bitemyapp:
Asked in IRC, was:

s/“RET”/“\C-m”/g

gracjan:
Yes, `RET` == `\C-m`.

bitemyapp:
@gracjan so why did it work before and how does this change things?

gracjan:
Value of `(kbd "RET")` is `\C-m`. I guess that is what made it work before.

bitemyapp:
@gracjan aahhhh, thank you. That would explain it :) :+1:

bitemyapp:
@gracjan a lot broke, including the ability to type `C`. I did a pull to see if a fix landed and got a merge conflict in 10+ files.

I don't think I'm going to have time to wrap this up anytime soon, apologies.

gracjan:
Thanks for the part you have done, somebody else will come by and pick up the work from place where you started.

Issue #630, opened 2015-05-05
Write script to ensure file copyright policy


gracjan:
The copyright and license notice can be put in a common COPYING file. There is no justifiable reason to also require that it be included in every file, since the individual files are not individually distributed except
as part of the entire package.

For similar discussions look for example at:
- https://github.com/iojs/io.js/pull/311
- https://github.com/rust-lang/rust/blob/00c856c0b1feac74df8b404fad94d74a0cf1624a/src/libstd/uv_ll.rs

Problems:

1. There is no COPYING file although it is claimed to be there.
2. Currently authors are listed in individual files, in somewhat ad-hoc manner. Most are in file header, some are in comments, some are listed just next to functions that they supposedly contributed.
3. Sometimes it is not really possible to know what attribution applies to.
4. Contributors of commits in github are just not listed anywhere and this does not sound right.
5. Files will undergo refactoring, for example `haskell-decl-scan.el` will be merged into `haskell-mode.el`. What to do with peoples names listed there?
6. Functions undergo refactoring. What to do with an attribute functions that gets split in two?

Good thing is that everything is GPL and that gives room to clean up. Does anybody have experience how to take attribution under control?

ivan-m:
Actually, for GPL you *are* meant to have the [header in every file](http://www.gnu.org/licenses/gpl-howto.html).

gracjan:
@ivan-m: does it say where authors should be listed?

ivan-m:
I don't think it does.

geraldus:
So…

1. > The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each release; for example, “Copyright 1998, 1999 Terry Jones” if some versions were finished in 1998 and some were finished in 1999. If several people helped write the code, use all their names.

2. > Always use the English word “Copyright”; by international convention, this is used worldwide, even for material in other languages. The copyright symbol “©” can be included if you wish (and your character set supports it), but it's not necessary. There is no legal significance to using the three-character sequence “(C)”, although it does no harm.

3. > You should also include a copy of the license itself somewhere in the distribution of your program. All programs, whether they are released under the GPL or LGPL, should include the text version<sup>[1][gpl-text]</sup> of the GPL. In GNU programs the license is usually in a file called COPYING.

4. > It is very important for practical reasons to include contact information for how to reach you, perhaps in the README file, but this has nothing to do with the legal issues of applying the license.

[...]

PR #688, opened 2015-06-02
Completion mechanics haskell-completions-grab-prefix implementation

geraldus:
This is long story. I want to make some improvements on completion stuff.
I will add changes step by step because a bit short in time and made this PR to have some feedback and help.

### DIFF
--- a/haskell-completions.el
+++ b/haskell-completions.el
@@ -28,6 +28,9 @@
 
 ;;; Code:
 
+(require 'haskell-mode)
+
+
 (defun haskell-completions-can-grab-prefix ()
   "Check if the case is appropriate for grabbing completion prefix.
 Returns t if point is either at whitespace character, or at
@@ -111,6 +114,57 @@ pragma is supported also."
                                )))))))))))))
       (when prefix-value
         (list prefix-start prefix-end prefix-value prefix-type)))))
+
+(defun haskell-completions-grab-identifier-prefix ()
+  "Grab completion prefix for identifier at point.
+Returns a list of form '(prefix-start-position
+prefix-end-position prefix-value prefix-type) for haskell
+identifier at point depending on result of function
+`haskell-ident-pos-at-point'."
+  (let ((pos-at-point (haskell-ident-pos-at-point))
+        (p (point)))
+    (when pos-at-point
+      (let* ((start (car pos-at-point))
+             (end (cdr pos-at-point))
+             (type 'haskell-completions-identifier-prefix)
+             (case-fold-search nil)
+             value)
+        ;; we need end position of result, becase of
+        ;; `haskell-ident-pos-at-point' ignores trailing whitespace, e.g. the
+        ;; result will be same for `map|` and `map  |` invocations.
+        (when (<= p end)
+          (setq end p)
+          (setq value (buffer-substring-no-properties start end))
+          (when (string-match-p (rx bos upper) value)
+            ;; we need to check if found identifier is a module name
+            (save-excursion
+              (goto-char (line-beginning-position))
+              (when (re-search-forward
+                     (rx "import"
+                         (? (1+ space) "qualified")
+                         (1+ space)
+                         upper
+                         (1+ (| alnum ".")))
+                     p    ;; bound
+                     t)   ;; no-error
+                (if (equal p (point))
+                    (setq type 'haskell-completions-module-name-prefix)
+                  (when (re-search-forward
+                         (rx (| " as " "("))
+                         start
+                         t)
+                    ;; but uppercase ident could occur after `as` keyword, or in
+                    ;; module imports after opening parenthesis, in this case
+                    ;; restore identifier type again, it's neccessary to
+                    ;; distinguish the means of completions retrieval
+                    (setq type 'haskell-completions-identifier-prefix))))))
+          (when (nth 8 (syntax-ppss))
+            ;; eighth element of syntax-ppss result is string or comment start,
+            ;; so when it's not nil word at point is inside string or comment,
+            ;; return special literal prefix type
+            (setq type 'haskell-completions-general-prefix))
+          ;; finally take in account minlen if given and return the result
+          (when value (list start end value type)))))))
 
 
 (provide 'haskell-completions)
--- a/haskell-completions.el
+++ b/haskell-completions.el
@@ -26,6 +26,14 @@
 ;; This package provides completions related functionality for
 ;; Haskell Mode such grab completion prefix at point, and etc..
 
+;; Some description
+;; ================
+;;
+;; For major use function `haskell-completions-grab-prefix' is supposed, and
+;; other prefix grabbing functions are used internally by it.  So, only this
+;; funciton have prefix minimal length functionality and invokes predicate
+;; function `haskell-completions-can-grab-prefix'.
+
 ;;; Code:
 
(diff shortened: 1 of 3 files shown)


geraldus:
Function handles all stuff here, and also returns prefixes for pragma names itself, e.g.
`{-# OP|` ⟹ `OP`
`{-# LANG|` ⟹ `LANG`
I've merged this functionality, because both cases occurs inside `{-# #-}` comment.

Any thought about how to make it a bit readable?

geraldus:
@gracjan updated

geraldus:
the rest should be easier, but I want to fix presentation buffer first

haskell-completions.el
```diff
@@ -0,0 +1,158 @@
...
+(require 'haskell-mode)
+
+(defun haskell-completions-can-grab-prefix ()
+  "Check if the case is appropriate for grabbing completion prefix.
+Returns t if point is either at whitespace character or at line
+end and preceeding character is not a whitespace or new line,
+otherwise returns nil."
+  ;; TODO Discuss this. In some cases it is better to provide completions if
+  ;; point is not at whitespace, for example when point is at closing
+  ;; parenthesis: (ma|) - point is after `a` character.
```
geraldus: In some cases it is better to provide completions if point is not at whitespace, for example when point is at closing parenthesis: `(ma|)` - point is after `a` character. Any thoughts?

geraldus:
Well, I believe I've covered all prefix cases originally defined here<sup>[1]</sup>.  Now I can write a general function for grabbing prefix at point which first invokes pragma-prefix function and if its result is nil it returns the result of identifier-prefix function.

And after that I can update completions at point function: split it to synchronous and asynchronous ones.
Synchronous function will operate like existing CAP-function, and asynchronous can provide all completions using REPL, i.e. provide completions for GHC options and language extensions running `:complete repl ":set <prefix>"`.

Any thoughts so far?

[1]:https://github.com/haskell/haskell-mode/blob/master/haskell.el#L66-L100

gracjan:
Sound like a very good idea!

geraldus:
A little progress here

gracjan:
Are there some commits here that you could spin-off as a separate pull request? This is getting rather big...

geraldus:
@gracjan I think I'm done with prefixes, can you review everything that done so far and it is could be merged after squashing some commits.  The next part is to update completion functions (still some research needed).

I have no strong opinion on some key points though:

1. Currently, all grabbing functions return `nil` if mark is activated.
2. Should completion list be provided when typing in the middle of word?

haskell-completions.el
```diff
@@ -0,0 +1,189 @@
+;;; haskell-completions.el --- Haskell Completion package
+
+;; FIXME copyright (c) 2015 Haskell mode community. All rights reserved.
```
geraldus: Is it okay to reserve copy rights to community rather than a person?

haskell-completions.el
```diff
@@ -0,0 +1,189 @@
...
+
+;; Authors:
```
geraldus: Originally, I've copied this from existing package, and replaced sigular `Author` to plural `Authors`. Not sure, is it right?

haskell-completions.el
```diff
@@ -0,0 +1,189 @@
...
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs; see the file COPYING.  If not, write to
+;; the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+;; Boston, MA 02110-1301, USA.
+
+;;; Commentary:
+
+;; This package provides functionality for all possible auto completions for
+;; Haskell Mode.  FIXME better description.
```
geraldus: Just a reminder not to forget to update package description

tests/haskell-completions-tests.el
```diff
@@ -0,0 +1,366 @@
+;;; haskell-completions-tests.el --- Tests for Haskell Completion package
+
+;; FIXME copyright (c) 2015 Haskell mode community. All rights reserved.
+
+;; Authors:
+;; Arthur Fayzrakhmanov <heraldhoi@gmail.com>
```
geraldus: Same copyright and authors comments as above.

gracjan: For licensing discussion see #630.

geraldus:
@gracjan I think GPL licence itself (`COPYING` file) could be spinned-off.  Also if GPL is the preferrable licensing method I would be nice to have some notice about licensing on [Contributing][contrib-wiki] page.

[contrib-wiki]:https://github.com/haskell/haskell-mode/wiki/Contributing

gracjan:
Overall this looks good to me. Tests provide assurance that this also works.

License stuff got merged in another commit, so it can go away. Please squash what is there to squash.

(Note: when you commit you can use `git commit --squash COMMITID` that will generate specially formatted commit message for commits that are supposed to be squashed.)

geraldus:
@gracjan I will tidy up things this evening, however, I'm still have no strong opinion about completions in the middle of strings.  Initially I've decided not to grab prefixes in the middle of string (hence no completions in this case) being influenced by some comments in **company-mode** sources<sup>[1][company-comments]</sup>.  But, when I write Elisp code **company-mode** works even in the mid of the word and it feels quite comfortable and useful, so this was wrong decision from the very beginning and provide no completions in the middle of string is likely a misfeature.

Now I'm leaning to other direction, but I don't know and can't recall how completions provision works in other packages (like **auto-complete**) and/or in other editors/IDEs  _(I can check how Sublime works in the evening).

[company-comments]:https://github.com/company-mode/company-mode/blob/master/company.el#L340-L347

gracjan:
I would love completions to work inside comments!

Note that QuasiQuotes are done using strings, so having completions there
can be also useful.

I vote for having completions everywhere!

geraldus:
OK, I've tested Sublime and it does not completes in the middle of string, so does **company-mode** with **elisp** code.  But them both provide completions when point is at punctuation, e.g. `co|)`, `com|,`, `comp|.`, and etc, and we really need it too for **haskell-mode** _(currently it's not supported)_.

I also want to have completions inside comments and string literals, this is one of the reasons of my efforts on completion improvements.  Currently, it is possible to grab prefix inside comments and strings, but I think I need to implement special completion type for these cases, e.g. `'haskell-completions-literal-prefix`.  I plan to make some hacking around `dabbrev-expand` function to get completions in this case, and as a free bonus we'll have a completions for local identifiers (in `let` blocks) and newly added things with unsaved buffer. 
If I'm not mistaken REPL does not provide completions for local identifiers, also it is required to reload file to have completions for things added since last load.  Similarly, it is required to save buffer when completions come from **TAGS** file.

As for quasi-quotes, my current haskell workflow have lot of them, so if it's possible to make QQs support a bit better I'd love to have it.  But I have nothing to say about this stuff yet.

geraldus:
pinging @gracjan, please review carefully final update

gracjan:
Code is nice, can be merged as is. Do you anything more you want to do about this?

How can I see this working besides of unit tests?

geraldus:
@gracjan nope, this is supposed to be merged as is, the next PR will touch completion functionality itself.  I'm testing this stuff manually running `M-: (haskell-completions-grab-prefix)`, etc. at various places of arbitrary haskell file.

Issue #689, opened 2015-06-03
Issue using type-at in spacemacs


fatlazycat:
Hi, not sure where the issue lies, within spacemacs or haskell-mode hence this issue.

Have raised https://github.com/syl20bnr/spacemacs/issues/1472

It appears something is sending the wrong columns to the type-at command.

Any ideas ? Could you help point me in the direction to debug this ?

Thanks

geraldus:
@fatlazycat can you `M-x toggle-debug-on-error` to enable debugging, then run type-at command and post here debugger output?

fatlazycat:
Would that goto the haskell-process-log ?

Is any different from the output 

    λ> :type-at /Users/graham/Development/GitHub/fatlazycat/NICTA-course/src/Course/ListZipper.hs 240    29 240 31 xs

    *Course.ListZipper| *Course.ListZipper| *Course.ListZipper| xs :: a -> List a -> List a
    λ> :type-at /Users/graham/Development/GitHub/fatlazycat/NICTA-course/src/Course/ListZipper.hs 240 31 240 32 xs

    *Course.ListZipper| *Course.ListZipper| *Course.ListZipper| <interactive>:1:1-2: Not in scope: ‘xs’

geraldus:
@fatlazycat do you have some error message when invoking type-at command or you just have unexpected result? In first case debugger output would be really helpful.

Also, please clarify output you posed: which command you run manually or both of them? I see different column numbers (29 -> 31, 31 -> 32) in above commands, can you also show me line 240?

fatlazycat:
Hmm, apologies seem to be getting myself confused. 

They changed the default emacs binding but their standard binding <spc> m h t now works showing the output in the haskell-process-log.

It just doesn't appear anywhere else such as the gutter, which I assume is a spacemacs / my config issue.

geraldus:
@fatlazycat if it works it's good. I've made some improvements on `haskell-process-show-type-at` recently, so if it is not works I want to know. Anyway, if you're facing any issues with `haskell-mode` please feel free to report them!

wedens:
I have the same issue with result printed only in haskell-process-log. Nothing is printed at the bottom of emacs (whatever it's called).

fatlazycat:
@wedens In a non-spacemacs emacs ?

wedens:
@fatlazycat  yeah. non-spacemacs

fatlazycat:
@geraldus hmm, how do you see the results ?

[...]

PR #698, opened 2015-06-05
Add missing parentheses

geraldus:
Pinging @gracjan. Please merge as soon as possible (:

Related to #689


gracjan:
This is interesting in the sense that our testing tools should have discovered this issue.

Why wasn't it signaled?

geraldus:
I have same question (: surprisingly Flycheck also didn't catch that.

Issue #701, opened 2015-06-06
Handle active region case better in `haskell-process-show-type-at`


geraldus:
The `:type-at` command works fine when you ask a type for several line, for example

``` haskell
main :: IO ()
main =
  do let x = "Hello, World!"
     putStrLn x
```

following command will infer type for region starts from `do` and ends at `x` on the next line:

``` haskell
λ> :type-at /Users/arthurfayzrakhmanov/Haskell/test/HelloWorld.hs 3 3 4 16
:: IO ()
```

and even without `x`:

``` haskell
λ> :type-at /Users/arthurfayzrakhmanov/Haskell/test/HelloWorld.hs 3 3 4 15
 :: String -> IO ()
```

The function `show-type-at` works fine and returns correct results in these cases _(active region wrapping several lines)_, the only caveat is that currently it passes marked expression as fifth argument to type at command and it confuses REPL and brings errors:

Haskell presentation buffer contents:

```haskell
do let x = "Hello, World!" :: String -> IO ()

<interactive>:50:6-13:
    No instance for (Show (String -> IO ()))
      arising from a use of ‘print’
    In a stmt of an interactive GHCi command: print it
```

Haskell process log contents:

``` haskell
-> :type-at /Users/arthurfayzrakhmanov/Haskell/test/HelloWorld.hs 3 3 4 14 do let x = "Hello, World!"
     putStrLn

<- do let x = "Hello, World!" :: String -> IO ()

   

   <interactive>:50:6-13:

       No instance for (Show (String -> IO ()))

         arising from a use of ‘print’

       In a stmt of an interactive GHCi command: print it
```

The fifth argument of `:type-at` is optional and plays auxiliary role:

``` haskell
λ> :type-at /Users/arthurfayzrakhmanov/Haskell/test/HelloWorld.hs 3 3 4 15 "TEST"
"TEST" :: String -> IO ()
λ> :type-at /Users/arthurfayzrakhmanov/Haskell/test/HelloWorld.hs 3 3 4 15 WHAT?
WHAT? :: String -> IO ()
```

So we have at least two options to make world better:
+ reduce expression using `haskell-utils-reduce-string`<sup>[1]</sup> before pass it to `:type-at` command
+ reduce expression like `haskell-mode-message-line`<sup>[2]</sup> does
+ replace multiline and long expressions with 
+ replace long expressions with some arbitrary string, such as `"expr"`.

Any thoughts?

[...]

Issue #700, opened 2015-06-05
Invoking haskell-process-show-type-at moves the point / cursor


fatlazycat:
Invoking haskell-process-show-type-at moves the point / cursor to the immediate right of the item you have asked the type of.

geraldus:
Ok, the culprit is `haskell-ident-pos-at-point` which is used in `show-type-at` under the hood.

fatlazycat:
Works me ! Thanks :)

PR #703, opened 2015-06-06
Minor improvements of `haskell-process-show-type-at` command 

geraldus:
Related to #700 and #701


geraldus:
I'm not sure is it good practice to write such mutating code, though Elisp is imperative if I'm not mistaken and I have no good idea how to beautify this.

PR #713, opened 2015-06-11
Make  `haskell-font-lock-symbols-alist` into a customizable variable.

sheganinans:


gracjan:
What does this pr do?

sheganinans:
Arrow syntax uses the (&&&) operator, currently haskell-mode makes it look
like (∧&) because of the line right underneath it.

geraldus:
I'm interest is there a unicode substitution for `&&&`?

sheganinans:
I don't know, although with a quick search I was able to find the asterism ⁂
U+2042 http://en.wikipedia.org/wiki/Asterism_%28typography%29 Which seems
to me like a great unicode replacement for the (***) operator!

gracjan:
In this case we need to stop doing it.  We have code that tries to prevent
&&& problems.

sheganinans:
Oh, I didn't know there was something in the works for the (&&&) problem.

I also pushed a replacement for (***), although I can put it in a separate PR if you want.

sheganinans:
There is also the upside down ampersand, ⅋, what do you think about that
geraldus?

sheganinans:
If it was up to me, I'd like for either just the ("&&&" . "&&&") hack or
both ("&&&" . "⅋" ) and ("***" . "⁂") . To keep the syntax symetric, since
it would be a shame if one of the operators only took up a single space and
the other three. I'd just rather have both take up three if we aren't going
to have either.

geraldus:
@sheganinans I totally agree with your last comment, however I'm quite unfamiliar with arrows and can't be certain about `⅋`.  If this is used for fanout arrow _(is it right name?)_ in the algebraic literature or by Haskell community I'll vote for this.  Same for `⁂`.

geraldus:
@gracjan if I understood correctly, this is some thing like hack rather that actual solution. I'm not have looked at unicode stuff myself yet, so can you point the right place to be updated?

Also I think we can keep this PR for better substitutions if offered changes are acceptable.

gracjan:
Note that original problem that &&& got composed in half is a long fixed problem, specifically in the line https://github.com/haskell/haskell-mode/blob/master/haskell-font-lock.el#L136 and the issue #582.

@sheganinans: it should be working ok for you without your hack.

As about turning &&& and *** into something else: turn defconst haskell-font-lock-symbols-alist into defcustom and then people will be able to fix their own special cases.

sheganinans:
Oh! Lovely, good to know. Yes, I must have been using a rather old version of haskell-mode. Otherwise, maybe we might want to get some sort of collective say on what people might want with arrow syntax? Say, we could propose the ⅋ and ⁂ changes on /r/haskell and see if maybe we get a better suggestion?

geraldus:
I have no Reddit account, but it will be good to know community opinion.

sheganinans:
I have one, I can post it if @gracjan gives the go ahead, he can obviously do it himself if he wants to.

gracjan:
I remember that very long time ago this was discussed on haskell-cafe (mailing list). I do not remember there was an agreement.

Still I would prefer that we do:
- [ ] turn defconst into defcustom
- [ ] people that feel this is good idea try it on themselves first for a month, come back with experiences

The reason is that &&& and *** replacements are pretty esoteric characters that may interact in a strange way with fonts, emacs rendering machinery, general layout and overall feeling needs to be proven.

sheganinans:
Alright, sounds good. I'd be satisfied with the defcustom proposal. I think it is a good balance between opening up the interface for people who want to add all sorts of unicode to their custom dsls and the people who would like to keep haskell-mode as vanilla as possible.

gracjan:
@sheganinans: will you make a PR?

sheganinans:
sure

sheganinans:
Just a sec, I guess simply s/defconst/defcustom/ is not going to suffice. Reading the docs.

gracjan:
Look in the builder:

```
In toplevel form:
haskell-font-lock.el:41:1:Error: defcustom for `haskell-font-lock-symbols-alist' fails to specify containing group
make: *** [haskell-font-lock.elc] Error 1
```

sheganinans:
Yup, I got the same error. I should build my files before I commit them. :) I'm still figuring out the customization system in emacs.

sheganinans:
Ok, I'm a little lost. I'm not sure exactly what the :type of the interface is supposed to be, I've tried :type '(alist :key-type string :value-type string) and a few others but it doesn't seem to work.

gracjan:
I know about as much as you. Are there other modes for emacs that customize
similar functionality?

sheganinans:
I haven't done much emacs customization at all, so you know much more than me.

gracjan:
This is good enough:

```
`sexp'
     The value may be any Lisp object that can be printed and read back.
     You can use `sexp' as a fall-back for any option, if you don't
     want to take the time to work out a more specific type to use.
```

sheganinans:
Never mind, it was the right thing, I was simply loading the old version of the files and not the ones i was editing.

sheganinans:
You may want some other :type, please change it to whatever interface you would prefer.

gracjan:
This is good, thanks!

sheganinans:
Thanks!

PR #714, opened 2015-06-11
Fix indentation unicode symbols.

eigengrau:
The GHC lexer actually uses the symbols ⤙ and ⤚ for arrow notation, and
these are the glyphs which the listed code-points and Unicode names in
the file refer to. However, the table from the GHC documentation, from
which the symbols in haskell-indentation.el seem to have been copied,
currently lists the wrong glyphs for these codepoints.

Cf.  [lexer source](https://github.com/ghc/ghc/blob/master/compiler/parser/Lexer.x). I’ve already filed a bug against the GHC docs.


gracjan:
Very good catch! Thanks!

Issue #715, opened 2015-06-11
haskell-indentation breaks with UnicodeSyntax


eigengrau:
Trying to use haskell-indentation-mode in haskell-mode 20150512.2319 using the following example will fail with `Illegal token: ←`.

```
{-# LANGUAGE UnicodeSyntax #-}

fun = do
  x ← "abc"
```

gracjan:
Note that we have a very nice testsuite in `tests/haskell-indentation-tests.el` where you can add a test for this case.

gracjan:
This one is working correctly now. Thanks for the report!

eigengrau:
Thank you!

PR #716, opened 2015-06-11
Test suite: Do checks on UnicodeSyntax, too.

eigengrau:
This automatically transforms check input to UnicodeSyntax and
runs the test twice; once on the original, once on the Unicode
version.

Note that I don’t really know what I’m doing with my Elisp. Is this
usable to you? In my test it catches #715.

Since this is using a dumb search&replace, instead of requiring
separate tests for Unicode cases, e. g. things like `->` can’t be used
in Haskell strings inside of the test cases, which is probably not an
issue.

### DIFF
--- a/tests/haskell-indentation-tests.el
+++ b/tests/haskell-indentation-tests.el
@@ -59,6 +59,7 @@ Example of lines:
    "^ ^       ^"))
 
 (ert-deftest haskell-indentation-check-2-unicode ()
+  :expected-result :failed
   "Handle underscore in identifiers (unicode)"
   (haskell-indentation-check
    "function = do"
@@ -76,6 +77,7 @@ Example of lines:
 
 (ert-deftest haskell-indentation-check-2a-unicode ()
   "Handle apostrophe in identifiers (unicode)"
+  :expected-result :failed
   (haskell-indentation-check
    "function = do"
    "  (_'x') ← return ()"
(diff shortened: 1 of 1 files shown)


eigengrau:
> Almost, but not quite. What you want is to match consecutive characters that compose a symbol and then only change a symbol that is in the list, so for example <<- should not be partially changed into < plus Unicode for <-.

Right, good catch. This would affect `-<` and `-<<`. Might be worked around by iterating over longer replacements first.

> This idea looks a bit too smart for my taste. It is worth to be more explicit in tests about what is and what is not tested as in case of failure error messages are incomprehensible.

Fair enough. I don’t know anything about ERT, but I wonder if one could do this automatically and still have nominally separate units for Unicode and ASCII (generating units on the fly before a test run?). A mixture case could be added manually. If the parsing haskell-indentation does is uniform, one could just have a single test case for indentation with UnicodeSyntax; qut looking over the file, there appeared to be a lot of individual literal references to things like `<-`, which is why I tought one might want to test the unicode equivalent of every regular test case.

gracjan:
For now please just copy&paste the tests that are interesting and change
Unicode where you need.

We can generalize later when we know how to do it.

Please mark those tests that fail as:

:expected_result :failed

eigengrau:
Ah, you added a new one. Rebased!

eigengrau:
> Please mark those tests that fail as: :expected_result :failed

Just for the ones that should fail, though, right? Or does this have something to do with CI?

gracjan:
Yes, make TravisCI happy. The actual fixes will come later.

gracjan:
Great! Thanks!

Now we need to actually fix it :)

eigengrau:
:)

PR #718, opened 2015-06-11
Completions package

geraldus:
Let's move a little by little.  This is starting point for completion package.

I'm unsatisfied with tests because when some assertions are failed it's quite hard to find which one is actual culprit.  I'd rather split all tests into separate functions, but this will bring a lot of boilerplate code.  I've looked at `defmacro` and `haskell-indentation-check` function but have found no good replacement yet and gave up.  So, let's merge this as is, and improve tests later.

Related #688

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,7 @@ ELFILES = \
 	haskell-compat.el \
 	haskell-compile.el \
 	haskell-complete-module.el \
+	haskell-completions.el \
 	haskell-customize.el \
 	haskell-debug.el \
 	haskell-decl-scan.el \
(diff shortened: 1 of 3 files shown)


gracjan:
About test:

What I usually define is a test helper function that is used like this:

```el
(defun check-can-grab-prefix (lines expected-result)
    (with-buffer
       (insert-lines)
       (find-where-is-$-and-place-cursor-there)
       (remove-$)
       (should (equal expected-result (haskell-completions-can-grab-prefix))))

(ert-deftest haskell-completions-can-grab-prefix-test () 
   "Test testing foobar"
   (check-can-grab-prefix
      '("{-# LANG$ #-}"  ;; input lines, $ is where cursor should be, $ is removed after placing cursor there
        "Other line")
      t)                 ;; or nil, this is the expected result 
```

PR #722, opened 2015-06-13
Fixed emacs version to support version 23

ilitzroth:


ilitzroth:
Changed supported version to back to 23

PR #732, opened 2015-06-19
New synchronous completion function

geraldus:
New function also provides a completions inside comments and string literals using `dabbrev` expansions, provide completions for pragmas (need some minor improvements for `CONLIKE` and `SPECIALIZE` pragmas), GHC options completions, language extensions, module names and arbitrary identifiers.

### DIFF
--- a/haskell-completions.el
+++ b/haskell-completions.el
@@ -37,6 +37,8 @@
 ;;; Code:
 
 (require 'haskell-mode)
+(require 'haskell-process)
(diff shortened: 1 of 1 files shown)


geraldus:
@gracjan I guess GHC is not available in Travis build environment and I have no idea how to write tests for completion function in this case.

gracjan:
Merged. We will think about tests later.

Issue #730, opened 2015-06-18
Turn haskell-ghc-supported-extensions and haskell-ghc-supported-options customization options to variables


geraldus:
I can't be certain, but `haskell-ghc-supported-extensions`<sup>[1][exts]</sup> and `haskell-ghc-supported-options`<sup>[2][opts]</sup> are likely useless, when you try unfold them it brings a huge list.  I can't imagine the case when it needed to extend these lists, or remove/modify items.  In both cases values are fetched by invoking shell commands (`ghc --show-options` and `ghc --supported-extensions`).  I think it's better to modify these options to variables.

[opts]:https://github.com/haskell/haskell-mode/blob/master/haskell-customize.el#L337
[exts]:https://github.com/haskell/haskell-mode/blob/master/haskell-customize.el#L331

ivan-m:
I know that shm uses `ghc --supported-languages` (which is a alias for `ghc --supported-extensions`) to complete `LANGUAGE` pragmas; possibly something similar could be done in haskell-mode itself?

(`haskell-ghc-supported-options` could then be used to help auto-complete fields in .cabal files...)

geraldus:
It's very strange to me, that we have following code<sup>[1][lang-exts]</sup> in `haskell-customize.el`:
```el
(defcustom haskell-language-extensions
  '()
  "Language extensions in use. Should be in format: -XFoo, -XNoFoo etc."
  :group 'shm
  :type '(repeat 'string))
```
Should it be move to **SHM** itself? @ivan-m are you talking about this? In my Emacs installation it's empty (as default value).

Looks like SHM get extensions from `Language.Haskell.Extensions`<sup>[2][hask-exts]</sup> using its binary<sup>[3][shm-src]</sup>.

@chrisdone can you clarify a bit? 

[lang-exts]:https://github.com/haskell/haskell-mode/blob/master/haskell-customize.el#L325-L329
[hask-exts]:http://hackage.haskell.org/package/Cabal-1.22.4.0/docs/Language-Haskell-Extension.html#
[shm-src]:https://github.com/chrisdone/structured-haskell-mode/blob/2f41350bc320d364fe8ebb1c41030cb3da516ae8/src/Main.hs#L256-L293

ivan-m:
Looks like something that wasn't moved over to shm properly.

[...]

PR #736, opened 2015-06-20
Turn few customizable options to variables

geraldus:
Turn `haskell-ghc-supported-extensions` and
`haskell-ghc-supported-options` into variables, move them outside
customize package to haskell-mode one.

Related to #730

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -336,17 +336,6 @@ same vein as `haskell-indent-spaces'."
   :group 'haskell
   :type '(repeat 'string))
 
-(defcustom haskell-ghc-supported-extensions
-  (split-string (shell-command-to-string "ghc --supported-extensions"))
-  "List of language extensions supported by the installed version of GHC."
-  :group 'haskell
-  :type '(repeat string))
-
-(defcustom haskell-ghc-supported-options
-  (split-string (shell-command-to-string "ghc --show-options"))
-  "List of options supported by the installed version of GHC."
-  :group 'haskell
-  :type '(repeat string))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Accessor functions
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -202,6 +202,20 @@ be set to the preferred literate style."
     map)
   "Keymap used in Haskell mode.")
 
+
+(defvar haskell-ghc-supported-extensions
+  (split-string (shell-command-to-string "ghc --supported-extensions"))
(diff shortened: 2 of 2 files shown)


ivan-m:
There is a possibility that someone might want these customised as a minimal subset that's supported by a range of GHC values (e.g. starts Emacs with ghc set to 7.10, at some point changes global settings so that /usr/bin/ghc points to 7.8 which has different supported extensions, etc.).

That said, I think the likelihood of this is minimal and unlikely to cause any problems.  I thought it should just be mentioned as part of the risk-assessment of changing this behaviour.

geraldus:
@gracjan don't merge this, I want to clarify doc strings a bit.

@ivan-m your statement sounds reasonable.  Let me share my thoughts about this.  If need to change some of these options I will do it imperatively rather than customizing these vars, because in the latter case my init file will grew up significantly as the list are too long, e.g.

```el
(setq haskell-ghc-supported-extensions (delete "whatever" haskell-ghc-supported-extensions))
```

Also, currently these vars are only used in completion functions, and I suppose few wrong completions are not critical, as you mentioned already.

What about this: turn these options into vars and define two customizations, e.g. `haskell-ghc-supported-extensions-override` and `haskell-ghc-supported-options-overrides`, a lists containing items of form `+ option` or `- option`.  For example if you want to exclude `RankNTypes` and include `OverloadedStrings` from/in supported extensions you can customize `haskell-ghc-supported-extensions-override` to `(list "+ RankNTypes" "- OverloadedString")`.  This is just example, we can provide some nicer way to customize this.  I think this is easier to manage for user than a huge default list of values.  To take this in account it is possible to write some helper functions, e.g. `haskell-get-supported-extensions-list` and `haskell-get-supported-options-list`.

ivan-m:
That makes sense to me.

geraldus:
@gracjan what do you think?
@ivan-m thank you, there is no much feedback on issues currently.

gracjan:
The proposed machinery requires a lot of effort that could be spend
elsewhere. Just imagine the amount of test you would need to write to make
sure it works as expected and the amount of documentation explaining what
is going on.

I'm afraid I will not be able to agree to pull this complexity in for the
sake of future haskell-mode contributors that will come after all of us
three have moved on.

geraldus:
@gracjan, so if I update doc string will you merge this?

gracjan:
Yes



W dniu niedziela, 21 czerwca 2015 Артур Файзрахманов <
notifications@github.com> napisał(a):

geraldus:
@gracjan updated

gracjan:
Baah. One more thing to check: does `structured-haskell-mode` work if you moved those variables? It might do `(require 'haskell-customize)` and there it can fail.

(Somebody asked what is a *submarine dependency*. Well, this is it).

gracjan:
Okay, I've checked this and no issues with shm. Lets proceed.

PR #743, opened 2015-07-02
`haskell-interactive-bring': Use `pop-to-buffer'.

### LINKED ISSUE #710: Why does `haskell-interactive-bring' call `delete-other-windows'?
I find it rather annoying that running ```haskell-interactive-bring``` deletes all my windows (via ```delete-other-windows``` when starting a new session. Why doesn't it just use ```switch-to-buffer``` or similar?

(If there's a sense in the current behavior, please make it at least customizable)

the-kenny:
This commit removes counter-intuitive logic when bringing a project buffer to the front.

One could argue that we want to make the old behavior configurable. If so, please comment here and I'll add a config option.

Fixes #710.

### DIFF
--- a/haskell.el
+++ b/haskell.el
@@ -369,12 +369,7 @@
   (interactive)
   (let* ((session (haskell-session))
          (buffer (haskell-session-interactive-buffer session)))
-    (unless (and (cl-find-if (lambda (window) (equal (window-buffer window) buffer))
-                             (window-list))
-                 (= 2 (length (window-list))))
-      (delete-other-windows)
-      (display-buffer buffer)
-      (other-window 1))))
+    (pop-to-buffer buffer)))
 
 ;;;###autoload
 (defun haskell-process-load-file ()


geraldus:
This behaviour is exact what I expect from `haskell-interactive-bring`.  Let's see what other people say.

ivan-m:
Can you please describe the behaviour that this implements and how it differs from the current behaviour?  I did look at #710 but don't really understand the problem (maybe I'm just used to dealing with the current behaviour).

geraldus:
@ivan-m I always have several windows layout (most of the time there is only two of them, but rarely I use three or four also).  If I start new session using `haskell-interactive-bring` it replaces all my windows with single one visiting session's (REPL's) buffer.  This is annoying to me, because in this case I have to split windows again, and revisit all needed buffers.  Though I use REPL occasionally, but usually I need it for type info and compilation only.  New behaviour does not breaks layout, just pops session buffer.

Here is a quick [demo](https://drive.google.com/file/d/0BwLfIDb7xEV7S0lQUG9QY3lNNFE/view?usp=sharing) of current behaviour.

the-kenny:
The old behavior removed all window-splits in the current frame when
bringing the interaction buffer to the front *except* when there were
exactly two windows (e.g. one split) and one of them was the interaction
buffer.

The new behavior relies on the standard Emacs function to bring some
background buffer to the front. It will show the buffer in *some*
window, preferably not the current one (so current buffer showing code
will still be visible) and splits if necessary.

If the interaction buffer is currently visible on another frame, that
frame will be made visible.


The reason I implemented this was that I mostly work on high-resolution
displays with a fullscreen emacs frame. That frame is split multiple
times, showing multiple code buffers.

The old behavior of `haskell-interactive-bring' threw away all these
splits and left me with just one gigantic haskell interaction buffer.


Using `pop-to-buffer' instead of self-written logic is more robust and
future-proof. It's also the "standard" way of bringing some buffer to
the front (See cider
https://github.com/clojure-emacs/cider/blob/master/cider-interaction.el#L411
or even inf-haskell: https://github.com/the-kenny/haskell-mode/blob/master/inf-haskell.el#L205)

Ivan Lazar Miljenovic <notifications@github.com> writes:

> Can you please describe the behaviour that this implements and how it differs from the current behaviour?  I did look at #710 but don't really understand the problem (maybe I'm just used to dealing with the current behaviour).
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/haskell/haskell-mode/pull/743#issuecomment-118449705

--

gracjan:
@the-kenny: Thanks for great explanation. Merged.

Issue #744, opened 2015-07-02
Add Stack ghc feature


johnwalker:
I'd really like to be able to use `stack ghc` instead of `cabal repl`. To do this, we just need to work inside haskell-process.el, right?

gracjan:
There is a thing called `process-type` and it currently supports `cabal-repl` and `ghci`. It should be pretty easy to add `stack ghc` to the list of supported methods.

johnwalker:
OK, I'll put in a PR today.

PR #746, opened 2015-07-04
Add support for interaction with stack ghci.

johnwalker:
Add stack ghci support from #744. Also fix a defcustom typo in haskell-process-args-cabal-repl.

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -44,7 +44,7 @@ Used for locating additional package data files.")
 (defcustom haskell-process-type
   'auto
   "The inferior Haskell process type to use."
-  :type '(choice (const auto) (const ghci) (const cabal-repl) (const cabal-ghci))
+  :type '(choice (const auto) (const ghci) (const cabal-repl) (const cabal-ghci) (const stack-ghci))
   :group 'haskell-interactive)
 
 (defcustom haskell-process-wrapper-function
@@ -110,6 +110,12 @@ when showing type information about symbols."
   :group 'haskell-interactive
   :type '(choice string (repeat string)))
 
+(defcustom haskell-process-path-stack
+  "stack"
+  "The path for starting stack."
+  :group 'haskell-interactive
+  :type '(choice string (repeat string)))
+
 (defcustom haskell-process-args-ghci
   '("-ferror-spans")
   "Any arguments for starting ghci."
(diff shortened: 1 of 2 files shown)


gracjan:
Thanks.

lunaryorn:
Mmh, it seems that `haskell-process-cabal-build` (e.g. `C-c C-c`) doesn't understand `stack` yet, does it?

johnwalker:
Thats right. For now, look to the bottom of https://github.com/chrisdone/chrisdone-emacs/blob/master/config/haskell.el

I think thats the right approach for haskell-mode, too?

peti:
Has anyone figured out a good way to switch easily between stack and cabal projects? (My emacs-fu is very weak, so it's quite possible that there's a simple and obvious solution that I'm unaware of). When I set `haskell-process-type` to `auto`, stack projects aren't recognized, unfortunately.

srhb:
The only way I've found is to use a .dir-locals.el on an ad-hoc basis, which doesn't feel very nice. I'd settle for a menu when I start a haskell process, really.

gracjan:
Starting a process already asks a lot of questions and we should strive to
remove all of those.

srhb:
I should have been more specific, because you're obviously right. It's a sane default to not ask too many questions. I guess I just want the option for a menu (because when you switch between cabal sandboxes, nix and stack, there's no good solution currently.)

gracjan:
I think that the current 'auto' option has been irrevocably spoiled as
there is no way to reliably detect cabal vs stack preference.

Also note that nix can have all three possibilities: nix-ghci, nix-cabal
and nix-stack, so it actually adds yet another dimension to the whole mix.

I wonder if we somehow could get around the whole mess.

peti:
Auto-detection cannot be 100% reliable, but IMHO it can be pretty good:

1. If the top-level directory relative to the current source module contains a `stack.yaml` file, then it's Stack.
2. If that directory contains a Cabal file *and* the directory layer above contains a `stack.yaml` file that mentions the Cabal file in its `packages` section, then it's Stack.
3. If there is a Cabal file but no Stack file, then it's Cabal.
4. If everything else fails, just load ghci.

ivan-m:
@peti except that falls through in the case of "this project has `stack` support, but I don't want to use it" (e.g. cloned from upstream, but you're still using `cabal`), which is why I believe the current 'auto' option checks if there's `stack.yaml` *and* `stack` is in the `PATH`.

PR #750, opened 2015-07-06
Fix regex for package version detection

cocreature:
At least for newer ghc versions the line here looks like
```
    It is a member of the hidden package ‘xml-conduit-1.3.0@xmlco_BftDMkVAlkEFsqbeX22lfX’.
```
so only the part before the `@` should be considered.


cocreature:
Updated so that it doesn't match newlines.

Issue #724, opened 2015-06-15
Tokenizer is needed


gracjan:
Plenty of code relies on local regexes to find tokens. All of these are half-backed.

We need a common `next-token` functionality as a building block for many more advanced parts.

Reference: #648, #628, #572, #549, #450, #229. Possibly many more.

geraldus:
Good idea. It will be great work I guess.

PR #757, opened 2015-07-09
Add haskell-lexeme.el

gracjan:
Implement Haskell tokenizer, with all bells a whistles, Unicode support, all GHC special cases and stuff.

Related to #724.

### DIFF
--- a/haskell-lexeme.el
+++ b/haskell-lexeme.el
@@ -1,0 +1,223 @@
+;;; haskell-lexeme.el --- haskell lexical tokens   -*- coding: utf-8 -*-
+
+;; Copyright (C) 2015 Gracjan Polak
+
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+
+;; This file is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


gracjan:
@kuribas, @geraldus: can you have a look if you see something obvious missing here?

haskell-lexeme.el
```diff
@@ -0,0 +1,233 @@
...
+      'conid)
+     ((or (equal char ?_)
+          (member category '(Ll Lo)))
+      'varsym)
+     ((and (>= char ?0) (<= char 9))
+      'number)
+     ((member char '(?\] ?\[ ?\( ?\) ?\{ ?\} ?\` ?\, ?\;))
+      'special))))
+
+(defun haskell-utils-match-string-equal (string &optional index)
```
geraldus: would `haskell-utils.el` be the better place for this function?

geraldus:
@gracjan I keep an eye on every pull request and issue, and this is no exception.  This looks good to me, but my level of Haskell experience does not allow me to give any valuable comments on this (besides small notice I've already put).  Overall, I think this is very important for present and future development.  Good stuff.

gracjan:
Thanks for the feedback.

Is there any other test that you can think of and should be implemented?

PR #758, opened 2015-07-09
Implement haskell-forward-sexp

### LINKED ISSUE #648: forward-sexp works poorly with operator names
In the following line

    x Array.! y

place point before `x` and then use `mark-sexp`. The marked strings will be:

    x
    x Array.
    x Array.! y

This is not correct behaviour for haskell-mode, because `x Array.` is not a meaningful haskell expression while `x Array.!` would be (in the sense of an operator slice, at least): this breaks navigation of haskell code by other emacs tools that rely on standard emacs functions such as `forward-sexp`.

As a possibility, haskell-mode could potentially provide its own `forward-sexp-function` that marks expressions suitable to haskell.

Also related is that `haskell-ident-at-pos` returns either `Array`, or `Array.`, depending on where point is placed within `Array.!`, which is probably why documentation for `Array.!` is not displayed when point is on top of it. (The command given to ghci in process log is `:info Array.`, which is wrong.) Ideally, after placing point on `Array.!`, `eldoc-mode` (or `haskell-doc-mode`) should show information about the operator `Array.!`.

gracjan:
Based on tokenizer work implement `forward-sexp` that works with Haskell.

(Note that this only works forward, backward falls back to Emacs native function.)

Fixes #648.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -136,6 +136,7 @@
 (require 'haskell-complete-module)
 (require 'haskell-compat)
 (require 'haskell-align-imports)
+(require 'haskell-lexeme)
 (require 'haskell-sort-imports)
 (require 'haskell-string)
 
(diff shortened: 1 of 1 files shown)


gracjan:
@ikirill, can you check if this works for you?

PR #767, opened 2015-07-19
Random refactoring

deepfire:
Cleanups, that are preparatory for `error-overlays`, whose desirability is being questioned.


gracjan:
Thanks!

PR #769, opened 2015-07-21
New variable to allow user to enable/disable prompts on starting REPL

vlatkoB:
- Added `haskell-process-load-or-reload-p` variable with default value `t` meaning "do make prompt", so current flow is not changed unless user changes the variable to `nil`.
-  Three functions changed just to check for `haskell-process-load-or-reload-p` variable and, depending on its value, make a prompt or accept the default:
    `haskell-session-new-assume-from-cabal`

    `haskell-utils-read-directory-name`

    `haskell-session-target`

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -22,6 +22,11 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Customization variables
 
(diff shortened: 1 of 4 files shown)


gracjan:
@vlatkoB: Almost great!

This should be `defcustom`.

Also make the default to NOT PROMPT. You will make a lot people happy that way. Okay?

vlatkoB:
Done.
Travis fails because `haskell-sort-imports.el` changed in the mean time. Should you fix that, or should I?

gracjan:
Already fixed. Please rebase on top of current master and push again.

gracjan:
Thanks!

Issue #741, opened 2015-07-01
New dabbrev completions too eager


geraldus:
Currently I'm writing some GHCJS code and didn't setup interactive version yet, so I always have dabbrev completions. With 10-20 opened buffers dabbrev completion method becomes too eager already, i.e. completions takes too long. It is possible to limit dabbrev list of buffers to search completions. For example, try to lookup nearest folder contains .git, or .cabal file, set some haskell-completions-top-folder local var, and when searching dynamic abbrevs take this var in account. When session is available, search for d-abbrevs in buffers assigned to current session.

What do you think?

gracjan:
I use dabbrev by M-/ command key. I would love to have M-/ do dabbrev,
completion kept only for more context-sensitive semantic information only.

geraldus:
If you don't use **company-mode** I think you does not encounter this issue.  Here is a little [demo](https://drive.google.com/open?id=0BwLfIDb7xEV7d3Vzd1FkNWh6Z1k).  Company mode, no REPL (hence completions are always comes from dynamic abbrevs), 50 buffers opened.

geraldus:
If someone want to fix this before I find a time this might be helpful [dabbrev.el](https://github.com/emacs-mirror/emacs/blob/master/lisp/dabbrev.el).

gracjan:
@geraldus: Why do you use `dabbrev--reset-global-variables`? It is here:

https://github.com/haskell/haskell-mode/blob/a8146714e12b2af6afe1a086636dcc4ee6bb7825/haskell-completions.el#L272

As I understand `dabbrev` has some caching mechanism to speed up expansions, resetting kills this optimization. Or am I mistaken?

gracjan:
@geraldus: My point about M-/ was that `company-mode` should NOT use dabbrev as source. I envision it like this:

- `company-mode` uses only contextual sources for completions: `:complete`, `PRAGMA`, `import`
- M-/ uses non-contextual sources for completions: abbrev, dabbrev

So 'smart' stuff is for company/completion, non-smart for dabbrev. Am I mistaken?

cocreature:
@gracjan I agree if people want to use dabbrev via company mode they can still add it themselves but I am not sure what we gain by doing that from haskell mode.

geraldus:
Sometimes REPL is not running for some reason (in my case there is no GHCJSi available yet) or user just does not want to run it, that is, `:complete` command is not available too.  So how company should look up completions?

[...]

PR #772, opened 2015-07-23
Introduce completion candidates limit, minor fixes

geraldus:
Don't merge yet, I didn't tested this stuff well and want to play with this before merge.  Please review.

### DIFF
--- a/haskell-completions.el
+++ b/haskell-completions.el
@@ -249,10 +249,10 @@ Returns nil if no completions available."
                      ;; for completions list in case of module name or
                      ;; identifier prefixes
                      (haskell-completions-sync-complete-repl pfx imp)))))
-          (when (or (equal '("") lst)
-                    (eql nil lst))
-            ;; complete things using dabbrev
-            (setq lst (haskell-completions-dabbrev-completions pfx)))
+          ;; (when (or (equal '("") lst)
+          ;;           (eql nil lst))
+          ;;   ;; complete things using dabbrev
+          ;;   (setq lst (haskell-completions-dabbrev-completions pfx)))
           (when lst
             (list beg end lst)))))))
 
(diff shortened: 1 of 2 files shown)


gracjan:
Please use defconst instead of defcustom. After enough of testing we can
arrive at a constant that will make almost everybody happy. If somebody
detect a use case where that number fails then we would like to receive a
pull request so that everybody benefits.

geraldus:
@gracjan ok, I think a hundred completions is quite comfortable value.

geraldus:
@gracjan I have no idea about this

```
In toplevel form:
haskell-sort-imports.el:128:1:Error: the function `copy-list' is not known to be defined.
make: *** [haskell-sort-imports.elc] Error 1
```

gracjan:
Use `cl-copy-list` and do `(require 'cl-lib)` in the beginning of the file.

geraldus:
Ok, but I didn't touch `haskell-sort-imports.el` myself.  Will create a separate PR for this issue.

gracjan:
I've fixed copy-list problem, just rebase over current master and you should be fine.

geraldus:
Cool, thanks (:

geraldus:
I did some quick tests, with hundred candidates limit dabbrev completions are still quite slow for short prefixes (2-4 chars), but performance increases for long prefixes.  The most comfortable limit value for short prefixes for me is about 10-20 candidates.

I got some insight: what about to set candidates limit on the fly depending of prefix length, i.e. set smaller limit for short prefixes and increase it for long prefixes?

gracjan:
I've never heard about dabbrev performance issues. Are you sure we are
using it right?

geraldus:
@gracjan, maybe I've took wrong words, I'm talking about issue described in #741 (here is a [demo](https://drive.google.com/open?id=0BwLfIDb7xEV7d3Vzd1FkNWh6Z1k)).

gracjan:
@geraldus: Code looks good. It would be good to find somebody besides me and you to play with this before merge.

geraldus:
Surely, I want to make some more changes, e.g. #776.

gracjan:
@geraldus: I was thinking about this a bit more. Questions:

1. Is that the limit is a problem only with dabbrev?
2. Is there a way to DO NOTHING about dabbrev in haskell-mode and leave this task to company-dabbrev-mode?

geraldus:
@gracjan answers :D 

1. No, it is not.  **ikirill** also reported some issues about REPL completions with big projects, but in common case limit issue is more related to dabbrev/
2. I think that it is possible and I'm going to check it.  I was not aware about dabbrev customizations in haskell-mode and that company have dabbrev backend out of box also.  I will push more updates soon.

geraldus:
@gracjan I've disabled dabbrev temporarily to test existing behaviour.  I've already noticed that now completion candidates in comments are case insensitive.  I need to find and look at mentioned dabbrev customizations in haskell-mode.

gracjan:
@geraldus: I've found these custom variables:

- dabbrev-case-distinction
- dabbrev-case-fold-search
- dabbrev-case-replace

We can set the buffer-local so that case sensitivity is kept. Would that work for you?

geraldus:
@gracjan I will be able this in about one or two days.

geraldus:
There were made so many changes, I think we can close this for now, I'm still busy, apologize :(

gracjan:
No problem. Take care of your stuff, return when you are ready. haskell-mode will be around.

PR #777, opened 2015-07-26
Less questions

deepfire:


gracjan:
Please, please, please, separate pull requests for separate defcustoms.
Pretty please.

Another please: it is not without merit to not use double negatives.

deepfire:
@gracjan, sure thing, no problem!

deepfire:
@gracjan, double negatives where?

deepfire:
Closing this, in favor of https://github.com/haskell/haskell-mode/pull/779

gracjan:
And remove `haskell-ask-also-kill-buffers`. It was too hard to find for you to the point that you created your own version, it will be too hard to find for anybody else.

deepfire:
Guess I'll reuse this PR, after all : -)

deepfire:
Quoting myself on IRC:
```
01:29 <deepfire> it's ugly, ugly, ugly..
01:29 <deepfire> haskell-session and haskell-process are basically married
01:30 <deepfire> different names, same entity, in fact..
01:30 <deepfire> it needs _deep_ refactoring
```

gracjan:
It underwent one refactoring already so it is much better than it used to
be.

Note that the biggest issues are with the file named `haskell.el`,
everything else is not as painful.

gracjan:
@deepfire: Do you plan to fix this pull request or subsume with some other? Should we close or not?

deepfire:
The only reason I reopened was to keep the discussion -- not to suggest the code to be merged.

Particularly, I didn't want to lose your comment about `haskell-ask-also-kill-buffers`.

PR #778, opened 2015-07-26
More loading

deepfire:
This mimics the `ghc-mod` experience.

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -102,6 +102,11 @@ when showing type information about symbols."
   :group 'haskell-interactive
   :type 'boolean)
 
+(defcustom haskell-load-on-visit nil
+  "Whether to load the buffer on visit."
+  :group 'haskell-interactive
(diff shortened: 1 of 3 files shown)


deepfire:
My problem with this is -- I cannot figure out the reason for the `haskell-process-load-file` error in the test suite.  It works fine outside of test..

geraldus:
@deepfire maybe `(require 'haskell)` is needed in `haskell-mode.el`, because `haskell-process-load-file` is defined there.

deepfire:
@geraldus, but haskell.el already requires haskell-mode!

deepfire:
Perhaps the tests are constructed in a wrong way..

deepfire:
TBH I don't understand when haskell.el is supposed to be loaded.

deepfire:
Closing in favor of https://github.com/haskell/haskell-mode/pull/780

PR #780, opened 2015-07-26
haskell-load-on-save:  new custom

deepfire:


gracjan:
I'm torn about this one. Can someone else voice opinion if reload on safe functionality is more useful or more problematic?

deepfire:
@gracjan, it's what `ghc-mod` does, and it adds *a lot* of interactivity into the process..

geraldus:
I always think about this behaviour, one thing I can say it is always possible to save and reload manually using single keystroke (I have `C-c C-l` bind to `haskell-process-load-or-reload`, which saves and reloads file).  Auto reload on save is a bit tedious for big projects, or projects with Template Haskell, because reloading takes notable time.

deepfire:
JFYI, this is what I add in my `~/.emacs` to obtain this behavior:
```
  (defun haskell-load-on-save ()
    (when (eq major-mode 'haskell-mode)
      (haskell-process-load-or-reload)))

  (pushnew 'haskell-load-on-save after-save-hook)
```

deepfire:
Probably no reason to keep this PR open.

Issue #776, opened 2015-07-24
haskell-process-get-repl-completions result contains unnecessary item


geraldus:
The result of GHCi REPL `:complete repl` command always outputs results summary in first line, it is consists of number of showed completions, number of all possible completions and of unused part of string for which completions were searched.  E.g.

```
Prelude> :complete repl 2 "map co"
2 6 "map "
"compare"
"concat"
```
In the example above `2` is a number of returned completions, `6` is a number of all completions (`"co"` could be completed with 6 unique completions) and `"map "` is unused part of string (which was ignored, completions were searched only for `"co"` part).  In case when completion requested for single word, for example `:complete repl "ma"`, an empty string is returned as unused part always.

Currently, `haskell-process-get-repl-completions`<sup>[1][src]</sup> returns mentioned unused part of input string as first element in resulting list:

```el
(defun haskell-process-get-repl-completions (process inputstr)
  "Perform `:complete repl ...' query for INPUTSTR using PROCESS."
  (let* ((reqstr (concat ":complete repl "
                         (haskell-string-literal-encode inputstr)))
         (rawstr (haskell-process-queue-sync-request process reqstr)))
    (if (string-prefix-p "unknown command " rawstr)
        (error "GHCi lacks `:complete' support (try installing 7.8 or ghci-ng)")
      (let* ((s1 (split-string rawstr "\r?\n" t))
             (cs (mapcar #'haskell-string-literal-decode (cdr s1)))
             (h0 (car s1))) ;; "<cnt1> <cnt2> <quoted-str>"
        (unless (string-match "\\`\\([0-9]+\\) \\([0-9]+\\) \\(\".*\"\\)\\'" h0)
          (error "Invalid `:complete' response"))
        (let ((cnt1 (match-string 1 h0))
              (h1 (haskell-string-literal-decode (match-string 3 h0))))
          (unless (= (string-to-number cnt1) (length cs))
            (error "Lengths inconsistent in `:complete' reponse"))
          (cons h1 cs))))))
```

[...]

PR #781, opened 2015-07-27
Add optional candidates limit arg, remove unused item from result

geraldus:
+ Add options argument to `haskell-process-get-repl-completions` setting
completion candidates limit
+ Remove unused item from resulting candidates list.  This item is not
completion itself but unused part of input string (see issue #776 for
more details).

### DIFF
--- a/haskell-process.el
+++ b/haskell-process.el
@@ -286,23 +286,30 @@ This uses `accept-process-output' internally."
     (haskell-process-queue-flush process)
     (car-safe (haskell-command-state cmd))))
 
-(defun haskell-process-get-repl-completions (process inputstr)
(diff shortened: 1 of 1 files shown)


geraldus:
@gracjan updated

vwyu:
It seems that this change introduces a problem. Here is what I get by typing some import expressions after λ> and pressing tab:

1) Pressing tab after "import Co" makes it "import CoCo".
2) Pressing tab after "import Con" makes it "import ConControl."

It looks like part of the string is repeated. I would expect the following behavior:

Expected 1) Pressing tab after "import Co" shouldn't change it.
Expected 2) Pressing tab after "import Con" should make it "import Control."

Issue #791, opened 2015-08-02
Can't compile from within emacs if $EMACS is not set.


wkunkel:
The makefile uses the `$EMACS` environment variable to find emacs if it is present, or falls back to `emacs` if it is not. However, emacs seems to set the `$EMACS` environment variable to `t` for all of its subshells if it does not already exist. The result is that running make from an emacs shell or with M-x compile fails with `/bin/sh: t: command not found`.

gracjan:
Is there a way to find out the real path to currently running Emacs?

wkunkel:
Good question! I can get the parent PID from a shell with bash's $PPID

wkunkel:
Sorry about closing/reopening, my browser did something weird.

gracjan:
@wkunkel: I see that your fix actually fixes the issue, so merged. Thanks.

PR #792, opened 2015-08-02
Fix not being able to compile from within emacs

wkunkel:
I opened an issue about this earlier and it turned out to be really easy to fix: https://github.com/haskell/haskell-mode/issues/791.


gracjan:
Thanks.

Issue #667, opened 2015-05-24
Fix haskell-customize


gracjan:
To remove `haskell-customize` do:

- [x] add `;;;###autoload` to every `defcustom` and `defface`
- [x] remove `haskell-customize` function

wkunkel:
Shouldn't `;;;###autoload` also be added to every `defgroup`?

gracjan:
As far as I understand documentation defgroup and autoload do not go well
together.

wkunkel:
Where in the documentation have you seen that? [This] (http://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html) seems to indicate that autoload has special behavior for `defgroup`, just like for `defcustom`.

gracjan:
Great! I must have read documentation for an older version of Emacs. In
that case can you provide autload definitions pull request?

wkunkel:
Removing `haskell-customize` seems to work for the most part, except for one snag: the group hierarchy is not visible through the `customize` or `customize-browse` interfaces until the file has been loaded. Groups can still be customized directly with `customize-group`, but they don't appear as subgroups until they are directly accessed or something in the same file is loaded.

gracjan:
Seems correct to me. As I checked customize groups in other modes are not
populated until lisp files are loaded.

Do other modes do something more?

wkunkel:
I don't know if other modes do something more. It seems like it would be a decrease in functionality from the current customize mechanism, however.

gracjan:
In that case lets do it.

Can you also change documentation that points to haskell-customize? Write
about M-x customize-group RET haskell.

hvr:
Fyi, adding autoload cookies to all defcustoms seems to be the likely cause for #1043

gracjan:
Closing because #1043 is fixed.

PR #794, opened 2015-08-03
Fix haskell-customize step #1

rosekunkel:
Add autoload cookies to every `defcustom`, `defface`, and `defgroup`, to complete the first step in https://github.com/haskell/haskell-mode/issues/667.


RespectTH123 (approved):
Chang

PR #795, opened 2015-08-04
haskell-load-opens-repl:  new custom

deepfire:


deepfire:
@gracjan, what about this one -- is it controversial?

There's no other way to obtain this functionality, AIUI..

gracjan:
defcustom is always super controversial.

Just think about the documentation that we will need to write. Or the
amount of testing will this need.

Is there a way to avoid defcustom?

deepfire:
It appears it's the wrong place to put the control in, I need to fix it -- as it stands, it prevents the REPL popping up in non-load situations too.. which is exactly the kind of problem I've been fixing with my recent pull request..

deepfire:
Closing this pull request, until the above issue is fixed..

deepfire:
@gracjan, about the way to avoid the defcustom.

There are two different possible workflows: REPL-based, and non-REPL-based.

One benefits from having the REPL-on-load, for the other it's a distraction.

deepfire:
@gracjan, maybe we can have a single global defcustom, which chooses the workflow, instead of the details:
1. REPL-based: no load-on-save/load-on-visit, no -Wall by default, but repl-on-load
2. IDE-based: load-on-save/visit, -Wall by default for warning overlays, no repl-on-load

What do you think?

geraldus:
@deepfire sounds good, but I think it's a bit complicated.

deepfire:
@geraldus, any ideas on how to make it simpler?

deepfire:
@geraldus, or do you think the IDE-based (think `ghc-mod`) workflow isn't worth supporting?

gracjan:
We should support both workflows. At the same time we cannot handle
testing, documentation and questions that will come if we do not make it
simpler.

deepfire:
Does the proposed merging of defcustoms into a single switch constitute as a step towards the desired simplification?

gracjan:
I think so. The devil is in the details and we need to spend more thinking
time if we are actually doing this right.

deepfire:
Quoting myself from IRC:
```
23:34 <deepfire> I think I see a possible way to go about it 
23:35 <deepfire> 1. introduce pseudo-defcustoms -- variables that ought to be
                 defcustom, but are not actual defcustoms
23:35 <deepfire> 2. introduce a command to enable/disable the IDE mode, by
                 flipping all the pseudo-defcustoms at once
23:36 <deepfire> 3. introduce a single defcustom that controls this call during
                 haskell-mode init
23:36 <deepfire> this way we get:
23:36 <deepfire> - modularity
23:36 <deepfire> - tiny user fuzzing attack surface
```

geraldus:
You rocks (:

deepfire:
What if I drop the last commit in this branch -- the one introducing the `defcustom`?

I think the remaining commits are still useful on their own.

gracjan:
The proper way to get the functionality you want is described in #807. There we can fix the issue at the core of the problem.

Functionality you want can be achieved by differentiating interactive calls vs non-interactive ones. If load or reload is triggered by `C-c C-l` or `C-c C-r` then bring the window, otherwise leave the decision to bring the window to higher level code that called load or reload.

gracjan:
The fixing of the problem at hand moved to #807, discussion of architectural change moved to #809.

PR #800, opened 2015-08-06
Remove haskell-customize function

wkunkel:
Step 2 of https://github.com/haskell/haskell-mode/issues/667


gracjan:
Thanks, that was liberating.

PR #808, opened 2015-08-08
Scylla charybdis united

deepfire:
`haskell-session` <- `haskell-process`

What works:
1. Loading a module and seeing the error overlays -- provided that the ~/.emacs has `s/haskell-process/haskell-session/g` applied.

Problems:
1. This is the purely mechanical phase of the merge.  The bulk of normalization will come later.
2. This changes the `defcustom` names.

### DIFF
--- a/doc/haskell-mode.texi
+++ b/doc/haskell-mode.texi
@@ -575,10 +575,10 @@ keybindings in the @code{haskell-mode-map} keymap with the respective
   '(progn
     (define-key haskell-mode-map (kbd "C-x C-d") nil)
     (define-key haskell-mode-map (kbd "C-c C-z") 'haskell-interactive-switch)
-    (define-key haskell-mode-map (kbd "C-c C-l") 'haskell-process-load-file)
+    (define-key haskell-mode-map (kbd "C-c C-l") 'haskell-session-load-file)
     (define-key haskell-mode-map (kbd "C-c C-b") 'haskell-interactive-switch)
-    (define-key haskell-mode-map (kbd "C-c C-t") 'haskell-process-do-type)
-    (define-key haskell-mode-map (kbd "C-c C-i") 'haskell-process-do-info)
+    (define-key haskell-mode-map (kbd "C-c C-t") 'haskell-session-do-type)
+    (define-key haskell-mode-map (kbd "C-c C-i") 'haskell-session-do-info)
     (define-key haskell-mode-map (kbd "C-c M-.") nil)
     (define-key haskell-mode-map (kbd "C-c C-d") nil)))
 @end lisp
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -121,9 +121,9 @@ If I break, you can:
   3. General config:    M-x customize-mode
   4. Hide these tips:   C-h v haskell-session-show-debug-tips")))))))
 
-(defun haskell-commands-process ()
+(defun haskell-commands-session ()
   "Get the Haskell session, throws an error if not available."
-  (or (haskell-session-process (haskell-session-maybe))
+  (or (haskell-session-maybe)
       (error "No Haskell session/process associated with this
       buffer. Maybe run M-x haskell-session-change?")))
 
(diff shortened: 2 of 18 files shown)


gracjan:
How to minimize risk of potential problems before merging?

deepfire:
I guess it just needs to be tested for a while..  I'm ok with keeping it up-to-date with current master for a month or so.

deepfire:
Affected `defcustom`s:
```
[deepfire@andromedae:~/src/haskell-mode]$ git log -p -G 'defcustom haskell-process-' orig/master..HEAD | grep defcustom\ haskell-process
-(defcustom haskell-process-load-or-reload-prompt nil
-(defcustom haskell-process-type
-(defcustom haskell-process-wrapper-function
-(defcustom haskell-process-path-ghci
-(defcustom haskell-process-path-cabal
-(defcustom haskell-process-path-stack
-(defcustom haskell-process-args-ghci
-(defcustom haskell-process-args-cabal-repl
-(defcustom haskell-process-args-stack-ghci
-(defcustom haskell-process-do-cabal-format-string
-(defcustom haskell-process-log
-(defcustom haskell-process-show-debug-tips
-(defcustom haskell-process-suggest-no-warn-orphans
-(defcustom haskell-process-suggest-hoogle-imports
-(defcustom haskell-process-suggest-hayoo-imports
-(defcustom haskell-process-hayoo-query-url
-(defcustom haskell-process-suggest-haskell-docs-imports
-(defcustom haskell-process-suggest-add-package
-(defcustom haskell-process-suggest-language-pragmas
-(defcustom haskell-process-suggest-remove-import-lines
-(defcustom haskell-process-suggest-overloaded-strings
-(defcustom haskell-process-check-cabal-config-on-load
-(defcustom haskell-process-prompt-restart-on-cabal-change
-(defcustom haskell-process-auto-import-loaded-modules
-(defcustom haskell-process-reload-with-fbytecode
-(defcustom haskell-process-use-presentation-mode
-(defcustom haskell-process-suggest-restart
```

gracjan:
Is there a way to split this somehow, for example rename a group of
functions, pull this in, rename some more, pull that in?

deepfire:
The problem is shared state.. untangling it piece-by-piece isn't going to be pretty -- certainly much harder than the suddenly-it's-the-same-thing approach.

deepfire:
Interestingly, I think it's also going to become harder to understand, if done piecemeal.

gracjan:
You might be right. Still, we need to do some proper engineering here and not break stuff for our users. Any idea how to do this?

geraldus:
I suppose we cannot rely on regression tests now, because interactive stuff is missing tests, isn't it?

deepfire:
The obvious breakage is `defcustom`s, which I can rename back into what they were -- no problem.

The hidden problems -- I think only a testing period can solve that.

deepfire:
The problem is finding willing testers, of course : -)

gracjan:
So you have replaced a software problem with a social problem :)

deepfire:
Well, the property that we want to test -- equality of computable functions -- is mathematically undecidable..

deepfire:
One engineering approach that comes to mind could have been based on a test suite..

gracjan:
It is hard to test functions that sometimes ask for user input. We need to refactor that feature out.

gracjan:
This depends on session internals:

https://github.com/chrisdone/haskell-flycheck/blob/master/haskell-flycheck.el

deepfire:
@gracjan, https://github.com/deepfire/haskell-flycheck/commit/41e4c2d596332878004fdc7dde415215a65bb007

deepfire:
@gracjan, not tested, though.

deepfire:
Also, it'll need a second update, after we perform the non-mechanical normalization part.

gracjan:
Are there any other projects depending on haskell-mode that use session
internals?

deepfire:
@gracjan, I just thought.. `haskell-flycheck` is something that is made obsolete by error-overlays.

gracjan:
@deepfire: `haskell-flycheck` uses a nice mechanism to let flycheck know about the errors. Should we use it ourselves?

What I mean is this:
- when `C-c C-l` is run collect errors and warnings in a session local variable
- when `flycheck` triggers `:start` function it will repeatedly run the flycheck error report callback with information taken from the run

gracjan:
@deepfire: We will need either 3 people (beside you) confirming that this actually works OR a comprehensive test suite.

I would say the latter option is more realistic to achieve.

deepfire:
@gracjan, I see, of course.

gracjan:
This also depends on session internals:

https://github.com/juiko/company-ghci

gracjan:
@deepfire: Do you think there are three people that tested this out by now? Can they confirm it works?

deepfire:
@gracjan, I didn't reach out to people about testing this..

gracjan:
I do not see a method how to push this forward. Too risky, we have to live with current architecture. Thanks for trying.

Issue #787, opened 2015-07-30
Auto-completion doesn't work right in new build


mathlover2:
I just got the new version of haskell-mode. The auto-complete function does not seem to work now.

Say one wants to write "sequence" and has written "sequ". Auto-completion now yields "sequsequence" instead of "sequence".

geraldus:
Oh, that's very sad.  Do you have some special package for haskell-mode or just plain `auto-complete`?  I've pushed some changes related to completions stuff recently, but it does not affected the format of result, there were just improvements of the way completions are collected.  I'm using `company-mode` personally and have no much time to test `auto-complete` separately.  Can you help with this?

mathlover2:
I checked the haskell-completions.el file, and nothing seems to be wrong when I invoke its functions in Emacs when in Haskell mode. As this is the file which you had changes merge into, I don't think the problem is there.

joehillen:
I've had this happen with company-mode. I haven't been able to make it happen consistently. I'll take a screenshot next time it does.

vwyu:
I don't use company-mode or any other auto completion modes. Here is what I get by typing some import expressions after λ> and pressing tab:

1) Pressing tab after "import Co" makes it "import CoCo".
2) Pressing tab after "import Con" makes it "import ConControl."

It looks like part of the string is repeated. I would expect the following behavior:

Expected 1) Pressing tab after "import Co" shouldn't change it.
Expected 2) Pressing tab after "import Con" should make it "import Control."

It looks like this issue might be related to removal of the unused string (h1) from completions (cs) in haskell-process-get-repl-completions function of haskell-process.el (related commit: f0466a2). Adding the unused string (h1) to the completion list (cs) solves this issue for me. But I am not sure if this change breaks something else instead. 

```elisp
(let ((cnt1 (match-string 1 h0))
              (h1 (haskell-string-literal-decode (match-string 3 h0)))
          (unless (= (string-to-number cnt1) (length cs))
            (error "Lengths inconsistent in `:complete' reponse"))
          (cons h1 cs)))
```

gracjan:
This seems to be fixed by #810.

PR #810, opened 2015-08-10
Add unused string in completions as well.

vwyu:
It seems that the issue #787 might be due to the "unused string" (h1) being not in completions list (cs). This commit reverts part of the commit f0466a2 back to the earlier version, where the unused string (h1) is placed at beginning of the completions list (cs).


gracjan:
For now accepted. There is something fishy going on with these functions though. Their return value should follow this protocol:

```
(start end collection . props)
```

http://www.gnu.org/software/emacs/manual/html_node/elisp/Completion-in-Buffers.html

PR #814, opened 2015-08-15
Better tests for ‘haskell-indentation-mode’

mrkkrp:
Here is better helper for testing, a macro to eliminate boilerplate and more dense tests. Three last tests are strange (commented). Also, I don't know if contents of current line should affect found indentations or collection of indentation columns is completely defined by context set in previous lines.

Criticism is welcome. I haven't silenced failing tests for now.

Note that it's always clear which "sub-case" failed because I compare positions too, so they are shown when something fails, this also helps ensuring that point can be placed in specified position.

### DIFF
--- a/tests/haskell-indentation-tests.el
+++ b/tests/haskell-indentation-tests.el
@@ -1,429 +1,509 @@
-(require 'ert)
-(require 'haskell-indentation)
-(require 'haskell-mode)
+;;; haskell-indentation-tests.el --- tests for indentation module
+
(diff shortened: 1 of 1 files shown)


mrkkrp:
@gracjan, Here, I fixed those issues. In general feel free to change tables of indentations, because you may want to give user more alternatives or something, just make sure it makes sense and it's consistent.

gracjan:
Thanks. This is good as it is. Now we need more meaningful tests.

PR #816, opened 2015-08-18
Suppress byte-compile warning

syohex:
`'` should be used rather than `#'` in `make-obsolete` and `#'` causes byte-compile warning.  You can get following warning by following steps.
1. byte-compile `haskell.el`
2. load `haskell.elc`
3. byte-compile `haskell.el`

```
haskell.el:67:2:Warning: `haskell-process-completions-at-point' is an obsolete
    function (as of June 19, 2015); use
    `haskell-completions-sync-completions-at-point' instead.
```

### DIFF
--- a/haskell.el
+++ b/haskell.el
@@ -64,8 +64,8 @@
             nil
             t))
 
-(make-obsolete #'haskell-process-completions-at-point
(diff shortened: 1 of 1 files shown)


gracjan:
Nice catch. Thanks.

PR #818, opened 2015-08-19
haskell-load: add haskell-goto-first-error

da-x:
I usually start to fix stuff from top to bottom. This helps finding the first error.

### DIFF
--- a/haskell-load.el
+++ b/haskell-load.el
@@ -345,6 +345,12 @@ correspondingly-named overlay properties of OVL."
 	(t
 	 (message "No further notes from Haskell compiler."))))
 
+(defun haskell-goto-first-error ()
+  (interactive)
+  (haskell-goto-error-overlay
+   (first-overlay-in-if 'haskell-check-overlay-p
+			(buffer-end 0) (buffer-end 1))))
+
 (defun haskell-goto-prev-error ()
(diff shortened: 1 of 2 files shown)


gracjan:
Hi,

Good function. Is it connected to anything?

da-x:
I set up my `haskell-goto-*` bindings to match those of my `flycheck` setup, which are different from the `haskell-interative-mode` defaults, which are `M-n` and `M-p` for `next` and `prev` respectively ever since 2012. I defer on this new binding for @chrisdone to decide.

gracjan:
We can't burden our users to somehow discover that this exists and then bind the keys themselves.

For this PR to be merged we need:
- [ ] user interface element where it is bound (keybinding)
- [ ] documentation in `haskell-mode.texi`
- [ ] a unit test case

da-x:
Added a binding. However, considering that `haskell-mode.texi` is very lacking around the `haskell-process` mode, and that the unit tests of `haskell-process` around UI are practically nonexistent, it's a bit over my head right now to extend on those. Do what you will :)

bergey:
Finishing this PR is on my list, too.

gracjan:
Good. Thanks.

Issue #817, opened 2015-08-18
Support testing


vlatkoB:
There seems to be no support for testing in `haskell-mode`. I tried this:

```elisp
(defun haskell-process-cabal-repl-test ()
  (interactive)
  (setq haskell-process-args-cabal-repl '("test"))
  (haskell-process-load-or-reload)
  (setq haskell-process-args-cabal-repl '("--ghc-option=-ferror-spans" "--with-ghc=ghci-ng"))
 )
```
and can run tests.

What would be the proper way to implement it? Maybe also add the function to run a test at point?

ivan-m:
Sure there is: `haskell-session-change-target` and then enter in the name of the test you want to edit/evaluate/build/etc.

That said, @mrkkrp's [`ebal`](https://github.com/mrkkrp/ebal) has some support for running test suites.

vlatkoB:
Thanks for pointing me to `haskell-session-change-target`. I do not recall I read about it in docs.

`ebal` looks interesting.

gracjan:
@vlatkoB: Can you try to describe the workflow in document `haskell-mode.texi`?

vlatkoB:
I added a small [description] (https://github.com/haskell/haskell-mode/pull/819). (Not sure why PR is not displayed here.)

Was this what you meant?
If yes, please review it. English is not my mother tongue, and see if style suits you.

vlatkoB:
Would be nice if `haskell-session-change-target` would display `ido` choice of sections, instead of typing the section name.

ivan-m:
Maybe grab the code that does that from https://github.com/mrkkrp/ebal ?

vlatkoB:
Just created a [PR](https://github.com/haskell/haskell-mode/pull/828) with a new function that enables choosing a target from available targets in `.cabal` file..

PR #819, opened 2015-08-19
Describe haskell-session-change-target for test & bencmark, #817

vlatkoB:
Small description of the `haskell-session-change-target` function and its usage for testing and benchmarking.


gracjan:
Thanks.

PR #828, opened 2015-08-22
Choose a REPL session target from list

### LINKED ISSUE #817: Support testing

vlatkoB:
New function `haskell-session-choose-target` enables choosing from available targets list from `.cabal` file.
It is just a wrapper around `haskell-session-change-target` which can still be used to select a target directly, by specifying it as an arg.

### DIFF
--- a/doc/haskell-mode.texi
+++ b/doc/haskell-mode.texi
@@ -601,32 +601,24 @@ TODO/WRITEME
 @findex haskell-session-change-target
 @vindex haskell-interactive-mode-hook
 
-With @code{haskell-session-change-target} you can change the section
-(defined in project's @file{.cabal} file) the interactive REPL session is
-started with.
+With @code{haskell-session-change-target} you can change the target for
+REPL session.
 
-After the session is started, you can switch the target for
+
+After REPL session started, in @code{haskell-interactive-mode} buffer invoke the
+@code{haskell-session-change-target} and select from available targets for
 
 @cindex testing
 - Testing
 
-In @code{haskell-interactive-mode} buffer invoke the
-@code{haskell-session-change-target} and enter the name of the test you
-wish to perform, i.e. ``test''.
-Answer ``yes'' to restart the session.
-
 @cindex benchmarking
 - Benchmark
 
-In @code{haskell-interactive-mode} buffer invoke the  @code{haskell-session-change-target}
-and enter the name of the benchmark you wish to perform, i.e. ``bench''.
-Answer ``yes'' to restart the session.
-
 - Executable
 
-In @code{haskell-interactive-mode} buffer invoke the  @code{haskell-session-change-target}
-and enter the name of the executable you wish to work with.
-Answer ``yes'' to restart the session.
+- Library
+
+Answer ``yes'' to restart the session and run your tests, benchmarks, executables.
 
 
 TODO/WRITEME
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -444,6 +444,21 @@ OTHER-WINDOW use `find-file-other-window'."
 (defun haskell-cabal-section-data-start-column (section)
   (plist-get section :data-start-column))
 
+(defun haskell-cabal-enum-targets ()
+  "Enumerate .cabal targets."
+  (let ((cabal-file (haskell-cabal-find-file)))
+    (when (and cabal-file (file-readable-p cabal-file))
+      (with-temp-buffer
+	(insert-file-contents cabal-file)
+	(haskell-cabal-mode)
+	(let (matches)
+	  (goto-char (point-min))
+	  (haskell-cabal-next-section)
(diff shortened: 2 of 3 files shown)


gracjan:
@vlatkoB: almost great!

I think `(interactive)` can do what you need it to do, look here: http://stackoverflow.com/questions/2382524/adding-completion-to-interactive

- [ ] please embed this functionality in an already existing function `haskell-session-change-target`
- [ ] look through `haskell-cabal.el` and see if there is anything you could use
- [ ] put the cabal section enumeration function in `haskell-cabal.el`
- [ ] `-mapcar` is not good, please find something elisp specific that fulfills the purpose

vlatkoB:
Almost all requests fulfilled, except the using something from `haskell-cabal.el`.
Actually, my first attempt used exclusively `haskell-cabal` functions:
```elisp
(defun haskell-cabal-extract-targets ()
  "Enumerate targets from .cabal file."
  (interactive)
  (let (matches)
    (goto-char (point-min))
    (haskell-cabal-next-section)
    (while (not (eobp))
      (push (haskell-cabal-section-value (haskell-cabal-section)) matches)
      (haskell-cabal-next-section))
    (reverse matches)))
```
However, it works correctly only when invoked inside `.cabal` buffer. When called with `with-temp-buffer`, results are incorrect. 
```elisp
(defun test-in-temp ()
  (interactive)
  (let ((cabal-file (haskell-cabal-find-file)))
    (when (and cabal-file (file-readable-p cabal-file))
      (with-temp-buffer
    (insert-file-contents cabal-file)
    (print (haskell-cabal-extract-targets))))))
```
Either I'm blind to some obvious mistake, or functions in `haskell-cabal` are somehow dependent on "real" `.cabal` file or something.

I asked on IRC for help, but no answer. I'm very fresh in elisp, and can't solve it without going too deep, for what I currently have no time.

In other words, any help is appreciated.

gracjan:
I just tested the two functions that you've put here and those work (well, almost). I get all the answers plus some additional things. 

- [ ] please use `haskell-cabal-source-bearing-sections` to only find things that can be a target

vlatkoB:
If you mean the functions in the post above, the result is not correct. Not only there are more results than needed, but the result for test is prefixed with `-suite`.
`("-version:  >=1.20" "-type:     Simple" "biblio" "biny" "-suite test1" "-suite test2" "benchy")`

The result of the function should be the same when in the buffer with `.cabal` file, and when using `with-temp-buffer`. Input is the same, so should be the output. I don't trust it. Too much IO. :-) 
At least until I find out what's causing the differences.

Can the author of `haskell-cabal` maybe comment the issue?

gracjan:
If you look into `haskell-cabal-mode-syntax-table` you will find:

```el
    (modify-syntax-entry ?- "w"  st)
```

That says that in buffers where `haskell-cabal-mode` is enabled `-` char should match `\\w` regexp. If buffer is not in `haskell-cabal-mode` minus char is not word character so it does not match.

vlatkoB:
It is working now. Changed `major-mode` of temp buffer to `haskell-cabal-mode`.
Can you test on your side?

gracjan:
Works for me. Please clean up two more things and we are good to go.

vlatkoB:
What `defvar` are you talking about?
There was one in previous commit to satisfy tests, but none in this commit. In this commit I merged two functions in above post and replaced the previous `haskell-cabal-enum-targets` function, as the commit message says.

gracjan:
There are two commits in this pull request. defvar is in line 449 of the first one.

vlatkoB:
I'm a little lost. Could use a little help.

gracjan:
It was me who was lost, everything is fine.

I have one more request though: can you make it use its own separate history variable? Have a read here:

http://www.gnu.org/software/emacs/manual/html_node/elisp/Minibuffer-Completion.html

There is an argument called `history` and it defaults to `minibuffer-history`. This is suboptimal, better is to use special history variable for cabal targets, lets say `haskell-cabal-targets-history`.

Can you do this?

vlatkoB:
Can you check?

gracjan:
@vlatkoB: Can you fix these two small things? Your pull request is valuable and it is a shame that it wait in queue.

vlatkoB:
@gracjan They were fixed the day you asked for them. `haskell-cabal-targets-history to symbol` above.

gracjan:
Fixes #817

PR #834, opened 2015-08-25
Fix hoogle lookup from local servers

sheyll:
Use a better way to check if the hoogle process is alive: Instead of
checking if the corresponding buffer-name is an alive process, store the
process handle in a global variable and check that.

The problem is that when using 'hoogle-lookup-from-local' without any
hoogle server running, a new local hoogle server is started, but the
code fails to recognize that server is running and instead repeatedly
reports that the server is not running offering to start it.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -859,27 +859,33 @@ is asked to show extra info for the items matching QUERY.."
 (defvar hoogle-server-process-name "emacs-local-hoogle")
 (defvar hoogle-server-buffer-name (format "*%s*" hoogle-server-process-name))
 (defvar hoogle-port-number 49513 "Port number.")
+(defvar hoogle-server-process nil "The process handle of the local hoogle server.")
 
 (defun hoogle-start-server ()
   "Start hoogle local server."
   (interactive)
-  (unless (hoogle-server-live-p)
-    (start-process
-     hoogle-server-process-name
-     (get-buffer-create hoogle-server-buffer-name) "/bin/sh" "-c"
-     (format "hoogle server -p %i" hoogle-port-number))))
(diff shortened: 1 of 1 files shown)


gracjan:
@sheyll: Have you had time to look into this?

sheyll:
@gracjan is this ok now?

gracjan:
Great, thanks.

The functionality around hoogle and hayoo could use refactoring and be merged. Can you look into improving code there?

jb55:
I think you need to pass `--local` to hoogle otherwise it uses `file://` links and browsers refuse to open those. Should I PR this?

jb55:
Not strictly related to this PR but I'm curious to why it's not there already because local hoogle links are broken for me otherwise...

gracjan:
I do not understand... What file links?

jb55:
Sorry, when running hoogle with `-d /path/to/docs`, which is the case for auto-generated nix environments:

`hoogle "$COMMAND" -d /nix/store/smcqcq332zwb9g6xqvdz4zsxgk3bm253-hoogle-local-0.1/share/doc/hoogle "$@"`

Without the `--local` flag, links to docs on the search results page are `file:///path/to/docs` instead of `http://localhost/file/path/to/docs` which prevents them from working due to security mechanisms in browsers.

I was just wondering if it makes to add `--local` while we're at it, or at least an option to customize the hoogle server arguments.

gracjan:
I still do not get it.

If hoogle produces wrong kind of links that make hoogle unusable then that
is a bug in hoogle and should be fixed there, shouldn't it?

jb55:
All I need is the ability to pass `--local` to `hoogle server`. I'll start a new issue.

PR #837, opened 2015-08-28
Fixed `haskell-indentation-newline-and-indent`

rietmann:
The "melpa-stable" release was correct, but I guess the recent
refactoring effort created a regression. The new version did a newline
before storing the current column/indentation. This means that <RET> 
would newline and be indented all the way at the left. This fixes
the behavior while following the new code style. (save column first,
then newline, then reindent)


gracjan:
I've merged this as is because it seems good. @rietmann, can you as a follow up task provide an ERT test case so that we do not have the same issue ever again?

PR #838, opened 2015-08-29
Use standard dabbrev completions for company mode

cocreature:
I left the options in, since they might still be useful for people who want to enable dabbrev themselves. Honestly I think even if @geraldus there is not a good case for enabling dabbrev by default since it is not specific to haskell-mode. If people like it they can enable it themselves (in fact company-mode does this by default).


gracjan:
What options did you leave in?

cocreature:
We set some dabbrev options in [ghci-script-mode.el](https://github.com/haskell/haskell-mode/blob/master/ghci-script-mode.el#L46) and in [haskell-mode.el](https://github.com/haskell/haskell-mode/blob/master/haskell-mode.el#L688).

PR #858, opened 2015-09-09
Use customize-variable

### LINKED ISSUE #857: There is no option package-archives
I installed fresh emacs24-nox, ran it, wanted to install haskell-mode, due to instructions:

> To install haskell-mode you need to add a package archive repository that distributes haskell-mode. Execute

> M-x customize-option RET package-archives

But there is not option `package-archives`. Maybe correct would be to advice to `customize-variable RET package-archives`?

k-bx:
`customize-option` doesn't work, since you also need to add `(require 'package)` to your `~/.emacs`, and requiring users to do so makes these instructions too complex, so usage of `customize-variable` looks best.

Resolves #857

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ To install `haskell-mode` you need to add a package archive repository that dist
 `haskell-mode`. Execute
 
 ```
-M-x customize-option RET package-archives
+M-x customize-variable RET package-archives
 ```
 
 and add


gracjan:
There is something fishy going on because:

;;;###autoload
(defalias 'customize-variable 'customize-option)

;;;###autoload
(defun customize-option (symbol) ...)

That means your patch does not really solve the problem in question.

PR #859, opened 2015-09-10
Quick Installation

k-bx:
Previous instructions didn't work for me because they required `(require 'package)` to be inside `.emacs` first.

I propose going this "put this to your .emacs" route. Not sure if it's much better, but at least it should work for clean emacs out of the box, unlike previous instructions. Adding one more item to previous instructions to put "require 'package" seemed to make this "quick installation" too big.

What do you think?


gracjan:
Well, I guess this just works... Still it is fishy to get into so many gotchas just to run haskell-mode.

k-bx:
@gracjan totally agree. Would be great to find a better solution.

PR #865, opened 2015-09-11
Don't spuriously set mark on xref push

### LINKED ISSUE #863: Jumping with M-. within the current buffer marks a region
If I use `M-.` to jump to a symbol definition, and that definition is in the same file, then when the point is moved, the region between the previous and new positions is selected. I would expect the region not to be activated.

I can patch this in due course, but I wanted to record the issue here.

purcell:
This fixes #863, whereby jumping to the definition of a symbol in the
same file was incorrectly activating the mark and region. This commit
also simplifies the original code, since the initial position marker can
simply be passed to xref-push-marker-stack. (The haskell-compat shim for
that function is also fixed here accordingly.)

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -364,12 +364,7 @@ position with `xref-pop-marker-stack'."
         (haskell-mode-handle-generic-loc loc)
       (call-interactively 'haskell-mode-tag-find))
     (unless (equal initial-loc (point-marker))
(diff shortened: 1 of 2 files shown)


gracjan:
Nice. So I have learned something about xref-push today. Thanks.

purcell:
:+1:

PR #866, opened 2015-09-12
Use backspace to move to the previous indentation position in haskell-indentation

gergoerdi:
This patch restores the "backspace on first colum unindents" functionality. 

I wasn't able to find any existing tests for interactive indentation commands so I didn't bother.

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -132,7 +132,19 @@ positions are allowed."
   (let ((keymap (make-sparse-keymap)))
     (define-key keymap (kbd "RET") 'haskell-indentation-newline-and-indent)
     (define-key keymap (kbd "<backtab>") 'haskell-indentation-indent-backwards)
+    (define-key keymap (kbd "<backspace>") 'haskell-indentation-unindent-or-delete-backwards)
     keymap))
+
+(defun haskell-indentation-at-beginning-of-line ()
+  (and
+   (not (= (current-column) 0))
+   (= (haskell-indentation-current-indentation) (current-column))))
+
+(defun haskell-indentation-unindent-or-delete-backwards ()
+  (interactive)
+  (if (haskell-indentation-at-beginning-of-line)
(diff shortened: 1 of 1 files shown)


gracjan:
Test cases are crucial for quality engineering. We do not want your functionality to accidentally stop working in a couple of weeks annoying everybody.

We try to roughly follow the process described in [Pull request checklist](https://github.com/haskell/haskell-mode/wiki/Pull-request-checklist).

Please add a test case.

gergoerdi:
OK, do you have an example of a test that tests interactive editing, that I can base my test cases on? I'd like the test to enter text, position the cursor, press backspace, and then check what has been deleted and where the cursor ends up.

gracjan:
haskell-indentation-tests.el has relevant code in the macro. You can
extract this into other function.

gergoerdi:
I've looked at `haskell-indentation-tests.el` before, but that one seems geared more towards testing where the indentation points are determined to be, than testing navigation between them.

gracjan:
That is the kind of tests that are done currently. Can you add different
kind of tests that will test interactive functions? We have more than one
interactive function to test.

PierreR:
@gergoerdi very useful fix indeed ! I don't think I feel confident enough to add the test case (as you said there is nothing alike)

gracjan:
I've added example tests in #994. You need to create a similar test.

gracjan:
I've been running with this commit for a while and I find it very confusing. Is this really what people want?

bergey:
I don't feel strongly either way.  I often want `haskell-indentation-indent-backwards`, I almost never want to delete indentation not to an indent point.  But `shift-tab` works fine for me.

gracjan:
I've run with this functionality for a month now and so far learned to fear by Backspace. This is not going in.

bergmark:
That echoes my experience with such functionality elsewhere.

PR #880, opened 2015-09-16
Ensure file reload when using -fbytecode

cmears:
With `haskell-process-reload-with-fbytecode` set to `t`, we first compile with object code, then switch to bytecode, touch the file and issue a `:reload`. However, if the `.o` file generated in the first compilation is still considered fresh -- because it has the same timestamp as the just-touched source file -- the reload doesn't do anything.

This patch makes the touch move the source code's timestamp into the near future, so that it's always newer than the object code. This way the reload will always recompile it. The mechanism is pretty ugly and a bit of a hack, but seems to work on Linux and OS X.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -143,29 +143,17 @@ If I break, you can:
 Restores -fobject-code after reload finished.
 MODULE-BUFFER is the actual Emacs buffer of the module being loaded."
   (haskell-process-queue-without-filters process ":set -fbyte-code")
-  (haskell-process-touch-buffer process module-buffer)
-  (haskell-process-queue-without-filters process ":reload")
-  (haskell-process-queue-without-filters process ":set -fobject-code"))
-
-;;;###autoload
-(defun haskell-process-touch-buffer (process buffer)
-  "Query PROCESS to `:!touch` BUFFER's file.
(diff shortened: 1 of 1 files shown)


gracjan:
Good idea, implementation needs to be better. Please do the following:

- [ ] use `set-file-times` described here http://www.gnu.org/software/emacs/manual/html_node/elisp/Changing-Files.html#index-set_002dfile_002dtimes
- [ ] use `current-time`
- [ ] use time calculations as described here http://www.gnu.org/software/emacs/manual/html_node/elisp/Time-Calculations.html#Time-Calculations
- [ ] check if the whole thing does not produce unwanted interaction with Emacs track of buffer time described here http://www.gnu.org/software/emacs/manual/html_node/elisp/Modification-Time.html
- [ ] use 2 seconds, as this is what lowest resolution filesystem (MSDOS-FAT) uses

That path should make this functionality much more manageable.

cmears:
This implementation with touch is not so nice. I've just changed the pull request to a different method, which uses ghci's `:load *<file>` form, which asks it to ignore any object files and interpret the module. This avoids any updating of file timestamps. What do you think of this way?

gracjan:
Nice. Much better than what was here before.

PR #881, opened 2015-09-16
Fix hoogle query to fpcomplete

psibi:
The fpcomplete hoogle service has been transferred to Stackage servers.
This patch makes sure that the hoogle query works with fpcomplete.


gracjan:
Thanks.

Issue #884, opened 2015-09-17
parse-error "Expecting then" "Expecting else" DoAndIfThenElse in indentation


k-bx:
For this code:

```haskell
main :: IO ()
main = do
    if True
    then putStrLn "True"
```

when you press enter after "True" – you get an error:

let: No catch for tag: parse-error, "Expecting then"

k-bx:
Added a PR with a failing test

gracjan:
This seems to be an instance of [DoAndIfThenElse](https://ghc.haskell.org/trac/haskell-prime/wiki/DoAndIfThenElse).

k-bx:
@gracjan yeah, seems like it. All compilers support it now with no flags needed, so would be nice for haskell-mode to support it, since it feels not right when code compiles, while indentation fails to work.

gracjan:
The way this extension is defined with this semicolons out of nowhere is
super strange. How to make indentation use this syntax will require some
ingenuity.

gracjan:
Actually we have this reported already as #321.

PR #886, opened 2015-09-17
Expecting then

k-bx:
Issue #884


k-bx:
Warning: this test really hurts for my terminal for some reason

gracjan:
Note that this is extended syntax for DoIfThenElse stuff.

gracjan:
Thanks. Now we need somebody to fix this :)

Issue #882, opened 2015-09-17
"Unexpected response from haskell process" when loading in new session


cmears:
Sometimes when you press C-c C-l to start a new session and immediately load a file, I get "Unexpected response from haskell process." I think this is due to a race condition.

When the new session starts, it issues a few commands to set the prompt, set verbosity and change directory. If the empty response to these commands comes late, they appear to haskell-mode to be in response to the load command, which doesn't match the expected output.

That is, you get this timeline:

send "cd"
send "load"
receive empty response (from the cd)
receive response "compiling, loading modules, etc." (from the load)

It doesn't always happen, (un)fortunately. You can see it in action in the log below -- the last few lines are the interesting bit. I've tried fiddling around with a few things but I don't know the right way to fix it.

```
("Starting inferior stack GHCi process using stack" "haskell" nil "stack" "ghci" "--ghc-options=-ferror-spans")

-> :set prompt "\4"

-> Prelude.putStrLn ""

-> :set -v1

<- Run from outside a project, using implicit global config

   Using resolver: lts-2.15 from global config file: /home/chris/.stack/global/stack.yaml

   

<- Configuring GHCi with the following packages: 

   

<- GHCi, version 7.8.4: http://www.haskell.org/ghc/  :? for help

   

<- Loading package ghc-prim ... 

<- linking ... done.

   Loading package integer-gmp ... linking ... done.

   Loading package base ... linking ... done.

   

<- Prelude

<- > 

-> :cd /home/chris/elisp/others/haskell-mode/

<- 

   

<- 

-> :load "/home/chris/elisp/others/haskell-mode/Test.hs"

<- 

<- [1 of 1] Compiling Test             ( /home/chris/elisp/others/haskell-mode/Test.hs, /home/chris/.stack/global/.stack-work/odir/Test.ghcio )
```

gracjan:
As I understand how command queue is supposed to work is that it should
follow strict request-response-request-response order. Pending commands
should be queued.

If what you say is true then that would explain some other transient bugs
reports.

cmears:
I investigated this some more and found that it should only be a problem at session start -- otherwise, all commands are run in the strict request-response order.  See pull request #887 for a fix.

PR #887, opened 2015-09-17
Fix out-of-order session startup.

cmears:
Previously the responses to the startup commands could be delayed,
arriving so late that they appeared to be in response to later
commands (e.g. a :load).  By setting the prompt last, only one prompt
marker will be sent for the whole startup procedure.

Should fix issue #882.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -87,9 +87,11 @@ You can create new session using function `haskell-session-make'."
     :state process
 
     :go (lambda (process)
-          (haskell-process-send-string process ":set prompt \"\\4\"")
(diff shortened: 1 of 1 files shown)


gracjan:
I've merged this because it is better than the version before.

The correct way to do this is to use ghci multi line capabilities. Can you look into this?

cmears:
Thanks for merging this. I'm happy to look more into this, but I'm not sure what you mean. If I understand it right, ghci's multiline input is for splitting one big command over several lines, but what we want is to send several independent commands in one line. I don't think ghci supports that.

PR #891, opened 2015-09-20
Fix tag generation to not read in lockfiles or other undesirables.

code-crimes:


gracjan:
Why is this even going though GHCi shell and interactive mode? TAGS could be generated directly.

gracjan:
Thanks.

code-crimes:
Haha yeah I was wondering that as well, good point :smile:. Thanks for the merge.

gracjan:
If you have time and skill, can you look into untangling this mess and
generating TAGS directly? I bet there is some premade ready code out there
in the internets that does exactly what we want.

code-crimes:
Sure, I'm not the most fluent in elisp, but I'd be happy to give it a shot! By directly you just mean shelling out to find + hasktags instead of through this ghci weirdness, correct?

gracjan:
Yes, exactly.

PR #893, opened 2015-09-23
some simple commands to send arbitrary text to a haskell process.

abelew:
This provides 3 small functions to send a line, region, and paragraph to a running haskell.
My commit follows:

Also a keybinding similar to ESS's for sending a paragraph (C-c C-n) which
I don't think conflicts with anything existing unless you get your lines crossed with haskell-interactive-mode in which case it would conflict with
'haskell-interactive-mode-prompt-next.

I have so far tested it out only on my simple playing around scripts, but it seems to work as advertised and evaluate arbitrary code from my emacs buffer.
The current workflow I am using to play in haskell is:
1.  Open my .hs file
2    meta-x run-haskell
3.   :l file.hs
4.  Make some changes and hit ^C^N
5.  Watch the haskell prompt to see what I did right/wrong.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -201,6 +201,7 @@ be set to the preferred literate style."
     (define-key map (kbd "C-c C-v") 'haskell-mode-enable-process-minor-mode)
     (define-key map (kbd "C-c C-t") 'haskell-mode-enable-process-minor-mode)
     (define-key map (kbd "C-c C-i") 'haskell-mode-enable-process-minor-mode)
+    (define-key map (kbd "C-c C-n") 'inferior-haskell-send-paragraph)
     map)
   "Keymap used in Haskell mode.")
 
--- a/inf-haskell.el
+++ b/inf-haskell.el
@@ -402,6 +402,36 @@ If prefix arg \\[universal-argument] is given, just reload the previous file."
     (setq inferior-haskell-seen-prompt nil)
     (comint-send-string proc str)))
 
+(defun inferior-haskell-send-region ()
+  "Send an arbitrary region to an existing inferior haskell."
+  (interactive "*")
+  (save-excursion
+    (let ((haskell-proc (inferior-haskell-process))
+          (region-string (buffer-substring (region-beginning) (region-end))))
+      (inferior-haskell-send-command haskell-proc region-string))))
+
+(defun inferior-haskell-send-paragraph ()
+  "Send the current paragraph to a running inferior haskell."
+  (interactive "*")
+  (save-excursion
+    (let ((haskell-proc (inferior-haskell-process)))
+      (forward-paragraph)
+      (setq region-end (point))
+      (backward-paragraph)
+      (setq region-start (point))
+      (setq region-string (buffer-substring region-start region-end))
+      (inferior-haskell-send-command haskell-proc region-string))))
+
+(defun inferior-haskell-send-line ()
+  "Send the current line to a running haskell process."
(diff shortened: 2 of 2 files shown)


cocreature:
How is this supposed to be useful? GHCi requires you to use a different syntax, e.g. you won't even be able to send a simple function over and define it there. Also reloading parts is significantly less useful with haskell because we have lexical binding so changing a definition won't change the code that previously used that definition.

I am not saying that there are no use cases, but I'd like to see some.

abelew:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

  Greetings,
  I found similar reasoning when poking on google to see why
send-paragraph etc didn't already exist.  I don't honestly understand
ghc/ghci well enough yet to properly address the argument -- but that is
exactly my point; having the ability to send a line to a haskell process
in order to test my code allows me to learn the language that much
faster so that I will be able to understand the argument.

  For example:
  I started doing the 99 problems in lisp in haskell.  The very first
problem is quite simple: get the last element of a list.

  After reading a few tutorials, I wrote:

mylast :: [a] -> [a]
mylast [] = []
mylast [a] = [a]
mylast ([a:b]) = mylast ([b])

and sent it to ghci -- which promptly told me I have no idea what I am
doing until I amended it to remove the square brackets from [a:b] and
[b].  Once I did that, and reloaded, I added

let mylist = [1,2,3,4,5]
let answer = mylast(mylist)
answer
let firstlist = [1,2,3,4,5,6]
let secondlist = [2,3,4,5,6,7]
mylast(firstlist)
mylast(secondlist)

to my buffer and hit control-c control-c (which I bound to
inferior-haskell-send-paragraph) and was able to see that it worked.
To me, this was a win.  Once I was confident I understood what was
going on, I commented it and added a main = do block.  Admittedly this
is stupid, but so am I at this point in my understanding of the language
.
  By the same token, I spent some time reading the documentation and I
still don't understand why I can't send any arbitrary expression to
ghci/hugs; it seems being able to do that would help someone stupid
like me learn the language.

  I hope you have a great day,
 -trey


On 09/23/2015 04:30 PM, Moritz Kiefer wrote:
> How is this supposed to be useful? GHCi requires you to use a 
> different syntax, e.g. you won't even be able to send a simple 
> function over and define it there. Also reloading parts is 
> significantly less useful with haskell because we have lexical 
> binding so changing a definition won't change the code that 
> previously used that definition.
> 
> I am not saying that there are no use cases, but I'd like to see 
> some.
> 
> --- Reply to this email directly or view it on GitHub: 
> https://github.com/haskell/haskell-mode/pull/893#issuecomment-14271976
7
>
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iF0EAREIAAYFAlYDI2kACgkQzgGyqJcfxquQPAD4maCfdXx/+r2cg6OBCQNsWhMn
wwDZgr6mn0QJ4CJf5wD/Zvp7tkMA1WIc2kf5cnZrncmldq4J0jludpcCWnePyQ8=
=UkB5
-----END PGP SIGNATURE-----

cocreature:
You are aware that the code that you write this way will *only* work in GHCi right? You can't compile that using GHC which is why I don't get the motivation behind doing that.

The only advantage to pressing C-c C-n over pressing C-c C-l (of course you can change those shortcuts) I can think of is that C-c C-l clears the state in ghci while your approach doesn't. IMHO not being able to load normal haskell code significantly outweighs the benefits here.

PR #904, opened 2015-10-01
Fix switching to the presentation buffer

chrisdone:
This fixes a regression. `haskell-presentation-present` is supposed to pop-to-buffer when there is none and we're not in the buffer, and if we are inside the buffer it's supposed to re-use this same buffer. This fixes it. `(with-current-buffer ...)` confuses the value of `major-mode`.


gracjan:
Good stuff!

PR #913, opened 2015-10-05
Mobile friendly README.md

gracjan:


ivan-m:
Wait, the GitHub mobile interface (which I can't use since GH doesn't like my phone :@) actually shows it better if you explicitly use an image rather than using Markdown? :o

gracjan:
No, the point is that it shows first three paragraphs. So we need to put logo inside the header.

Issue #923, opened 2015-10-10
haskell-process changes the directory which causes loaded modules to be unloaded


11Kilobytes:
When you use stack-ghci as your haskell-process-type, I think that haskell-process ought not to change directories after loading the repl, since all modules of the project are loaded correctly. 

There is no advantage to changing directories for stack-ghci and cabal-repl.

11Kilobytes:
There is no advantage to changing directories for stack-ghci and cabal-repl.

gracjan:
I think you are right. Can you provide a pull request?

11Kilobytes:
Sure, I'll do that right now.

PR #925, opened 2015-10-10
Don't change directory when using stack-ghci.

11Kilobytes:
Fixes https://github.com/haskell/haskell-mode/issues/923.


gracjan:
Thanks.

PR #927, opened 2015-10-12
Add rather baroque indentation test which cause parse failure

Shimuuar:
I added another test case which causes parse failure. It's quite complicated but couldn't reduce it further


gracjan:
Good catch. Thanks.

gracjan:
Parser happily skips over the comma and never sees it. Bummer.

PR #928, opened 2015-10-12
Yet another parser failure

Shimuuar:
Gotta catch em all I guess


gracjan:
This one has a workaround. But there is still something fishy going on with handling 'do' inside a list or inside a record.

PR #929, opened 2015-10-13
Reloading: haskell-process-file-loadish needs the current buffer as an argument

sleepomeno:
Currently nil is the argument for the buffer called in haskell-process-reload-file such that I always get "Haskell process command errored with: (wrong-type-argument stringp nil)" in the mini buffer.


gracjan:
Good catch, thanks!

PR #933, opened 2015-10-16
Honour curly braces in target sections

### LINKED ISSUE #534: `haskell-session-change-target` should list available targets
There's already a few functions that deal with finding sections, but none to find *all* sections and return them in a list (e.g. for use with ido).

vlatkoB:
Accept target sections written with curly braces, i.e. remove them from targets listing.

Fixes #534.


gracjan:
Have you though about creating a couple of test cases for this functionality?

It seems way more complicated than it was expected initially.

vlatkoB:
Well, the change itself is actually not complicated. Split target name on space, take head and replace with "library" if head is "empty".
If `haskell-cabal-section-value` would return the correct value (without a brace), function could be reverted to the previous commit.
This is a just workaround.

Also, not sure if `haskell-cabal-section-value` should return something (like "library") for `library` section (which has no value), or leave it empty, as is now.
If that would be the case, even the last commit could be removed.

What say you?

gracjan:
Writing some unit tests for this doesn't take much time and is pretty is
task. There is no excuse to not do this.

vlatkoB:
I made a `test-data` directory with sample `.cabal` file, added `haskell-cabal.el` to  `ELFILES` in `Makefile`, and made `haskell-cabal-tests.el` in `tests` with test function
```elisp
(ert-deftest haskell-cabal-enum-targets-1 ()
  "Test enumerating .cabal targets."
  (with-temp-buffer
    (haskell-cabal-mode)
    (setq default-directory (expand-file-name "./test-data/"))
    (should (equal '("library" "test-1" "bench-1" "bin-1")
		   (haskell-cabal-enum-targets)))))
```
`ert` in Emacs passes test, but `make check` from shell does not.
Little help...

vlatkoB:
Thanks.

vlatkoB:
Test is updated in the fix as well. Or do you mean to test the whole process, with `haskell-session-change-target`?

gracjan:
I thought about having a couple different xxx.cabal files, but now I see that you have covered all cases wit one cabal file. Seems good enough.

Can you quash all 4 commits into 2: one with functionality and one with the tests? We will merge when you do this.

gracjan:
hmm, branch has conflicts. Can you try to rebase again on top of master?

gracjan:
Thanks.

PR #934, opened 2015-10-16
Add failing tests for multiline string literals

Shimuuar:
Also add one passing test for multiline tuples. It fails however with older haskell-mode so it makes sense to add it as regerssion test


gracjan:
Good stuff, keep those coming.

gracjan:
@Shimuuar: Want to have a laugh look here: #944. :)

PR #935, opened 2015-10-16
Add ido/helm interface for haskell-interactive-mode-history

tsukimizake:
Hello.

`haskell-interactive-mode-history-next/previous` requires too many key-pushes when the user wants to trace an old history.
I think it would be better if these kind of functions are provided as options.

Thanks.

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -38,6 +38,7 @@
 (require 'ansi-color)
 (require 'cl-lib)
 (require 'etags)
+(require 'helm)
 
 (defvar haskell-interactive-mode-history-index)
 (make-variable-buffer-local 'haskell-interactive-mode-history-index)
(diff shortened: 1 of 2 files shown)


gracjan:
Is this connected to anything or just two functions?

tsukimizake:
These two functions are self-contained, except for using `haskell-interactive-mode-history`.
Is it what you mean?

tsukimizake:
By the way, I'm not sure about the way of adding dependency to elisp package. The travis-cl fails. Is there something wrong?

gracjan:
Is this functionality connected to a keybinding or Emacs hook or menu item?

tsukimizake:
I see. As they are declared as interactive functions, they can be called by `M-x ...`.
No keybind yet. Of course they can be bound by `define-key`, but I didn't conceive good keybind.

purcell:
`helm` is too large an additional dependency for `haskell-mode`.

In general you shouldn't be deciding completion methods for users: just use `completing-read`, and this will be overridden by users of `helm` or `ido-mode` as necessary.

tsukimizake:
Ah, that's the point. That's what I should have done. Thanks for advice.
I'll be back!

PR #939, opened 2015-10-16
Remove Unicode input method

### LINKED ISSUE #938: Remove Unicode input method
Remove Unicode input method.

gracjan:
Fixes #938.


gracjan:
Some people actually like this, so lets keep it for a while.

Issue #861, opened 2015-09-10
Indentation mode needs to be able to recover from parser errors


arybczak:
If there is a parse error while trying to determine indentation (e.g. after pressing Enter), the mode completely breaks down. In the past it dropped to the beginning of the next line, currently the Enter key just stops working. This makes the whole mode unreliable to use.

My proposition is that it should be ensured that if the indentation breaks, the mode will still be more or less usable (at least that one is able to go to the next line) until it's able to work again (by getting out of the context it can't parse possibly, I'm not sure how it's implemented). As for the indentation level, I think the best course of action would be to just stay at the indentation level it's at when going to the next line (which is the right course of action the majority of time).

Shimuuar:
Yes this major usability problem. IMO enter should insert newline if indenter fails. There are always some weird corner case or plain invalid haskell which will cause parser errors so haskell-mode should fall back on some simpler heuristic

gracjan:
Closing in favor of #962.

PR #943, opened 2015-10-17
Insert newline even when parse fails

Shimuuar:
Fix for #861 

This is very hacky and rely on fact that expression return string
when parse fails and integer otherwise. I'd love to hear
suggestion on how to implement it better

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -259,19 +259,24 @@ negative ARG.  Handles bird style literate Haskell too."
   (if (haskell-indentation-bird-outside-code-p)
       (progn
         (delete-horizontal-space)
-        (newline))    
-    (catch 'parse-error
-      ;;  - save the current column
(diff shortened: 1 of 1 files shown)


gracjan:
Does not look good enough. And even if it did we absolutely need a unit
test case fur such functionality.

gracjan:
Hi @Shimuuar, can you work on a test case first and later on we will also fix the code?

Shimuuar:
Well I don't expect this patch to be merged. It's a hack after all. Submitting bad patch is good way to get things moving. 

Unit test is difficult in this case. We need reliable way to trigger parse error and ideal indenter shouldn't raise parser error ever. Any parse error could disappear in the future. We need two tests. One to ensure that parse error happens and another check that function-with-overly-long-name works correctly in case of parse error.

I'll do it. But I don't have much time so it'll be slow.

gracjan:
Closing in favor of #962.

PR #947, opened 2015-10-18
Remove dynamic indentation showing

gracjan:
haskell-indentation had the option to put overlay underlines at positions eligible as indentation points. This was supposed to help usability and debugging of indentation algorithm. It practice is was slow, distracting and a test suite helps development much better.

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -40,8 +40,6 @@
 ;; TODO `haskell-indentation-find-indentation' — fix it, get rid of "safe"
 ;; version
 
-(require 'hl-line)
-(require 'syntax)
 (require 'cl-lib)
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


PierreR:
Is it what `hi2` provided at the time before it was integrated within `indentation` ?

If so I personally don't mind the removal but many users might actually like it ... Why not just disable the option by default ?

gracjan:
Reasons:
- it is 170 lines less to support and maintain
- some potential contributors feel intimidated by the size of `haskell-indentation`, this might get us below their scare limit
- the first question after integrating hi2 was 'how to disable this?'
- second question was 'it interacts badly with my other modes'
- it was super slow, now has caching and throttling, but it is still slow
- it is disabled by default meaning that nobody will find the switch to turn it on
- we have zero unit tests for this functionality
- we have zero documentation for this functionality

PierreR:
Somehow all these questions should have come up when `hi2` was integrated. I was not aware the integration had been a complete failure.

Don't want to blame anybody. I know there is a shortage of human resource and I understand the decision. Thanks for the quick response (I haven't forgot I need to contribute myself and write some docs about spacemacs ;-)

gracjan:
@PierreR: My response was about `haskell-indentation-show-indentations` **only** as it was a nice attempt and we have learned more in practice. Therefore I do not understand your comment in your first paragraph.

I do understand and endorse your second paragraph, though :)

PR #958, opened 2015-10-26
Remove horizontal whitespace based smart indentation mode haskell-simple-indent

gracjan:
*Rationale*:
- `haskell-indentation-mode` became reliable enough for general use
- `haskell-indentation-mode` lost its magic and now has a nice predictable behavior
- simple indent despite its name is also a 'smart' mode with surprising behavior
- people think that simple indent does constant space indentation and this is not true, simple indent tries to be smart
- it relies on whitespace between tokens and Haskell defines token boundaries in a different way
- community effort should be more focused


ivan-m:
Any particular reason why?  There might be people that prefer not to use a "smart" indentation mode and just manually indent/un-indent.

gracjan:
@ivan-m: I updated my comment to state the rationale for the move. Are you using this mode yourself or were you stating a theoretical question?

ivan-m:
Just theoretical.  I'm happy enough with your justifications (I didn't realise it tried to be smart, and thought it just did indent/un-indent).

PR #959, opened 2015-10-26
Remove haskell-checkers.el

gracjan:
haskell-checkers used to provide:
- 'scan' interface (limited), as far as we know nobody is using that
- 'hlint' interface, superseeded by much better flycheck-hlint in flycheck project


ivan-m:
I don't know about "much better": I still tend to prefer the `hs-lint` package that's shipped with hlint as it can update the errors for you (which I don't think flycheck can do); I see that haskell-checkers also has support for doing this kind of update.

gracjan:
I guess nobody knew that haskell-checkers has this functionality (only for hlint). Haskell Mode should strive to provide [Complete functionality](https://github.com/haskell/haskell-mode/wiki/Complete-functionality).

Flycheck is moving in the direction described in https://github.com/flycheck/flycheck/issues/530.

Proposition:
- you keep using the external package for now
- community helps flycheck to provide the functionality

I do not think we can afford keeping unused haskell-checkers around knowing that there are better options cooking elsewhere.

ivan-m:
I did a quick search on GitHub, and it seems that those people who - like myself - `require` haskell-checkers aren't using it (I think the template `.emacs` setup for haskell-mode had it at one stage even though it wasn't used, so a few of us copy-pasted it in).

Possibly document the option of getting hs-lint more clearly for people who might want that functionality, especially  since it seems flycheck won't be getting it any time soon.

bbatsov:
Removing this makes sense.

PR #961, opened 2015-10-26
Remove scc functions

gracjan:


ivan-m:
Please don't do this; I find these functions quite useful.  I tend to use it with:

```emacs
(defun toggle-scc-at-point (&optional arg)
  "Insert or kill (with universal-argument) an SCC annotation at
point."
  (interactive "P")
  (if (equal arg nil)
      (haskell-mode-insert-scc-at-point)
    (haskell-mode-kill-scc-at-point)))

(define-key haskell-mode-map (kbd "C-c C-s") 'toggle-scc-at-point)
```

gracjan:
Well, so part of the functionality is in your personal config file. We need to decide which direction to move, either all of it ends up in haskell-mode or all of it ends up in your config file.

I've written up what is a 'Complete functionality' in the wiki https://github.com/haskell/haskell-mode/wiki/Complete-functionality. Quoting here inline:

1. functionality is available via a **defined keybinding**
2. functionality is available via a **menu option**
3. functionality is documented in **official manual**
4. functionality does not require additional elisp code from user
5. functionality works reasonably well with **default settings**
6. code implementing the functionality has enough **quality** to not be a burden on Haskell Community
7. code is sufficiently **general** to serve many users
8. **error handling** is in place and is user friendly
9. code is reasonably well covered by **unit tests**

How do we get there with SCC annotations?

ivan-m:
I think these functions already match criteria 7 at least (though now that I think about it, it might be possible to more automatically determine that if it's in an SCC declaration to remove it).

I can copy these definitions back into my config if you decide that it shouldn't belong in haskell-mode; I just think that utility functions like these are useful enough for a range of people that they might as well be defined once rather than everyone re-defining them.

Maybe some kind of `haskell-auxiliary.el` file?

gracjan:
@ivan-m, listen, two unit tests, two paragraphs in documentation, an official keybinding and scc functionality stays with haskell-mode. Can you contribute that?

ivan-m:
I'll have a go this weekend.

bergey:
I've started working on the necessary docs and cleanup.  I still need to write unit tests, and I suppose figure out how to add menu entries.

https://github.com/bergey/haskell-mode/tree/scc

bergey:
@gracjan I think that this can be closed now that #986 is merged.

PR #963, opened 2015-10-27
Make hasktags command customizable

MichaelXavier:
Because of some some pretty odd reasons in my setup, I found myself needing to customize the find/hasktags command that haskell-mode runs to generate a TAGS file. Some valid reasons for this being customizable could include having different versions of either hasktags or GNU find. Went ahead and loaded this into my own editor and it appears to work fine.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -33,6 +33,7 @@
 (require 'haskell-presentation-mode)
 (require 'haskell-utils)
 (require 'highlight-uses-mode)
+(require 'haskell-customize)
 
 ;;;###autoload
 (defun haskell-process-restart ()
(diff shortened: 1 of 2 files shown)


gracjan:
Good idea, but we need to improve the execution.

If you look through recent changes to this piece of code you can't find a discussion resulting in the following observation: 'find tags' should never go through haskell-process, it should be invoked directly, outside of cabal repl. (I would link to that discussion if I had time to find it.)

Can you please first help with untangling this functionality and then we get proper defcustoms in place? Thanks.

gracjan:
I found the discussion in #891. My comment updated accordingly.

gracjan:
Closing in favor of #964.

PR #965, opened 2015-10-27
Recognize top level constructs in indentation

gracjan:
Keep top level constructs like 'data', 'import', 'class', 'instance' at top level.


ivan-m:
Just to check/make sure: I believe it's valid for `data`, etc. to be indented as long as *everything* is indented: does this change support that?  (My GHC isn't working atm so I can't verify this myself.)

gracjan:
haskell-indentation never supported any other indentation level than 0.
This change does not undo this limitation.

2015-10-28 11:15 GMT+01:00 Ivan Lazar Miljenovic <notifications@github.com>:

PR #971, opened 2015-10-30
[Doc] Add stack related configuration for compilation

PierreR:


PierreR:
Another try which is a kind of a compromise

While reading the manual, I sometimes found it difficult to read because a lot of information are buried in the text. That's the reason I kept the elisp snippet.

Let me know if that was a bad idea ;-)

gracjan:
I'd love to reformat the whole section into an enumerated list of customization variables and describe their meaning.

PR #973, opened 2015-11-01
Extract hoogle/hayoo code into haskell-hoogle.el and tidy up

purcell:
- Moved existing code to fresh module
- Fixed checkdoc warnings
- Renamed hoogle-* to use consistent haskell-hoogle-* prefix


gracjan:
I like this. Why does it not build?

purcell:
I think the branch got temporarily deleted & re-pushed while the build was running. I've restarted the build and expect it to run through successfully this time. :-)

PR #984, opened 2015-11-09
Add hayoo choice as part of hoogle url.

psibi:


gracjan:
Good idea. What do we do with all the code that is specific for hayoo and does not seem to be useful anymore?

psibi:
Honestly, I would like them to be removed from the codebase. But I think old time haskell-mode emacs users, may use them.

gracjan:
Can you submit another pull request that removes no longer used stuff? We will announce it, wait proper time, remove it if nobody raises objections.

PR #986, opened 2015-11-11
Clean up SCC code

bergey:
Following discussion at https://github.com/haskell/haskell-mode/pull/961, I've tried to make the SCC functions (which I use) more pleasant to use and to maintain.  Please let me know if you want things done differently.

I wrote a new function to toggle SCC's, so only one keybinding is necessary.  I refactored the existing code to avoid spurious errors about "can't remove SCC" when adding an SCC.  I kept the old functions around for backwards compatibility, but I don't really care if they go away.

Along the way I found a bug, which I think I've fixed and adequately described in the commit messages.

### DIFF
--- a/doc/haskell-mode.texi
+++ b/doc/haskell-mode.texi
@@ -278,6 +278,15 @@ and available packages.
 @image{anim/company-mode-import-statement}
 @end ifhtml
 
+@section Profiling and Debugging support
+
+When profiling code with GHC, it is often useful to add
+@uref{https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/profiling.html#cost-centres,
+cost centres} by hand.  These allow finer-grained information about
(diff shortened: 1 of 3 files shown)


gracjan:
This is really good!

Please use [`make-obsolete`](http://www.gnu.org/software/emacs/manual/html_node/elisp/Obsolete-Functions.html) for the functions that are no longer useful, we will remove them a couple of releases from now.

bergey:
Thanks for pointing out `make-obsolete`.  I didn't know how to do that in elisp.

bergey:
Ah, whoops, I didn't realize the third argument to `make-obsolete` is mandatory.  Maybe this will pass CI.

gracjan:
We will see if there is anybody bitching about `C-c C-s` taken for something else :)

PR #987, opened 2015-11-12
Remove hayoo related commands. 

psibi:
All the different services for hoogle are now placed under the customization variable `haskell-hoogle-url`.


psibi:
Note that there are other hayoo related stuff in `haskell-customize.el`, `haskell-commands.el` and `haskell-load.el` which I haven't removed because they are being used in those components.

gracjan:
I kind of like hayoo. Would be better to integrate hoogle and hayoo together.

PR #992, opened 2015-11-14
Use haskell-process-load-file, not -load-or-reload

### LINKED ISSUE #807: Refactor load-or-reload
User interface around load or reload needs a bit of work:

- [ ] Bind `C-c C-l` directly to `haskell-process-load-file`
- [ ] Rename `haskell-process-reload-file` to `haskell-process-reload`
- [ ] Bind `C-c C-r` directly to `haskell-process-reload`
- [ ] Make sure that in case of a lacking session error message are helpful
- [ ] Mark `haskell-process-load-or-reload` as obsolete
- [ ] Mark `haskell-reload-p` as obsolete

bergey:
- Rename haskell-process-reload
- keybindings for -process-load-file & -process-reload
- remove references to -process-load-or-reload
- mark -process-load-or-reload obsolete

closes #807

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -605,8 +605,8 @@ default (please follow GHCi-ng README available at URL
 
 \\<haskell-interactive-mode-map>
 To make this function works sometimes you need to load the file in REPL
-first using command `haskell-process-load-or-reload' bound to
-\\[haskell-process-load-or-reload].
+first using command `haskell-process-load-file' bound to
+\\[haskell-process-load-file].
 
 Optional argument INSERT-VALUE indicates that
 recieved type signature should be inserted (but only if nothing
--- a/haskell.el
+++ b/haskell.el
@@ -41,7 +41,8 @@
 
 (defvar interactive-haskell-mode-map
   (let ((map (make-sparse-keymap)))
-    (define-key map (kbd "C-c C-l") 'haskell-process-load-or-reload)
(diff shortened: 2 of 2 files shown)


bergey:
Apparently emacs byte-compiler warns about using obsolete variables in obsolete functions.  This branch dosen't compile with `make` at the moment.  I think there are two options:
- don't mark `haskell-reload-p` obsolete until `-process-load-or-reload` is actually removed
- don't promote warnings to errors in the Makefile

Which would you prefer?

bergey:
Also, to my surprise, `haskell-process-reload` starts a new session if there isn't one and loads the whole package.  So I haven't taken any action towards "Make sure that in case of a lacking session error message are helpful".

gracjan:
@bergey, Ugh, the obsolete problem is annoying. Currently we mark a function obsolete after is has been defined. So during the definition is not considered obsolete and therefore it is not expected to call other obsolete functions. I wonder if using http://www.gnu.org/software/emacs/manual/html_node/elisp/Declare-Form.html would circumvent this problem because it marks a function obsolete while the functions is being defined.

bergey:
I just rebased onto master and stopped marking the variable obsolete.  I put a note on my calendar a year from today to follow up by removing the obsolete functions and marking the variable obsolete.  I'm still happy to change that date or address this some other way if you prefer.

gracjan:
LGTM, and do not worry about making stuff obsolete, we will revisit depending on how fast the project moves forward.

PR #995, opened 2015-11-21
Add case for haskell-debug-parse-module regex to parse

codygman:
I tried debugging a simple Haskell program in haskell-mode from spacemacs and got an error from this regex not matching. This change makes it work.

### DIFF
--- a/haskell-debug.el
+++ b/haskell-debug.el
@@ -713,9 +713,9 @@ variances in source span notation."
 For example:
 
 X                ( /home/foo/X.hs, interpreted )
-
+Main             ( /home/foo/X.hs, /home/foo/X.o )
 "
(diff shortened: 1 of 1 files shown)


gracjan:
Can we get some tests with this change?

codygman:
@gracjan Sure, no problem. I couldn't find where tests go though. I also did see anything about tests in CONTRIBUTING.md.

gracjan:
Create a new test file called tests/Haskell-debug-tests.el.

gracjan:
I see that this is better than it used to be, but still quite wrong: no national letters, no other character legal in file names, no spaces. I guess we will revisit when somebody actually hits the problem.

PR #996, opened 2015-11-21
Remove duplicate filename in Makefile

codygman:
This duplicate filename was throwing this warning:

Makefile:114: target 'check-haskell-cabal' given more than once in the same rule


gracjan:
I see those files were supposed to be listed alphabetically, and then somebody broke the order.

codygman:
I wasn't sure what had happened, I just wanted to silence the error ;)

gracjan:
It happened before you or me came to this project.

PR #997, opened 2015-11-21
Remove haskell-indentation-parse-error

### LINKED ISSUE #861: Indentation mode needs to be able to recover from parser errors

### LINKED ISSUE #962: Remove possibility of 'Illegal token' from haskell-indentation
There are two places where 'Illegal token' can be triggered in haskell-indentation:

1. In `haskell-indentation-with-starter`: the case there seems to be when a lexical structure expect a closing token/keyword and the token is not present:

```el
(defun haskell-indentation-with-starter (parser &optional end where-expr?)
 ...
 (cond 
   ...
      (end (haskell-indentation-parse-error
             "Illegal token: %s" current-token))))))
```
The expected token at this point is one of closing ')', ']', '}'. I do not see why we could not eat everything up to a closing token, maybe treating layout as taking priority.

2. In `haskell-indentation-parse-to-indentations`: top level problem.

It seems to be a variation on the same, as top level list is a layout delimited list.

Solution seems to be:
1. React to leading keywords only `if`, `let`, `import`, `case` etc.
2. React to following keywords only in context of a leading keyword, so `then` and `else` only as part of `if`. 
3. Allow partial statements, that is `if` that is not followed by `then`

### LINKED ISSUE #969: Can't edit a buffer while resolving merge conflict because `Invalid Token` error.
Hi,

Indentation-mode can't handle a file with git conflict marks in it.

Regards

### LINKED ISSUE #970: C pragma and `Illegal token: else`
Hi,

```haskell
{-# LANGUAGE CPP #-}
module Test where

foo :: Int -> Int
#ifdef FOO
foo = succ
#else
foo = pred
#endif
```
gets me `Illegal token: else` message when I hit ENTER.

Thanks for your time,
Regards

### LINKED ISSUE #985: Stop blocking me from typing my code
Sometimes I get blocked and I cannot enter newlines. Message in minibuffer says:

> Illegal token: import

I believe it's `haskell-indentation-mode`. My opinion is that it should never block user from typing.

Not sure to reproduce this reliably, but I'll find out when I have time.

gracjan:
Fixes #962. Fixes #985. Fixes #861. Fixes #970. Fixes #969.

Note that there is still a possibility to get an empty list of indentations, that fails indentations also.

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -132,12 +132,6 @@ set and deleted as if they were real tabs."
                'haskell-indentation-mode
                "2015-05-25")
 
-(defun haskell-indentation-parse-error (&rest args)
-  "Create error message from ARGS, log it and throw."
-  (let ((msg (apply 'format args)))
-    (message "%s" msg)
-    (throw 'parse-error msg)))
-
 (defvar haskell-literate) ; defined in haskell-mode.el
 
 (defun haskell-indentation-bird-p ()
@@ -456,10 +450,7 @@ indentation points to the right, we switch going to the left."
           (haskell-indentation-first-indentation)
         (setq current-token (haskell-indentation-peek-token))
         (catch 'parse-end
-          (haskell-indentation-toplevel)
-          (unless (eq current-token 'end-tokens)
-            (haskell-indentation-parse-error
-             "Illegal token: %s" current-token)))
+          (haskell-indentation-toplevel))
         possible-indentations))))
 
 (defun haskell-indentation-first-indentation ()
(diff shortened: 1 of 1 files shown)


alanz:
No more hanging?

gracjan:
Not yet. There is one more thing to fix.

alanz:
Ok, progress though. Great.

gracjan:
@alanz: This is about right. Never throws, sometimes gets lost, but that is expected.

Issue #981, opened 2015-11-07
M-j to continue a comment on the next lines indents the next line


emmanueltouzery:
in emacs while in the middle of a comment, it's possible to ask to insert a carriage return and continue the comment on the next line.

eg, if "|" is the cursor position:

    -- my comment|

you press M-j and you should get:

    -- my comment
    -- |

but I get:

    -- my comment
        -- |

the comment continuation on the second line should not be indented.

gracjan:
Are you able to add a test case to `tests/haskell-indentation-mode-tests.el`?

emmanueltouzery:
I can try. I opened the file and it looks straightforward to add new tests, however I didn't manage to run the exists tests. I also didn't find guidance about that in the contribution guide or readme? Before I can modify tests I must be able to run the existing ones. I guess it's basic knowledge for authors of emacs-lisp packages, but it's knowledge I just don't have currently...

gracjan:
make check

emmanueltouzery:
ok this test accurately reproduces the problem:

```lisp
(ert-deftest haskell-indentation-altj-comment ()
  (switch-to-buffer (get-buffer-create "another.hs"))
  (haskell-mode)
  ;; (insert "-- comment") ;; <-- with this insert, it passes
  (insert "main = do\n    return ()\n\n-- comment")
  (execute-kbd-macro (kbd "M-j"))
  (should (equal 2 (count-lines (point-min) (point))))
  (should (equal 3 (- (point) (line-beginning-position)))))
```

i guess fixing the problem itself would be considerably more involved..?

gracjan:
Add:

    :expected-result :failed

and then this test case qualified for a pull request.

As about fixing it really: indentation mode currently ignores comments (treats them as whitespace), so it aligns your second comment to the tokens that were even before the first comment. Comments should be represented as a new token type. This requires some thought, but you may give it a try.

gracjan:
Interestingly the test that we have for this work on Emacs 25.

fice-t:
@gracjan That's because they changed the function in the core. There's this included part:
```elisp
(comment-indent-function
 ;; If the previous comment is on its own line, then
 ;; reuse its indentation unconditionally.
 ;; Important for modes like Python/Haskell where
 ;; auto-indentation is unreliable.
 (if (save-excursion (goto-char compos)
                     (skip-chars-backward " \t")
                     (bolp))
     (lambda () comment-column) comment-indent-function))
```

[...]

PR #1010, opened 2015-12-01
Failing testcase for bug #981: M-j to continue a comment on the next lines indents the next line

emmanueltouzery:
pull request for the failing testcase for bug #981 as discussed in the bug itself.


gracjan:
Nice, this will really help! Thanks.

PR #1014, opened 2015-12-03
Add haskell-tab-indent.el

spwhitton:
This file provides `haskell-tab-indent-mode`, a simple indentation mode for Haskell projects which require tabs for indentation and do not permit spaces (except for where clauses, as a special case).

The user may use TAB to cycle between possible indentations.

[This coding style](https://propellor.branchable.com/coding_style/) of a project I like to contribute to prompted me to write this mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,7 @@ ELFILES = \
 	haskell-simple-indent.el \
 	haskell-sort-imports.el \
 	haskell-string.el \
+	haskell-tab-indent.el \
 	haskell-unicode-input-method.el \
 	haskell-utils.el \
 	inf-haskell.el
--- a/haskell-tab-indent.el
+++ b/haskell-tab-indent.el
@@ -91,7 +91,9 @@
 
 (defun haskell-tab-indent--previous-line-tabs ()
   (save-excursion
-    (forward-line -1)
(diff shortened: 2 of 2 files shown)


ivan-m:
That is an odd coding style, since AFAIK pretty much everyone in the Haskell community recommends using spaces rather than tabs (and Cabal in fact prohibits usage of tab characters in `.cabal` files).

spwhitton:
I know! I dislike it. But prominent projects use it (e.g.
[git-annex](https://git-annex.branchable.com/coding_style/)) so it would
be useful to have support for it in haskell-mode.

gracjan:
Interesting coding style. Note though that haskell-mode already has 3 indentation styles and Haskell Mode Community is just not able to support and maintain one more mode.

Can you release this as a separate package?

spwhitton:
I'd be willing to support it.  Any issues could be tagged for me to
handle and submit pull requests to fix.  If you are sure that you don't
want to include it, I'll release as a separate package and it could be
linked to from the haskell-mode wiki.

gracjan:
Yes, please create a separate package. Thanks.

spwhitton:
Published independently and added to the wiki page on indentation.

PR #1036, opened 2015-12-25
Implement font-lock for quasi quoted XML, HTML and JavaScript

gracjan:


geraldus:
Nice!  Also, I would mention these shakespearen-QQs:
+ `whmalet`
+ `ihamlet`
+ `xshamlet`.

PR #1055, opened 2016-01-06
Require at least Emacs 24.1

### LINKED ISSUE #847: Drop Emacs 23 support
We have introduced lexical binding, currently we use it in backward compatible manner. This requires us to jump through some hops to keep the appearances. An example can be found here:

https://github.com/gracjan/haskell-mode/blob/9a6fa5073ac505c41f085d492f39e8ff415a7d13/haskell-indentation.el#L899-L904

```el
  (haskell-indentation-with-starter
   `(lambda ()
      (let ((implicit-layout-active nil))
        (haskell-indentation-separated
         #',parser ,sep ,stmt-sep)))
   end))
```

On top of this Emacs 23 lacks build in `cl-lib` and does not support `package.el` out of the box. Emacs 23 is old: 23.1 was released 2009-07-29, 23.4 was released 2012-01-29.

Emacs 24 and coming Emacs 25 are much better and are easy to install on every supported platform.

Steps to take:
- [x] make sure we do not alienate anybody by dropping Emacs 23
- [x] make last release supporting Emacs 23
- [x] remove TravisCI build for Emacs 23
- [x] make haskell-mode fail if enabled under Emacs less than 24
- [x] add a note in documentation that release 13.18 was the last one to support Emacs 23
- [x] remove other mentions of Emacs 23 from the documentation
- [x] remove Emacs 23 from README
- [x] remove Emacs 23 from wiki

Comments?

gracjan:
Fixes #847.

Note that we will merge this one after nearest melpa stable release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ dist: trusty
 
 env:
   matrix:
-    - EMACS=emacs23           TARGET=check
     - EMACS=emacs24           TARGET=check
     - EMACS=emacs-snapshot    TARGET=check
     - TARGET=deploy-manual
@@ -16,10 +15,6 @@ matrix:
     - env: EMACS=emacs-snapshot    TARGET=check
 
 install:
-  - if [ "$EMACS" = "emacs23" ]; then
-        sudo apt-get update &&
-        sudo apt-get install --no-install-suggests --no-install-recommends --assume-yes emacs23-nox;
-    fi
   - if [ "$EMACS" = "emacs24" ]; then
         sudo apt-get update &&
         sudo apt-get install --no-install-suggests --no-install-recommends --assume-yes emacs24-nox;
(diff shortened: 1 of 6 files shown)


fice-t:
@gracjan I thought about this when I saw a lot of `set (make-local-variable 'x)` instead of `setq-local`. I don't think 24.1 and 24.2 include `cl-lib` by default, so shouldn't support now only go back to 24.3?

gracjan:
I have neither 24.1 nor 24.2 handy (and we do not automatically test with those), but as I understand 24.1 has package manager and can install from Melpa. Our package explicitly depends on cl-lib so package manager should download cl-lib as a dependency for haskell-mode.

gracjan:
I just checked and 24.1.1 and 24.2.1 pass the test suite.

fice-t:
Ah, that makes sense.

PR #1061, opened 2016-01-10
Use plain string face for quasi quotes

gracjan:


gracjan:
Nobody liked the background anyway...

PR #1064, opened 2016-01-12
Add keybinding for `haskell-cabal-visit-file`

bergey:
Instead of recommending this binding in the manual, provide it by
default, and document that it the manual.

This follows my understanding of https://github.com/haskell/haskell-mode/wiki/Complete-functionality

What's the recommended way to test functions that inspect the filesystem?  `el-mock`?  Some other mocking library?


gracjan:
We use `el-mode.el` already. It is even included verbatim in `tests/compat/el-mock.el` so that it works across all test cases across all supported Emacsen.

PR #1067, opened 2016-01-13
Add test for SQL font-lock

gracjan:


gracjan:
And what is with this? Works for me...

PR #1068, opened 2016-01-13
Add test for 'type role' font-lock

### LINKED ISSUE #1057: Font lock Role annotations
I just noticed #1056 and realised that [Role annotations] are also not properly font locked.

[Role annotations]: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/roles.html#role-annotations

Some examples that demonstrate that GitHub's syntax highlighting has the same problem :smiley::

```haskell
{-# LANGUAGE RoleAnnotations #-}

type role Ptr representational
data Ptr a = Ptr Addr#

type role Foo nominal representational phantom
data Foo a b c = Foo a b
```
The `role` keyword must only be highlighted after `type`, analogous to `family`. But what about `phantom`, `representational`, and `nominal`? They are also special keywords in this particular context, but specifying a regex for this seems nontrivial.

mrBliss:
Tests #1066 which fixes #1057. The test succeeds.

I tried to follow the conventions that I could find, but if you want me to move the test up in the file, rename it, or make any other change, let me know.

### DIFF
--- a/tests/haskell-font-lock-tests.el
+++ b/tests/haskell-font-lock-tests.el
@@ -554,3 +554,11 @@ if all of its characters have syntax and face. See
      ("code3" t haskell-definition-face)
      ("Comment3" t haskell-literate-comment-face))
    'literate))
+
+(ert-deftest haskell-type-role ()
+  "fontify \"role\" after \"type\""
+  (check-properties
+    '("type role Ptr representational")
+    '(("type" "w" haskell-keyword-face)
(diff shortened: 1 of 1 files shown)


gracjan:
Perfect! Thanks!

gracjan:
Can you add some more? We aren't testing `type family` nor `type instance`...

mrBliss:
Ok, will do.

PR #1070, opened 2016-01-13
More font lock tests

mrBliss:
More font-lock tests, as requested in #1068.


gracjan:
Lets wait till it builds.

PR #1083, opened 2016-01-14
Define haskell-package struct's slots

fice-t:
`haskell-package`'s slots were not defined, and that was creating a
byte-compiler warning that the lexical variable `pkg` was unused.

I made this commit just to shut up the byte-compiler (honestly, I just started using this mode). 

As an aside, am I missing something, or is the struct not actually used? The `haskell-package-get*` functions use `haskell-package-parse` to create the struct, but nothing seems to actually use the result (I'm not sure how something could without the slots defined).

### DIFF
--- a/haskell-package.el
+++ b/haskell-package.el
@@ -55,7 +55,14 @@
             name
             version))))
 
-(cl-defstruct haskell-package "Haskell package object.")
+;; Haskell package object
+(cl-defstruct haskell-package
+  name version id license copyright maintainer stability homepage
+  package-url description categories authors is-exposed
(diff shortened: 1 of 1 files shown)


bergey:
Your change makes sense to me.

I think you are right about the struct.  I suspect the whole module is unfinished, meant to be used for something unwritten.  I'm not sure, though.

gracjan:
There is also haskell-cabal.el that is also unfinished.

gracjan:
@fice-t, I appreaciated your last contribution to haskell-mode. Can you help out a little more? For example issue #669 would help a lot. What do you think?

PR #1087, opened 2016-01-16
Add website script reference to manual

gracjan:


ivan-m:
Why are you putting analytics into the manual? There are many of us that do our utmost to block such scripts, and it seems even more annoying when they're embedded into something that is built locally.

gracjan:
We would like to know the popularity of haskell-mode. What other method do you propose?

ivan-m:
MELPA has a download counter; Marmalade might have something similar as well.

PR #1089, opened 2016-01-16
Utils coverage

geraldus:
Don't merge yet!

@gracjan I have no strong opinion on `type-at-command-composition` test.  It looks very silly and feels like testing function's logic reimplementation.  Maybe there are some other aspects we'd better to test?

I will add few other tests in few days.

Also, maybe it's better to move prettification stuff, e.g. comments, indentation, to another PR?

### DIFF
--- a/tests/haskell-utils-tests.el
+++ b/tests/haskell-utils-tests.el
@@ -7,96 +7,96 @@
 
 (ert-deftest simple-import-parse ()
   (should (equal "A.B.C"
-		 (with-temp-buffer
-		   (insert-lines "import A.B.C")
-		   (goto-char (point-min))
-		   (forward-line 0)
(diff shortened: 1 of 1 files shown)


tests/haskell-utils-tests.el
```diff
@@ -1,102 +1,179 @@
+;;; haskell-utils-tests.el --- Tests for Haskell utilities package
+
+;; Copyright © 2016 Athur Fayzrakhmanov. All rights reserved.
```
geraldus: Please don't treat this as self-praise or etc. this is GPL requirement.  Also everyone who contribute to haskell-mode files should list themselves as authors.

geraldus:
I will add tests for `haskell-utils-parse-repl-response` and `haskell-utils-reduce-string`, but we still have no decision about async-stuff and hooks.

PR #1092, opened 2016-01-17
Improve haskell-utils-reduce-string. Add tests

geraldus:
If everything is fine, merge.

### DIFF
--- a/haskell-utils.el
+++ b/haskell-utils.el
@@ -100,12 +100,13 @@ execusion."
     (remove-hook
      'post-command-hook #'haskell-utils-async-update-post-command-flag t)))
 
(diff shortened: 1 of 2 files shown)


haskell-utils.el
```diff
@@ -100,12 +100,13 @@ execusion."
     (remove-hook
      'post-command-hook #'haskell-utils-async-update-post-command-flag t)))
```
geraldus: @gracjan I've spent few hour trying to make it record changes to flag during ert running test.  I have tried `command-execute`, `call-interactively`, `execute-kbd-macro`, and changing modes but flag remains empty all the time.  Can you help me with this?

PR #1093, opened 2016-01-17
Add post-command-hook related tests

geraldus:
Don't merge yet, I opened this to see coverage stats first.

@gracjan I've spent few hour trying to make it record changes to flag during ert running test. I have tried `command-execute`, `call-interactively`, `execute-kbd-macro`, and changing modes but flag remains empty all the time. Can you help me with this?

### DIFF
--- a/tests/haskell-utils-tests.el
+++ b/tests/haskell-utils-tests.el
@@ -218,4 +218,30 @@ removed."
            " TestTest Test test"
            (haskell-utils-reduce-string " Test\r\nTest\n    Test test"))))
 
+(ert-deftest post-command-hooks ()
+  "Test commands related `haskell-utils-async-post-command-flag'.
+Tests flag updates and `post-command-hook' cleanup."
+  (with-temp-switch-to-buffer
+    ;; set some random value to flag to test that it will be reseted
+    (setq haskell-utils-async-post-command-flag "non nil")
+    (haskell-utils-async-watch-changes)
+    ;; now flag should be empty
+    (should (null haskell-utils-async-post-command-flag))
+    ;; execute some commands
+    (save-excursion (insert "Hello World!"))
+    (execute-kbd-macro (kbd "SPC"))
+    ;; now flag should not be empty
+    (should (not (null haskell-utils-async-post-command-flag)))
+    ;; check that hook was installed
+    (should (cl-member #'haskell-utils-async-update-post-command-flag
+                       post-command-hook
+                       :test #'equal))
+    ;; check that flag was cleaned up
+    (haskell-utils-async-stop-watching-changes (current-buffer))
+    (should (null haskell-utils-async-post-command-flag))
+    ;; check that hook was removed
+    (should (not (cl-member #'haskell-utils-async-update-post-command-flag
+                            post-command-hook
+                            :test #'equal)))))
+
 ;;; haskell-utils-tests.el ends here


tests/haskell-utils-tests.el
```diff
@@ -225,4 +225,52 @@ removed."
...
+(ert-deftest post-command-hooks ()
+  "Test commands related `haskell-utils-async-post-command-flag'.
+Tests flag updates and `post-command-hook' cleanup."
+  (with-temp-buffer
+    ;; flag value should be reseted to nil when started watching chages
+    (setq haskell-utils-async-post-command-flag "non nil")
+    (haskell-utils-async-watch-changes)
+    (should (null haskell-utils-async-post-command-flag))
+    ;; now invoke some commands and test flag value
+    (insert "Hello World!")
```
gracjan: Use:

```el
      (execute-kbd-macro "DEF")
```

tests/haskell-utils-tests.el
```diff
@@ -225,4 +225,52 @@ removed."
            " TestTest Test test"
            (haskell-utils-reduce-string " Test\r\nTest\n    Test test"))))
 
+(ert-deftest post-command-hooks ()
+  "Test commands related `haskell-utils-async-post-command-flag'.
+Tests flag updates and `post-command-hook' cleanup."
+  (with-temp-buffer
```
gracjan: Use:

``el
(with-temp-switch-to-buffer ...
``

Note that `with-temp-switch-to-buffer` is currently defined in `haskell-indentation-tests.el`, it should be moved to the common file for test utilities we have been talking about.

tests/haskell-utils-tests.el
```diff
@@ -225,4 +225,52 @@ removed."
...
+    ;;               command-found)
+    ;;           (while list
+    ;;             (when (equal #'goto-char (car list))
+    ;;               (setq command-found t))
+    ;;             (setq list (cdr list)))
+    ;;           command-found))
+
+    ;; check that hook was installed
+    (should
+     (let ((list (cl-copy-list post-command-hook))
```
gracjan: Use `cl-member`:

https://www.gnu.org/software/emacs/manual/html_node/cl/Lists-as-Sets.html

gracjan:
Summary:
- `insert` does not execute post command hooks, but `execute-kbd-macro` does
- a temporary `switch-to-buffer` is needed for `execute-kbd-macro` to work correctly
- there are nice functional functions in `cl-lib.el`

tests/haskell-utils-tests.el
geraldus: This gives error during test:
> (end-of-buffer)

`(goto-char (point-min))` does not help.

geraldus:
Well, I've found only two ways to cover `haskell-utils-async-update-post-command-flag` — to call it directly or explicitly run post-command hooks `(run-hooks 'post-command-hook)`.  The former is just a coverage hack and in fact tests nothing, while latter looks good.  I guess post-command hooks wont't run until test body is not fully evaluated.  This method also work with `with-temp-buffer`.

geraldus:
Don't merge this yet.  Let's merge test-utils first, then I'll rebase code _(I guess this will be needed, won't it?)_.

geraldus:
Ok, this is almost total coverage! :chart_with_upwards_trend:

Any good ideas about last uncovered function?

gracjan:
Use `with-temp-switch-to-buffer` in place of `with-temp-buffer` and then use `execute-kbd-macro` in place of `(run-hooks 'post-command-hook)`. It will work correctly.

geraldus:
Well, we cannot test interactive stuff in batch mode, because:

> <forcer> Geraldus: Part of batch mode is that it does not *have* interactivity ...

> An important caveat to this approach is that it will not work in a batch-mode emacs session, presumably because emacs doesn't process keyboard input at all in batch mode.

So, I'll remove last commit in few minutes, @gracjan can we have some special init file to run Emacs not in batch mode _(if this is even possible on Travis)_, then run interactive tests with mentioned hacks [[1]] and finally exit Emacs?
By the way, can't contact you via IRC whole day, you're constantly online, does this mean that you receives all messages I sent you?

gracjan:
@geraldus: I never saw your interactive commit so I do not know what you are talking about.

I have some IRC bouncer set up, so I get all messages when I get back to computer.

geraldus:
@gracjan wow, Evil's registers prevented me from loosing it comletely:

```elisp
(ert-deftest read-directory-name ()
  "Test `haskell-utils-read-directory-name'.
This is very simple test, it just runs
  `haskell-utils-read-directory-name' function with default
  directory set to home dir.  Since we have no clue about
  evnironment the only way to test this function will be simulate
  RET keypress and compare result to home folder path - the
  latter should be a prexif ot former."
  (let* ((home (getenv "HOME"))
         (unread-command-events (listify-key-sequence (kbd "RET G-g")))
         (result (haskell-utils-read-directory-name "> " home)))
    (should (string-prefix-p home result))))
```
while this woks fine in non-batch mode, Travis builds are failing after 10 minutes timeout.

gracjan:
In the particular case of `haskell-utils-read-directory-name` it is better to use `el-mock` to mock `read-directory-name`. We have `el-mock` dependency already so this is good.

geraldus:
@gracjan I didn't catch how to use it, can you explain a bit?

gracjan:
Example of [`el-mock` usage](https://github.com/haskell/haskell-mode/blob/master/tests/haskell-process-tests.el#L31):

```el
    (mocklet (((haskell-session-name "dummy-session") => "dumses1"))
                     (haskell-process-compute-process-log-and-command "dummy-session" 'ghci)))
```

So in our situation this would be something like:

```el
    (mocklet (((read-directory-name ? ? ?) => ?))
                     (should (equal "xxx" (haskell-utils-read-directory-name ? ? ?)))
```

geraldus:
@gracjan let me not to be greedy in this respect and leave this final piece to you (or someone else).  Sorry, I need some time to learn that mocklets and have some other tasks to do in nearest time :pensive:

PR #1100, opened 2016-01-22
Implement haskell-goto-first-error

bergey:
Supercedes https://github.com/haskell/haskell-mode/pull/818

### DIFF
--- a/tests/haskell-load-tests.el
+++ b/tests/haskell-load-tests.el
@@ -1,0 +1,85 @@
+;;; haskell-load-tests.el
+
+;;; Code:
+
+(require 'ert)
+(require 'haskell-test-utils)
+
+(require 'haskell-load)
+
+(defun insert-errors ()
+  (insert "import Control.Applicativ\nimport Data.Mayb\nimport Data.String")
+  (goto-char 1)
+  (let ((applicativ (progn
+                      (search-forward "Control.Applicativ")
+                      (make-overlay (match-beginning 0) (match-end 0)))))
+    (overlay-put applicativ 'haskell-check t)
+    (overlay-put applicativ 'haskell-msg-type 'error)
+    (overlay-put applicativ 'haskell-msg "Could not find module ‘Control.Applicativ’\n    Perhaps you meant Control.Applicative (from base-4.8.1.0)\n    Use -v to see a list of the files searched for."))
+  (let ((mayb (progn
+                (search-forward "Data.Mayb")
+                (make-overlay (match-beginning 0) (match-end 0)))))
+    (overlay-put mayb 'haskell-check t)
+    (overlay-put mayb 'haskell-msg-type 'error)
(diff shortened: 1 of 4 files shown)


tests/haskell-load-tests.el
```diff
@@ -0,0 +1,82 @@
...
+;;; Code:
+
+(require 'ert)
+(require 'haskell-test-utils)
+
+(require 'haskell-load)
+
+(defun insert-errors ()
+  (insert "import           Control.Applicativ\nimport           Data.Mayb\nimport           Data.String")
+  (let ((a (make-overlay 18 36))
```
gracjan: Please use `re-search-forward` and then `(match-begin 0)` and `(match-end 0)` to put the overlays. Should somebody come and need to change these tests they will be intimidated by character counting and won't do it.

tests/haskell-load-tests.el
```diff
@@ -0,0 +1,82 @@
...
+  (with-temp-switch-to-buffer
+   (insert-errors)
+   (goto-char 1)
+   (haskell-goto-first-error)
+   (should (equal (point) 18))))
+
+(ert-deftest goto-first-error-after ()
+  (with-temp-switch-to-buffer
+   (insert-errors)
+   (goto-char 92)
```
gracjan: Please use `re-search-forward` instead of `goto-char`.

tests/haskell-load-tests.el
```diff
@@ -0,0 +1,82 @@
...
+   (haskell-goto-first-error)
+   (should (equal (point) 18))))
+
+(ert-deftest goto-first-error-between ()
+  (with-temp-switch-to-buffer
+   (insert-errors)
+   (goto-char 37)
+   (haskell-goto-first-error)
+   (should (equal (point) 18))))
```
gracjan: Please either use `re-search-forward` or use `looking-at-p` to check if we are in the correct spot in the source file.

gracjan:
Very nice! Please get rid of those hardcoded numbers and then we are good to merge.

bergey:
@gracjan I force pushed the changes you requested.

tests/haskell-load-tests.el
```diff
@@ -0,0 +1,85 @@
...
+
+;;; Code:
+
+(require 'ert)
+(require 'haskell-test-utils)
+
+(require 'haskell-load)
+
+(defun insert-errors ()
+  (insert "import           Control.Applicativ\nimport           Data.Mayb\nimport           Data.String")
```
gracjan: One more nitpick: Can I have one space here between `import` and module names?

Counting spaces might be intimidating for future developers, too.

bergey:
Makes sense.  I started from real Haskell code, and the extra spaces are what `stylish-haskell` does.  But they do make the tests harder to read otherwise.  I force pushed again.

gracjan:
Perfect, thanks.

PR #1114, opened 2016-01-28
Allow haskell-process-path-* to be lists.

### LINKED ISSUE #1112: haskell-process-path-* now allowed to be lists
The customize specs for haskell-process-path-ghci & friends say they can be lists of strings. But the code actually only works with strings.

(I want to use `("nix-shell" "--run" "ghci")`.)

mikesperber:
Fixes #1112.


gracjan:
Thanks.

Issue #964, opened 2015-10-27
Invoke hasktags more directly


gracjan:
Currently invoking hasktags goes via haskell-process, that is a roundabout way of doing it.

Invoke hasktags without the need for haskell-process.

Issue lifted from discussion in #891 and in #963.

Pinging @code-crimes and @MichaelXavier.

bergey:
I've started looking at this.  Two questions:

1. Would it be ok to depend on [async.el](https://github.com/jwiegley/emacs-async)?  For haskell-tags-on-save I'd like to call `hasktags` asyncronously, but I'd like to keep supporting `and-then-find-this-tag`.

2. Is there a test case that shows why we need the `find` command?  In my tests `hasktags` does a better job indexing all the Haskell files and ignoring all the build & backup files without the `find`.

gracjan:
1. I would prefer the blocking version first. Dependencies are inherently tricky on Emacs.
2. Later on the equivalent of a single function [`async-start-process`](https://github.com/jwiegley/emacs-async/blob/master/async.el#L199) can be pulled in verbatim.
2. I guess hasktags used to need find, this seems to be no longer the case. Although it needs to be checked if huge directories like '.git' or '.svn' or '.darcs' are ignored as they should be.

bergey:
Are dependencies really that bad these days?  Or do you mean that you still support emacs-23?  I dislike to maintain inline copies of something so generic as `async-start-process`.

gracjan:
I'd rather have inline copy for now, thanks.

geraldus:
Other question, should be generated tags table visited automatically?  Recently Emacs start to ask me almost every time I call `magit-status` something like this "TAGS was updates, visit new file?", and it annoys me (I'm not certain but I think this is started after `projectile` installation).

geraldus:
Before calling `hasktags` should we make some `.cabal` file lookup first?

MichaelXavier:
@geraldus I'm not sure if this will solve your case but you may want to set `tags-revert-without-query` to `t`.

geraldus:
@MichaelXavier thanks, I'll give it a try.

geraldus:
So, what is the plan?  Should we try to look up cabal file first going upward for some fixed depth?  If this is a case, should we:
a) parse cabal file to check if current file is listed as module in any project section
b) cache cabal file location somehow?

gracjan:
`haskell-cabal.el` already has this functionality.

[...]

PR #1118, opened 2016-01-28
Ignore stack and cabal files from hasktags

sLite:
As discussed on IRC.

As the windows version didn't include a find not applied there, and i have no windows environment to test on.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -712,7 +712,7 @@ function `xref-find-definitions' after new table was generated."
                (format ":!cd %s && %s | %s"
                        (haskell-session-cabal-dir
                         (haskell-process-session (car state)))
-                       "find . -type f \\( -name '*.hs' -or -name '*.lhs' -or -name '*.hsc' \\) -not \\( -name '#*' -or -name '.*' \\) -print0"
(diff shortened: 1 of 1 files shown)


gracjan:
Very close. Note that `find` has `-prune` option that would help a lot because it can stop `find` from entering unwanted directories and doing a lot of useless work:

```
find . 
    -type d \( bad_directories \) -prune
    -o
    -type f \\( -name '*.hs' -or -name '*.lhs' -or -name '*.hsc' \\) -not \\( -name '#*' -or -name '.*' \\)
    -print0
```

Please change to using `-prune`. For more inspiration `rgrep` elisp function has a lot of logic like that.

gracjan:
@sLite: Note also #964.

sLite:
@gracjan had some time issues lately, but i will look into this by the end of the week.

But i'm afraid that my experience with EL isn't nearly enough to tackle #964 

Putting that aside, is the allegedly decrease in coverage something i have to care about to get this PR merged? I don't see how changing the find commandline should change code coverage, but maybe i'm missing something :>

gracjan:
coverage is not a problem. coveralls seems to misreport this and that, we will fix the coverage reporting later. Right now this is just a hint not a rule.

sLite:
I've changed the ignored directories to -prune and moved from ignoring ```./dist/build``` to ignoring the ```./dist``` as a whole.

Please tell me if you think that could cause any troubles.
I did this because sometimes (i'm not exactly sure when and why) stack or cabal also place ```.hs``` files in ```./dist/production```

gracjan:
This is good as it is.

gracjan:
Thanks.

PR #1121, opened 2016-01-30
Remove suggested imports completely

### LINKED ISSUE #1109: haskell-process-suggest-remove-import leaves blank line rather than completely remove it
Current implementation of `haskell-suggest-remove-import` just clears line contents but do not removes it when user answers `y`.  I propose to remove entire line, e.g.
```hs
import A
import D -- ^ this import is redundant
import B

-- current behavior
import A

import B

-- proposed
import A
import B
```

geraldus:
Do not leave blank line when removing import statement.

Now covers multiline imports properly when remove or comment out option
selected.

Will think about some test, don't merge yet.

Fixes #1109

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -581,7 +581,7 @@ SESSION in specified FILE to remove IMPORT on given LINE."
         (cl-case (read-event
                   (propertize (format "%sThe import line `%s' is redundant. Remove? (y, n, c: comment out)  "
                                       (if (not first)
-                                          "Please answer n, y or c: "
+                                          "Please answer y, n or c: "
                                         "")
                                       import)
                               'face
@@ -591,9 +591,9 @@ SESSION in specified FILE to remove IMPORT on given LINE."
            (save-excursion
              (goto-char (point-min))
              (forward-line (1- line))
-             (goto-char (line-beginning-position))
-             (delete-region (line-beginning-position)
-                            (line-end-position))))
+             (let ((bounds (haskell-interactive-mode--import-statement-bounds)))
+               (delete-region (car bounds) (cdr bounds))
+               (kill-line 1))))
           (?n
            (message "Ignoring redundant import %s" import))
           (?c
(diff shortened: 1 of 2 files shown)


geraldus:
@gracjan I have no good test case for this.

chrisdone:
This is a regression! :scream: 

I'm reverting this PR.

@geraldus The reason this feature only blanks (rather than removes) the line and does not remove it is because then any future import line removals will have the wrong line number. Try this feature on this module:

``` haskell
module X where

import Data.Char
import Data.Foldable
import Prelude
import Control.Monad
x :: ()
x = ()
```

It prompts: (1) `Remove Data.Char` and removes `Data.Char`, then (2) `Remove Data.Foldable?` and removes `Prelude`. Then prompts `Remove Prelude?` hit `y` and it throws a type error.

So leaving the import lines there is the solution I went with.

* The alternative is to keep track of which lines have been removed and compute line offsets to properly remove the right lines.
* An easier alternative is to blank all the lines, then put the line numbers in a hashtable. Then at the end of the import removing process (e.g. after modules have loaded), go ahead and actually delete those lines from bottom to top.

 If you want to do that work, feel free. :+1: I might do it myself, but let's coordinate in this PR if one of us starts working on it.

geraldus:
@chrisdone ahh, I've noticed some issues with this, but haven't time to investigate what's going on, thanks.  I think I could take second approach.

chrisdone:
@geraldus Cool! :+1:

geraldus:
@chrisdone, by the way, we have discussed this a bit earlier and decided, that rather prompting user during loading process it is much better to inform him that there are redundant imports and suggest to run some command manually to remove/comment out this stuff.  The issue with current behavior (at least for me) that I usually miss this prompts and wait for some time before I finally realize that I have to answer what to do.  Also another a bit annoying thing is that I have to save manually all modified files.

So, I want to know other people opinion about following two points:

1. do not prompt user during compilation, warn him and suggest run special command to handle redundant imports instead
2. save all files after modifications

Ping @gracjan and other active people

gracjan:
I'd imagine overlays (similar to those provided by Flycheck), with the ability to navigate between them and then do fixups. Then we could have overlays signaling errors and warnings and some of those could have one (or more) automatic actions associated.

vlatkoB:
Can removing imports be done from the end? 
Than removing lines wouldn't affect other removals and line numbers are preserved.

geraldus:
@vlatkoB in this case we need some stack of all redundant imports.  I'm going to implement this soon https://github.com/haskell/haskell-mode/issues/1109#issuecomment-187441042

Issue #1126, opened 2016-02-01
Error: defcustom for ‘haskell-indent-spaces’ fails to specify type


aplanas:
Using last emacs from git, and last haskell-mode from git, make produces this error:

```
In toplevel form:
haskell-mode.el:831:1:Error: defcustom for ‘haskell-indent-spaces’ fails to specify type
Makefile:106: recipe for target 'build-25.1.50.1' failed
make: *** [build-25.1.50.1] Error 1
```

aplanas:
With this patch I am able to compile:

```
diff --git a/haskell-mode.el b/haskell-mode.el
index d45384e..62fabad 100644
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -830,7 +830,8 @@ Note that negative arguments do not work so well."
 
 (defcustom haskell-indent-spaces 2
   "Number of spaces to indent inwards."
-  :group 'haskell)
+  :group 'haskell
+  :type 'integer)
 
 ;; Like Python.  Should be abstracted, sigh.
 (defun haskell-check (command)
```

gracjan:
Fixed by #1128.

PR #1128, opened 2016-02-02
Fix snapshot build

geraldus:
Fix snapshot build
#1126


geraldus:
Ooops, something wrong with Ubuntu archives.  @gracjan if I'm not mistaken you can manually restart Travis builds?

Issue #1125, opened 2016-02-01
Indenting data types seems broken and not configurable


karljs:
I realize there is a lot going on with indentation right now, but this seemed like a pretty extreme case of something simple not working at all.  I looked through the other issues, too, and didn't spot anything.  Here's the scenario.  Start defining a new data type:

    data Foo = Bar

Newline, pipe character, indent, now gives:

    data Foo = Bar
      | Baz

And repeated tab presses do nothing to indent this further.  I want to align the pipe character with the equals character, as is common when not using GADTs.

If I already have two or more lines in the data type definition, then adding more after that seems to work just fine.  Thinking maybe this was just a new default, I looked through the customization options, but didn't see one that corresponds to this.

gracjan:
Thanks for the report. Can you add a test case in `tests/haskell-indentation-tests.el`?

karljs:
Should be no problem.  I'll take a look when I have some more spare minutes.

gracjan:
Great!

gracjan:
@karljs: How do you indent this fragment:

```haskell
data User = User
  { userName :: String }
          | Admin
  { adminName :: String }
```

?

karljs:
I wouldn't claim to have the definitive answer to this, but I would probably do the following.

```
data User = User { userName :: String
                 , ...}
          | Admin { adminName :: String
                  , ...}
```

This also seems to work fine for me currently.

gracjan:
Thanks. We try to reconcile two styles of indentation to reduce a bit of complexity. Hence the question.

PR #1129, opened 2016-02-03
Add indentation tests for constructor and deriving

karljs:
This adds two tests for indenting simple data type definitions.  I didn't entirely understand the numbering scheme, so I just made an educated guess and added them at the end.  If that's wrong, let me know and I can fix it.

See #1125 for a brief explanation.

### DIFF
--- a/tests/haskell-indentation-tests.el
+++ b/tests/haskell-indentation-tests.el
@@ -896,13 +896,13 @@ foo x
               (2 2)
               (3 2))
 
-(hindent-test "55 data constructor on separate line" "
+(hindent-test "55* data constructor on separate line" "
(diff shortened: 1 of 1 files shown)


gracjan:
Add a star to mark tests as failing, so `55*` and `55a*`.

gracjan:
Great, thanks, that helped a lot. It is already fixed, will be merged when Travis finishes its job.

Indentation could use some more test cases. For example we miss test case for LambdaCase:

https://coveralls.io/files/1940562436#L580

Can you add it, @karljs?

karljs:
@gracjan No problem.  I will take a look at the other test case you mentioned in the next day or two.

PR #1130, opened 2016-02-04
Align data declaration contructors in indentation

### LINKED ISSUE #1125: Indenting data types seems broken and not configurable

gracjan:
Fixes #1125.

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -648,8 +648,10 @@ For example
            (haskell-indentation-add-indentation current-indent)
            (throw 'parse-end nil)))
         ((string= current-token "=")
-         (haskell-indentation-separated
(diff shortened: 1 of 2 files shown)


geraldus:
:+1:

Issue #975, opened 2015-11-01
Improve error handling of exec of external programs if they are absent


gracjan:
Basically we need to render this page useless:

https://github.com/haskell/haskell-mode/wiki/PATH

Tasks to do:
- [ ] implement exec function that handles errors
- [ ] exec function should detect when external program is absent
- [ ] it should use `display-error` with information about
- [ ] what program we looked for
- [ ] which customization variable can be used to adjust it
- [ ] what is the current `PATH` or `exec-path` variable
- [ ] info that `PATH` may differ if Emacs was run from command line vs from GUI
- [ ] change all invocations to use this canned procedure

gracjan:
Updated to note handling of both direct paths and list of path plus arguments.

PR #1135, opened 2016-02-05
Docstrings and documentation for haskell-process-path-*.

mikesperber:
(In the wake of #1112.)

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -115,19 +115,28 @@ when showing type information about symbols."
 
 (defcustom haskell-process-path-ghci
   "ghci"
-  "The path for starting ghci."
+  "The path for starting ghci.
+This can either be a single string or a list of strings, where the
+first elements is a string and the remaining elements are arguments,
+which will be prepended to `haskell-process-args-ghci'."
   :group 'haskell-interactive
   :type '(choice string (repeat string)))
 
 (defcustom haskell-process-path-cabal
   "cabal"
-  "Path to the `cabal' executable."
+  "Path to the `cabal' executable.
+This can either be a single string or a list of strings, where the
+first elements is a string and the remaining elements are arguments,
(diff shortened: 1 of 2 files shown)


haskell-customize.el
```diff
@@ -115,19 +115,28 @@ when showing type information about symbols."
 
 (defcustom haskell-process-path-ghci
   "ghci"
-  "The path for starting ghci."
+  "The path for starting ghci.
+This can either be a single string or a list of strings, where the
+first elements is a string and the remaining elements are arguments,
+which will be prepended to `haskell-process-args-gci`."
```
gracjan: Typo: s/gci/ghci/g. Same with other docstrings. Thanks.

mikesperber: Gracjan Polak <notifications@github.com> writes:

>> @@ -115,19 +115,28 @@ when showing type information about symbols."
>>  
>>  (defcustom haskell-process-path-ghci
>>    "ghci"
>> -  "The path for starting ghci."
>> +  "The path for starting ghci.
>> +This can either be a single string or a list of strings, where the
>> +first elements is a string and the remaining elements are arguments,
>> +which will be prepended to `haskell-process-args-gci`."
>
> Typo: s/gci/ghci/g. Same with other docstrings. Thanks.

Done.

-- 
Regards,
Mike

gracjan:
Perfect, thanks.

Have you seen anything else you could help with?

mikesperber:
Not at the moment. I'm afraid I'm only a casual Haskell, and, hence, haskell-mode, user. But I'll keep an eye open, as always.

gracjan:
How about issue https://github.com/haskell/haskell-mode/issues/975? There were six places where haskell mode invokes external programs and those could use informative error handling. We get regular issue reports about missing executables. Do you think you could tackle #975?

Issue #895, opened 2015-09-28
Haskell Interactive Shell Stuck in Insert Mode (spacemacs)


gracjan:
`haskell-interactive-mode` somehow interacts badly with spacemacs:

https://github.com/syl20bnr/spacemacs/issues/445

PierreR:
I would close this. I am personally OK with the workaround as described here:
https://github.com/syl20bnr/spacemacs/issues/3271

As I see it,  the link issue is not really a `haskell-mode` issue

Issue #876, opened 2015-09-14
Can't press enter in haskell interactive mode


alokedesai:
When I press `C-c C-l` in a haskell buffer, a haskell REPL opens successfully, but I can't press enter to evaluate a haskell expression. Any ideas what this could be? haskell-mode is the only package I have installed and I've also tried reinstalling haskell-mode and even emacs to no avail.

Here's my `.emacs` file in case you need it

```
(add-to-list 'load-path "~/.emacs.d/el-get/el-get")

(unless (require 'el-get nil 'noerror)
  (with-current-buffer
      (url-retrieve-synchronously
       "https://raw.githubusercontent.com/dimitri/el-get/master/el-get-install.el")
    (goto-char (point-max))
    (eval-print-last-sexp)))

(add-to-list 'el-get-recipe-path "~/.emacs.d/el-get-user/recipes")
(el-get 'sync)

(add-to-list 'load-path "~/structured-haskell-mode/elisp")
(require 'shm)

(add-hook 'haskell-mode-hook 'structured-haskell-mode)


(require 'haskell-interactive-mode)
(require 'haskell-process)
(add-hook 'haskell-mode-hook 'interactive-haskell-mode)

(custom-set-variables
 '(haskell-process-suggest-remove-import-lines t)
 '(haskell-process-auto-import-loaded-modules t)
   '(haskell-process-log t))
```

gracjan:
Can you try to set to nil each of the variables in turn:

```el
'(haskell-process-suggest-remove-import-lines t)
'(haskell-process-auto-import-loaded-modules t)
'(haskell-process-log t)
```

so that only one of those is set to `t` and then report what happened?

alokedesai:
I've tried every possible permutation of setting t to nil in these three lines and none of it worked

gracjan:
Is there anybody else out there that has similar problem?

alokedesai:
I've seen this [stackoverflow post](http://stackoverflow.com/questions/31220370/emacs-interactive-haskell-repl-non-response-if-either-the-cabal-or-working-direc) and another forum post somewhere that I can't remember that both describe similar problems but no fix has been described.

gracjan:
33 upvotes and you are the first one to report the issue. Thanks for doing that. 33 people will be happy when we fix this one!

yamafaktory:
Hi there,
I am also experiencing the same weird issue trying to play with the Haskell REPL, i.e. not being able to press enter.
[...]

PR #1136, opened 2016-02-06
Add front-sticky property to read-only text in REPL

fice-t:
This makes it so that you can't insert new text inside read-only text.

This doesn't feel like the best way of doing it (namely the prompt-specific part), but the other ways that I could think of would need an overhaul of the system or they resulted in surprisingly hard to debug errors, so I just settled on this.

The main reason for this is to prevent people from inserting behind the read-only space at the end of the default prompt and wondering why they can't hit RET to evaluate their input (which happened to me enough times).

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -298,26 +298,24 @@ do the
 (defun haskell-interactive-mode-prompt (&optional session)
   "Show a prompt at the end of the REPL buffer.
 If SESSION is non-nil, use the REPL buffer associated with
-SESSION, otherwise operate on the current buffer.
-"
+SESSION, otherwise operate on the current buffer."
   (with-current-buffer (if session
                            (haskell-session-interactive-buffer session)
                          (current-buffer))
     (goto-char (point-max))
-    (insert (propertize haskell-interactive-prompt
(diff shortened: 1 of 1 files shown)


gracjan:
Is #876 related?

gracjan:
Ill merge this tomorrow, unless somebody else chimes in with comments. The `sticky` parts isn't exactly my area of expertise.

fice-t:
I don't think so. It affected the spacemacs guys though (#895).

geraldus:
One annoying thing related to read-only text in REPL is that if I yank output of invoked command and paste it as input for next command I can't modify it, and removing readonly text property is a headache (I guess there could be some silly solution which I'm not aware of).

fice-t:
Shouldn't you be able to modify yanked read-only text? I'm able to with these changes.

Also I found this:
```
yank-excluded-properties is a variable defined in `simple.el'.
Its value is
(category field follow-link fontified font-lock-face help-echo intangible invisible keymap local-map mouse-face read-only yank-handler)


Documentation:
Text properties to discard when yanking.
The value should be a list of text properties to discard or t,
which means to discard all text properties.
```
which includes the read-only property.

PR #1140, opened 2016-02-07
Use `haskell-interactive-switch` in process-load

### LINKED ISSUE #955: haskell-process-load-or-reload and haskell-interactive-switch-back
Hey,

I like to use `haskell-process-load-or-reload` function because it switches to REPL and loads buffer into it. The only thing I don't like about it - it doesn't play well with `haskell-interactive-switch-back` function. What do you think about reusing `haskell-interactive-switch` in `haskell-process-load-or-reload`?

bergey:
This sets a variable with the source buffer so we can switch back.

This is pretty bizzare, though.  Both before and after this commit,
`haskell-session-interactive-buffer` switches to the buffer if it's
newly created, but if the buffer already exists, just returns the buffer
without visiting.  This function shouldn't switch, and callers
should decide if they want to switch.

closes #955

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -519,7 +519,7 @@ FILE-NAME only."
         (with-current-buffer buffer
           (haskell-interactive-mode)
           (haskell-session-assign s))
-        (switch-to-buffer-other-window buffer)
+        (haskell-interactive-switch)
         buffer))))
 
 (defun haskell-process-cabal-live (state buffer)


bergey:
If you merge this now, I can open a new issue about checking which uses of `haskell-session-interactive-buffer` rely on the undocumented switching behavior, and refactoring to fix them.  I'm not sure when I'll have time to actually do that searching.

gracjan:
Thanks. The switching behavior is one of the more confusing parts when working with `haskell-interactive`. Please try to sort this out when you have time.

PR #1149, opened 2016-02-11
Fix #1146 by treating . as punctuation

### LINKED ISSUE #1146: Reconsider word breaks for haskell-cabal-mode
One thing I dislike in `haskell-cabal-mode` that functions such
* `left-word`
* `right-word`
* `backward-kill-word`

and etc. do not considering point `.` as word break char, for example 
```cabal
  exposed-modules:  Internal.Type
```
if I place point at beginning of exposed module name `Internal.Type` and invoke `right-word` it moves point to line end, while I expect that point should go to beginning of "Type" (like in `haskell-mode`).  Now if I invoke `backward-kill-word` it will delete whole module name, while I expect it will leave "Internal." part untouched.

mrBliss:
In `haskell-cabal-mode` with the point indicated with the vertical bar below, `forward-char` should leave the point after `Some` instead of after `Type`.
```haskell
Exposed-Modules:   |Some.Internal.Type
```

This was caused by the following line:
```emacs-lisp
(modify-syntax-entry ?. "w"  st)
```

Which makes "." a word constituent instead of a punctuation character. This was added in aa970092a1e812ff78af2c1d65484b4281ce2234, "merge caballero with haskell-cabal". The fix: remove the line above.

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -80,7 +80,6 @@
     ;; We could use font-lock-syntactic-keywords, but is it worth it?
     ;; (modify-syntax-entry ?-  ". 12" st)
     (modify-syntax-entry ?\n ">" st)
-    (modify-syntax-entry ?. "w"  st)
     (modify-syntax-entry ?- "w"  st)
     st))
 
--- a/tests/haskell-cabal-tests.el
+++ b/tests/haskell-cabal-tests.el
@@ -50,6 +50,24 @@
       (haskell-cabal-previous-subsection)
       (haskell-cabal-previous-section))))
 
+(ert-deftest haskell-cabal-period-is-a-word-break ()
+  (with-temp-buffer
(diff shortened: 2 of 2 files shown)


gracjan:
I wonder what will break if we merge this...

mrBliss:
When I tried to do the same thing for hyphens, as suggested by bergey in #1146, many tests started failing. The combination of `\w` in regexps and hyphens in sections headings and/or targets was the reason. As periods are not used there (or are they used in targets?) I don't think much will break.

gracjan:
Lets wait the usual day before merging. If nobody protests by tomorrow we will merge it.

geraldus:
Thanks! (:

bergey:
@mrBliss Thanks for looking into the hyphens.  I'll take a shot at them when I have (a lot) more time.

PR #1152, opened 2016-02-11
Better string animation

gracjan:


geraldus:
New animation a bit blurry

gracjan:
Hmm, wrong version of imagemagick probably.

alanz:
The new one is half the size in each dimension. Is this intended?

gracjan:
Other images, those already present in documentation were made on non-retina computer. The new one is made on a retina one. Screen capture produces twice as big image on retina displays, so the natural step is to resize to 50%.

As I see all of those things produce artefacts. So how to make a screenshot that does not introduce issues?

PR #1153, opened 2016-02-11
[WIP] Add Semantic support for Haskell

### LINKED ISSUE #1122: Research Wisent/Semantic
@kuribas: Thanks for the tip about [Wisent](https://www.gnu.org/software/emacs/manual/html_node/wisent/index.html). So far this looks really good.

From a short research it seems that we can write a pretty easy and forgiving parser that will be able to parse layout vs non-layout and type vs expressions.

As experiment I'll try to implement #899.

gracjan:
Fixes #1122.


gracjan:
@kuribas: Here is what I currently have with Semantic.

gracjan:
I give up. This is too much work to get semantic right for Haskell.

PR #1159, opened 2016-02-15
Improve font-lock

### LINKED ISSUE #628: In 12.34e-56 (a single number) both . and - are colorized as operators

### LINKED ISSUE #1157: Font lock module qualification differently
Currently we can recognize tokens that are qualified by module path. For example:

```
Control.Monad.>>=
Data.List.sort
Data.Map.Map
Data.HList.:~:
Control.Category..
```

We have a regexp that knows about module prefix and the unqualified part. It means that we can highlight differently the parts:

1. The module names.
2. The dots that are module separators (last dot can be different if we care).
3. The unqualified identifier, constructor or symbol.

Note that we do good job of differentiating the dot operator even if it is qualified.

Note2: Unicode prettification composition could be built in in this mechanism so that it is done in the same pass.

gracjan:
Fixes #1157. Fixes #628.

### DIFF
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -455,7 +455,8 @@ Regexp match data 0 points to the chars."
          (font-lock-syntactic-face-function
           . haskell-syntactic-face-function)
          ;; Get help from font-lock-syntactic-keywords.
-         (parse-sexp-lookup-properties . t))))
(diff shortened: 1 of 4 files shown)


gracjan:
new-font-lock-scheme

PR #1161, opened 2016-02-16
HTML manual fixup machinery

### LINKED ISSUE #1160: HTML manual Retina resolution animated gifs
Write an Emacs function that will change IMG tags in documentation from:

    

to

    

where `W` and `H` are half of gif width and height. (Half because screenshots are made on retina display).

Invocation should be:

    emacs -l haskell-manual-gif-size-fixup.el -f haskell-manual-gif-size-fixup-batch-and-exit *.html

gracjan:
Fixes #1160


gracjan:
Suddenly pink background... this is a surprise!

PR #1163, opened 2016-02-18
Make RET in REPL create new prompt with blank input

fice-t:
This creates a new prompt that doesn't affect the history or send
any information to the GHCi process. This only occurs when on the
same line as the last prompt.

This makes the REPL more like the others you'd find in Emacs, and also more like GHCi itself.


gracjan:
Nice, thanks.

PR #1165, opened 2016-02-19
Revert "Remove suggested imports completely"

chrisdone:
Reverts haskell/haskell-mode#1121


gracjan:
What is wrong with this?

cocreature:
@gracjan See https://github.com/haskell/haskell-mode/pull/1121#issuecomment-186160932

PR #1166, opened 2016-02-20
Add haskell-auto-import

bch29:
I've added a new set of functions to make working with imports nicer. This is the first elisp code I've written beyond configuring my own emacs, so feedback would be much appreciated.

## Functionality provided

`haskell-auto-import-unqualified` prompts the user for a module name and sub-imports, then adds the unqualified import line, or modifies it if the module is already imported. For example, imagine the imports look like this:
```haskell
module MyModule where

import Data.List (nub, catMaybes, sort)
import Data.Ord (comparing)
```

Then asking it to import `intercalate, sort` from `Data.List` will change it to this:

```haskell
module MyModule where

import Data.List (nub, catMaybes, sort, intercalate)
import Data.Ord (comparing)
```

The other functions do pretty much what you'd expect. Notable is `haskell-auto-import-qualified-and-unqualified`, which prompts for a module name, qualifier and sub-imports. For example, if you give it `Data.Map`, `Map` and `Map`, it will add the lines:

```haskell
import Data.Map (Map)
import qualified Data.Map as Map
```

## Caveats

Due to my somewhat ad-hoc regex parsing methods, these functions aren't able to deal with datatype operators with sub-sub-imports properly. For example, asking it to include as a sub-import something like `(+-+)(A, B)` will not do the right thing. Sub-sub-imports work fine otherwise, though. It will import `MonadState(get, put, modify), MonadWriter(tell)` properly, and even merge the sub-sub-imports in this case.

### DIFF
--- a/haskell-auto-import.el
+++ b/haskell-auto-import.el
@@ -1,0 +1,286 @@
+;;; haskell-navigate-imports.el --- A function for cycling through Haskell import lists -*- lexical-binding: t -*-
+
+;; Copyright (C) 2016  Bradley Hardy
+
+;; Author: Bradley Hardy <bradleyhardy@live.com>
+
+;; This file is not part of GNU Emacs.
+
+;; This program is free software: you can redistribute it and/or
+;; modify it under the terms of the GNU General Public License as
+;; published by the Free Software Foundation, either version 3 of
+;; the License, or (at your option) any later version.
+
+;; This program is distributed in the hope that it will be
+;; useful, but WITHOUT ANY WARRANTY; without even the implied
+;; warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+;; PURPOSE.  See the GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public
+;; License along with this program.  If not, see
+;; <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Use of the exported interactive functions should never leave the imports in
+;; an invalid state, as long as all names provided by the user exist in the
+;; right place. Adding an unqualified import twice will merge the sub-imports.
+
+;; This module works independently of most of the rest of haskell-mode, just
+;; requiring haskell-navigate-imports.
+
+;; Exports four interactive functions:
+;; 1. haskell-auto-import-unqualified
+;; 2. haskell-auto-import-unqualified-all
+;; 3. haskell-auto-import-qualified
+;; 4. haskell-auto-import-qualified-and-unqualified
+
+;; Example usage:
+
+;; (require 'haskell-auto-import)
+;; (define-key haskell-mode-map (kbd "C-c i i") 'haskell-auto-import-unqualified)
+;; (define-key haskell-mode-map (kbd "C-c i a") 'haskell-auto-import-unqualified-all)
+;; (define-key haskell-mode-map (kbd "C-c i q") 'haskell-auto-import-qualified)
+;; (define-key haskell-mode-map (kbd "C-c i x") 'haskell-auto-import-qualified-and-unqualified)
+
+;;; Code:
+
+(require 'haskell-navigate-imports)
+(require 'cl-lib)
+
+(defun haskell-auto-import-unqualified (module-name sub-imports)
+  "Add a new unqualified import of the provided module and sub-imports. If
+sub-imports is left blank, import everything in the module."
+  (interactive "MModule Name: \nMSub-Imports (comma separated, leave blank to import all): ")
+  (let* ((all-infos (haskell-auto-import-module-import-info module-name))
+         (non-qualified (member-if '(lambda (info) (not (car (cdr (cdr info))))) all-infos))
+         (new-sub-imports-list
+          (haskell-auto-import-parse-import-parens sub-imports)))
+    (if non-qualified
+        (let* ((info (car non-qualified))
+               (keep-info (butlast info 1))
+               (imported-items (car (last info 1)))
+               (new-info
+                (append keep-info
+                        (if new-sub-imports-list
+                            (list
+                             (haskell-auto-import-normalize-imported-items
+                              (append imported-items new-sub-imports-list)))
+                          ()
+                            ))))
+          (haskell-auto-import-replace-import module-name new-info))
+      (progn
+        (let ((old-point (point)))
+          (if all-infos
+              (goto-char (+ 1 (car (car all-infos))))
+            (haskell-navigate-imports))
+
+          (insert
+           (haskell-auto-import-print-module-import
+            module-name
+            (list nil nil new-sub-imports-list)))
+          (newline)
+
+          (goto-char old-point))))))
+
+(defun haskell-auto-import-unqualified-all (module-name)
+  "Add a new unqualified import of the provided module. Everything in that
+module is imported."
+  (interactive "MModule Name: ")
+  (haskell-auto-import-unqualified module-name ""))
+
+(defun haskell-auto-import-qualified (module-name qualifier)
+  "Add a new qualified import of the provided module with the provided
+qualifier. Attempts to place the new import next to existing imports of the same
+module if they exist."
+  (interactive "MModule Name: \nMQualifier: ")
+  (let* ((all-infos (haskell-auto-import-module-import-info module-name))
+         (qualified
+          (member-if '(lambda (info)
+                        (car (cdr (cdr info))))
+                     all-infos))
+         (same-qualifier
+          (member-if '(lambda (info)
+                        (string= (car (cdr (cdr info))) qualifier))
+                     all-infos)))
+    (if same-qualifier
+        (message "That module is already imported with the same qualifier. Doing nothing.")
+      (let ((old-point (point)))
+        ;; If the module is already imported in some capacity, put the new
+        ;; import above the existing ones
+        (if all-infos
+            (goto-char (+ 1 (car (car all-infos))))
+          (haskell-navigate-imports))
+
+        (insert
+         (haskell-auto-import-print-module-import
+          module-name
+          (list t qualifier nil)))
+        (newline)
+
+        (goto-char old-point)))))
+
+(defun haskell-auto-import-qualified-and-unqualified (module-name qualifier sub-imports)
+  "Add a new import, both qualified and unqualified. For example, given the
+module `Data.Map', qualifier `Map' and sub-imports `Map', this will add the lines:
+
+import Data.Map (Map)
+import qualified Data.Map as Map"
+
+  (interactive
+   "MModule Name: \nMQualifier: \nMSub-Imports (comma separated, leave blank to import all): ")
+  (haskell-auto-import-qualified module-name qualifier)
+  (haskell-auto-import-unqualified module-name sub-imports))
+
+(defun haskell-auto-import-parse-import-parens (contents)
+  "Given the contents of some import parentheses, return a list of pairs which
+each consist of '(imported-item sub-imports), where sub-imports is the list of
+items imported from the main item (e.g. in `import ModuleName(DataType(Con1,
+Con2))')
+
+Does NOT currently work properly with datatype operators with sub-imports (e.g.
+`import Foo.Bar ((**)(A, B))') or with sub-imports which are operators (e.g.
+`import Foo.Bar (A((&&), (||)))') "
+  (let* ((exploded (split-string contents "," nil "\\s-"))
+         (parts
+          (reduce
+           '(lambda (acc str)
+              (let ((cur (car acc))
+                    (rest (cdr acc))
+                    (strlist (string-to-list str)))
+                (if cur
+                    (if (member ?\) strlist)
+                        (cons () (cons (cons str cur) rest))
+                      (cons (cons str cur) rest))
+                  (if (and
+                       (member ?\( strlist)
+                       ;; If the close paren is present too, we have a whole
+                       ;; token
+                       (not (member ?\) strlist)))
+                      (cons (list str) rest)
+                    (cons () (cons (list str) rest))))))
+           exploded
+           :initial-value (list ()))))
+    (mapcar
+     '(lambda (xs)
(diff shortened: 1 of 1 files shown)


haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
+;;; haskell-navigate-imports.el --- A function for cycling through Haskell import lists -*- lexical-binding: t -*-
```
geraldus: Don't forget to change package name here (;

geraldus:
Also, would like to follow Emacs' built-in linter suggestions related to docstrings, e.g. all function arguments should be mentioned in docstrings (capitalized), separate sentences with two whitespace chars and etc.?

As for functionality, I have no strong opinion right now, I believe interactive mode suggests to import things automatically already (at least there was such functionality in past), but in case of non-interactive haskell-mode it looks quite useful.  One thing I'm often missing is some command to qualify/unqualify existing import, have you considered to write such function?

bch29:
This doesn't suggest adding imports, it just lets you add them (and sub-imports, and qualification) with a command instead of doing it manually. There's definitely some overlap with the automatic import adding though, and it might be useful to hook this into that by letting you choose precise sub-imports: as it stands, the imports added by the suggestions will only import the whole module, which is often not desirable.

Qualifying/unqualifying existing imports is something that could probably be done quite easily with this, since I already have functionality for changing existing imports in terms of adding sub-imports. I'll give that a go.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+          "\\(?:(\\(\\(?:[a-zA-Z0-9', \t\f\n\r]*\\|(.*?)\\)*\\))\\)?"
+          )))
+    (if (string-match regexp contents start)
+        (let ((is-qualified (match-string 1 contents))
+              (qualified-name (match-string 2 contents))
+              (imported-items-str (match-string 3 contents))
+              (start-point (match-beginning 0))
+              (end-point (match-end 0)))
+          (list start-point end-point
+                (if is-qualified t nil)
```
chrisbarrett: In elisp, any non-nil value is truthy, so you can replace this entire `if` expression with `is-qualified`.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+                qualified-name
+                (if imported-items-str
```
chrisbarrett: Use `when` instead of `if` if the alternative condition is always `nil`.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+
+(require 'haskell-navigate-imports)
+(require 'cl-lib)
+
+(defun haskell-auto-import-unqualified (module-name sub-imports)
+  "Add a new unqualified import of the provided module and sub-imports. If
+sub-imports is left blank, import everything in the module."
+  (interactive "MModule Name: \nMSub-Imports (comma separated, leave blank to import all): ")
+  (let* ((all-infos (haskell-auto-import-module-import-info module-name))
+         (non-qualified (member-if '(lambda (info) (not (car (cdr (cdr info))))) all-infos))
```
chrisbarrett: Use `cl-member-if`. No need to quote lambdas. Consider using `cl-destructuring-bind` or pattern matching with `pcase` instead of nested chains of `car`s and `cdr`s.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+         (new-sub-imports-list
+          (haskell-auto-import-parse-import-parens sub-imports)))
+    (if non-qualified
+        (let* ((info (car non-qualified))
+               (keep-info (butlast info 1))
+               (imported-items (car (last info 1)))
+               (new-info
+                (append keep-info
+                        (if new-sub-imports-list
```
chrisbarrett: Use `when` instead of `if` here, since the alternative case is always nil.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+                            (list
+                             (haskell-auto-import-normalize-imported-items
+                              (append imported-items new-sub-imports-list)))
+                          ()
+                            ))))
+          (haskell-auto-import-replace-import module-name new-info))
+      (progn
```
chrisbarrett: The `progn` here is redundant, because it surrounds a single form (the `let` expression).

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+  (haskell-auto-import-unqualified module-name ""))
+
+(defun haskell-auto-import-qualified (module-name qualifier)
+  "Add a new qualified import of the provided module with the provided
+qualifier. Attempts to place the new import next to existing imports of the same
+module if they exist."
+  (interactive "MModule Name: \nMQualifier: ")
+  (let* ((all-infos (haskell-auto-import-module-import-info module-name))
+         (qualified
+          (member-if '(lambda (info)
```
chrisbarrett: Again, use `cl-member-if` and don't quote lambdas. :)

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+each consist of '(imported-item sub-imports), where sub-imports is the list of
+items imported from the main item (e.g. in `import ModuleName(DataType(Con1,
+Con2))')
+
+Does NOT currently work properly with datatype operators with sub-imports (e.g.
+`import Foo.Bar ((**)(A, B))') or with sub-imports which are operators (e.g.
+`import Foo.Bar (A((&&), (||)))') "
+  (let* ((exploded (split-string contents "," nil "\\s-"))
+         (parts
+          (reduce
```
chrisbarrett: Use `cl-reduce`, don't quote lambda.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+           '(lambda (acc str)
+              (let ((cur (car acc))
+                    (rest (cdr acc))
+                    (strlist (string-to-list str)))
+                (if cur
```
chrisbarrett: Use `cond` instead of nested `if`s.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+          (member-if '(lambda (info)
+                        (string= (car (cdr (cdr info))) qualifier))
+                     all-infos)))
+    (if same-qualifier
+        (message "That module is already imported with the same qualifier. Doing nothing.")
+      (let ((old-point (point)))
+        ;; If the module is already imported in some capacity, put the new
+        ;; import above the existing ones
+        (if all-infos
+            (goto-char (+ 1 (car (car all-infos))))
```
chrisbarrett: There's a successor function called `1+` that you can use here. :)

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+                    (haskell-auto-import-parse-import-parens imported-items-str)
+                  nil)
+                )))))
+
+(defun haskell-auto-import-module-import-info (module-name &optional start)
+  "Returns information about every import of the given module after `start'
+(which is set to 0 if nil). Each member of the returned list has the same format
+as that returned from `haskell-auto-import-module-single-import-info'."
+  (let ((next-info (haskell-auto-import-module-single-import-info module-name start)))
+    (if next-info
```
chrisbarrett: Use `when`.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+        (end-point (car (cdr info)))
+        (rest-info (cdr (cdr info)))
+        (old-point (point))
+        )
+    (delete-region (+ 1 start-point) (+ 1 end-point))
+    (goto-char (+ 1 start-point))
+    (insert (haskell-auto-import-print-module-import module-name rest-info))
+    (goto-char old-point)))
+
+(provide 'haskell-auto-import)
```
chrisbarrett: Missing package footer here.

chrisbarrett:
Maimed you PR with tips and comments. :) 

As a general point, your code might be easier to work with if you use plists or alists to create associative data structures, much in the same way you would create dedicated data structures or records in Haskell instead of using raw lists. Every function in this package needs to know the precise ordering and meaning of the elements in the lists, which makes the internals brittle.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+                    (if (member ?\) strlist)
+                        (cons () (cons (cons str cur) rest))
+                      (cons (cons str cur) rest))
+                  (if (and
+                       (member ?\( strlist)
+                       ;; If the close paren is present too, we have a whole
+                       ;; token
+                       (not (member ?\) strlist)))
+                      (cons (list str) rest)
+                    (cons () (cons (list str) rest))))))
```
chrisbarrett: It's better to use `nil` instead of empty list literals `()` to represent falsey values. They mean the same thing, but `nil` is easier to spot since it's visually distinct from the rest of the parens in the code structure.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+
+import Data.Map (Map)
+import qualified Data.Map as Map"
+
+  (interactive
+   "MModule Name: \nMQualifier: \nMSub-Imports (comma separated, leave blank to import all): ")
+  (haskell-auto-import-qualified module-name qualifier)
+  (haskell-auto-import-unqualified module-name sub-imports))
+
+(defun haskell-auto-import-parse-import-parens (contents)
```
chrisbarrett: In modern elisp style, functions that are part of the package interface are named as `<packagename>-<fname>`, i.e. `haskell-auto-import-parse-import-parens`. Functions that are internal (helper functions and other such functions you don't want clients to depend on) have a double dash after the package name, as `<packagename>--<fname>`. This looks like an internal function, so I'd recommend renaming it to `haskell-auto-import--parse-import-parens`.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+    (if next-info
+        (let ((next-start (car (cdr next-info))))
+          (cons next-info (haskell-auto-import-module-import-info module-name next-start)))
+      ()
+      )))
+
+(defun haskell-auto-import-normalize-imported-items (imported-items)
+  "Normalize imported items, e.g. `one(a), two(a, b), two(a, c), one' --> `one,
+two(a, b, c)'"
+  (let ((grouped (-group-by 'car imported-items)))
```
chrisbarrett: I'd recommend using either `dash` or `cl-lib` functions, but not both. Be consistent. :)

Also, when passing function names as arguments, use sharp quotes (`#'symbol`), i.e.:
```lisp
(-group-by #'car imported-items
```
This provides a hint to the byte-compiler that this should resolve to a function at runtime, so you get a byte-compiler warning if that function isn't defined.

haskell-auto-import.el
```diff
@@ -0,0 +1,286 @@
...
+              ")")
+             ""))
+        ))
+   imported-items ", "))
+
+(defun haskell-auto-import-print-module-import (module-name info)
+  (let ((is-qualified (car info))
+        (qualified-name (car (cdr info)))
+        (imported-items (car (cdr (cdr info)))))
+    (concat
```
chrisbarrett: `format` might be easier to read than `concat` here if you let-bind some of the `if` expressions.

haskell-auto-import.el
chrisbarrett: I notice that haskell-mode doesn't use dash, so you'll have to stick to the built-ins.

gracjan:
Hi @bch29,

If this is really your first elisp code than congratulations, this is really good!

Having said that I think we are not able to pull in your code inside `haskell-mode`. The reasons are as follows:

Haskell Mode due to the size of code is already on the verge of collapsing under its own weight, as hinted by many people that felt intimidated to contribute to the code base. Haskell Mode has already a lot of functionality that is used approximately by zero people due to either:
1. They never discovered this functionality.
2. They forgot about it completely.
3. They remember it exists, but forgot which keybinding does what.
For example did you know that `haskell-indent-mode` offers `C-c C-=` or `C-c C-|` but I expect nobody uses it because due to the issues listed above.


Functionality in Haskell Mode needs to be:
1. Discoverable.
2. Memorable.

gracjan:
For more information about our philosophy what we accept and we sadly need to reject please read [Complete functionality](https://github.com/haskell/haskell-mode/wiki/Complete-functionality) in our wiki.

bch29:
@chrisbarrett Wow, thanks for the comprehensive feedback! I didn't expect this when I said feedback would be appreciated :-)

@gracjan Is the problem that you can't accept this much new code at all, or will this be acceptable if I make it conform to all of the points listed in the complete functionality page?

If it's the latter, can you give me any pointers on where in the haskell-mode code I should look when it comes to providing key bindings and menu options? I searched for quite a while, and couldn't find where most of the keys are bound. I could only find the bindings for the really core functionality like `haskell-process-load-file`.

gracjan:
@bch29, I think my point is more general than Complete functionality, although that is hygene factor I just think it is much easier to write `import X as Y (z)` than to remember the keybinding for that. So I do not see myself using your functions much.

Still, this is my personal opinion and if enough people chime in that they see using it as their daily tool then we can discuss.

Note though: there is quite some really good code you produced that could be used in other scenarios. Currently we have a mess in everything import/module related and your code could serve as building blocks to clean up those places.

bch29:
@gracjan the main reason I wrote this was that I was getting fed up of constantly having to write

```haskell
import Data.Map (Map)
import qualified Data.Map as Map

import Data.Set (Set)
import qualified Data.Set as Set

import Data.Vector (Vector)
import qualified Data.Vector as Vector

-- etc etc etc
```

at the top of almost every file. The rest of the functions were sort of a by-product of making that functionality work nicely even when existing imports are there. I suppose I can always just use it myself if you don't think most people would find it useful enough to remember the keybinding.

As for using this to contribute to other bits of the project, I'll definitely look into it!

gracjan:
@bch29, have you seen https://github.com/haskell/haskell-snippets ? It has some import related functionality.

Snippets are well known, just like ido, so the 'discoverability' problem goes away because people know how to use it.

gracjan:
I guess you could add a snippet to haskell-snippets that covers this specific use case.

Issue #1088, opened 2016-01-16
haskell-hoogle command's *hoogle* window results are not clickable


dmich2:
When I get `haskell-hoogle` results in the `*hoogle*` window (as opposed to opening hoogle results in browser), I am unable to get any further details on the choices shown. I can't select an entry and hit `<Return>` for example to see the details on the chosen function.

geraldus:
@dmich2, thanks.  However, now I'm thinking about what that further *details* should actually be, I mean what information you expect to see and where to get it from?

dmich2:
For example, if in a haskell buffer I type `he`, company-mode pops up a list that shows `head Prelude`. If I then hit `<F1>` it shows in `*company-documentation*` buffer:
```
Prelude head :: [a] -> a

Extract the first element of a list, which must be non-empty. 

From package base
head :: [a] -> a
```
See https://github.com/iquiw/company-ghc#34display-hoogle-document-as-doc-buffer for a screen-shot.

dmich2:
Possibly related to issue #215

geraldus:
@dmich2 good, I ask because I'm not using hoodle in Emacs, so I do not know what it can.  Maybe this is a time to get used.  So, we can get some detailed documentation from hoogle indeed.

@gracjan What will you say, it looks like reasonable addition to functionality?  I vote to clickable search results in hoogle buffer, it should not be hard to implement this I guess.

gracjan:
Can either of you or both write a more detailed specification how this should be implemented?

We need a spec that is detailed enough so it deserves 'well-defined-task' label.

geraldus:
@gracjan I did some investigation.

First,
> If prefix argument INFO is given, then `haskell-hoogle-command' is asked to show extra info for the items matching QUERY..

But it shows extra documentation only for single entry, not for whole list, e.g.
`(haskell-hoogle "map")` gives long list of matching results,
while `(haskell-hoogle "map" t)` shows documentation only for `Prelude.map`.

This function relies on **external** `hoogle` executable.

So, we can achieve our goal only if:
+ we'll write some `hoogle` output parsing
+ if there is an easy way to find identifiers in output, then we can make them clickable. The easiest way to implement navigation is to execute same `haskell-hoogle` command with clicked item as QUERY argument and prefix argument set to `t`, e.g.
```
haskell-hoogle "map":
Prelude map :: (a -> b) -> [a] -> [b]
Data.List map :: (a -> b) -> [a] -> [b]
…
[...]

PR #1167, opened 2016-02-21
Fix haskell-hoogle to work with buffer local var

luntain:
Read the `haskell-hoogle-command` variable before switching to the help window which removes the buffer local `haskell-hoogle-command` from the scope.

In other words: This is a tiny change, I just extracted the command text into a variable defined outside the context of the new buffer so that the right buffer local `haskell-hoogle-command` is used.


gracjan:
Great catch, thanks. Do you see any other things you could help with? #1088 is hoogle related, too.

PR #1168, opened 2016-02-21
Add test sources for haskell-trigger-suggestions

geraldus:


gracjan:
Great! Lets wait till we have some time to push this further, overvall the plan is more or less:

1. Create a testsuite matrix that will do basically: `GHC-version` X `Your-files`.
2. We get N tests out of that.
3. When new compiler is released or new warning is discovered we will add new version or new trigger file and regenerate.

The above will ensure that we support reasonable set of GHC versions and warnings.

Issue #1171, opened 2016-02-21
Support non-interactive completions


geraldus:
Let's:

+ [x] hardcode completion candidates for keywords
+ [x] introduce `haskell-completion-at-point` (language, pragma name, options, extensions, keywords) in `haskell-completions.el`
+ [x] add it to `completion-at-point-functions` in `haskell-mode.el`

For interactive part:
+ [x] rename `haskell-completions-sync-completions-at-point` to `haskell-sync-repl-completion-at-point`
+ [x] use `haskell-completion-at-point` internally
+ [x] replace `haskell-completions-at-point-function` in `completion-at-point-functions` with  `haskell-sync-repl-completions-at-point`

I'd prefer a bit longer names: `haskell-completion-completion-at-point` and `haskell-completions-sync-repl-completions-at-point`.

Docs:
+ take a look at docs, and make necessary changes and clarifications.
  + [x] find out if section 4.2.5 is relevant (#1182)
  + [x] find out contextual-space usage (#1182)
  + [x] update section 4.3:
    + [x] mention non-interactive completions
    + [x] mention keyword completions
    + [x] mention requirement of active process (session associated to buffer)

Related #215 #1169 #1184

geraldus:
@gracjan can you take care about section 4.3 of manual?

gracjan:
Okay. Maybe not that soon, but yes...

geraldus:
Cool, then I can proceed with some other stuff.

PR #1172, opened 2016-02-21
Support simple completions in non-interactive context.  Support keyword completions 

geraldus:
Related #1171

### DIFF
--- a/haskell-completions.el
+++ b/haskell-completions.el
@@ -300,15 +300,23 @@ PREFIX should be a list such one returned by
 
 
 (defun haskell-completions-completion-at-point ()
-  "Provide a list of completion candidates.
-This function is supposed to be used in non-interactive context.
-It provides completions for haskell keywords, language pragmas,
+  "Provide completion list for thing at point.
+This function is used in non-interactive `haskell-mode'.  It
+provides completions for haskell keywords, language pragmas,
 GHC's options, and language extensions, but not identifiers."
   (let ((prefix (haskell-completions-grab-prefix)))
     (haskell-completions--simple-completions prefix)))
 
 (defun haskell-completions-sync-repl-completion-at-point ()
-  "A `completion-at-point' function using the current haskell process.
+  "A completion function used in `interactive-haskell-mode'.
+Completion candidates are provided quering current haskell
+process, that is sending `:complete repl' command.
+
+Completes all possible things: everything that can be completed
+with non-interactive function
+`haskell-completions-completion-at-point' plus identifier
+completions.
+
 Returns nil if no completions available."
   (let ((prefix-data (haskell-completions-grab-prefix)))
     (when prefix-data
--- a/haskell-completions.el
+++ b/haskell-completions.el
@@ -1,6 +1,6 @@
 ;;; haskell-completions.el --- Haskell Completion package -*- lexical-binding: t -*-
 
-;; Copyright © 2015 Athur Fayzrakhmanov. All rights reserved.
+;; Copyright © 2015-2016 Athur Fayzrakhmanov. All rights reserved.
 
 ;; This file is part of haskell-mode package.
 ;; You can contact with authors using GitHub issue tracker:
@@ -63,8 +63,51 @@
         "WARNING")
   "A list of supported pragmas.
 This list comes from GHC documentation (URL
-`https://downloads.haskell.org/~ghc/7.10.1/docs/html/users_guide/pragmas.html'.
-")
+`https://downloads.haskell.org/~ghc/7.10.1/docs/html/users_guide/pragmas.html'.")
+
+(defvar haskell-completions--keywords
+  (list
+   "as"
+   "case"
+   "class"
+   "data family"
+   "data instance"
+   "data"
+   "default"
+   "deriving instance"
+   "deriving"
+   "do"
+   "else"
+   "family"
+   "forall"
+   "foreign import"
+   "foreign"
+   "hiding"
+   "if"
+   "import qualified"
+   "import"
+   "in"
+   "infix"
+   "infixl"
+   "infixr"
+   "instance"
+   "let"
+   "mdo"
+   "module"
+   "newtype"
+   "of"
+   "proc"
+   "qualified"
+   "rec"
+   "then"
+   "type family"
+   "type instance"
+   "type"
+   "where")
+  "A list of Haskell's keywords (URL `https://wiki.haskell.org/Keywords').
+Single char keywords and operator like keywords are not included
+in this list.")
+
 
 (defun haskell-completions-can-grab-prefix ()
   "Check if the case is appropriate for grabbing completion prefix.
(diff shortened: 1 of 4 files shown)


geraldus:
I'm modifying REPL completion function right now and can push few commits here.

gracjan:
Good stuff. Can you remove these things from `haskell-interactive-mode`?

geraldus:
Ok it looks like to work properly.

@gracjan didn't catch your point about interactive mode, what should I remove? (:

haskell-completions.el
```diff
@@ -220,36 +274,71 @@ result only if prefix length is not less than MINLEN."
...
-            (list beg end lst)))))))
+        (when (not (eql typ 'haskell-completions-general-prefix))
+          ;; do not complete things in comments
+          (if (cl-member
+               typ
+               '(haskell-completions-pragma-name-prefix
+                 haskell-completions-ghc-option-prefix
+                 haskell-completions-language-extension-prefix))
+              ;; provide simple completions
+              (haskell-completions--simple-completions prefix-data)
```
gracjan: As I understand there will be two completions hooks installed at the same time, one by `haskell-mode` other by `interactive-haskell-mode`. The non-interactive one will return completions for PRAGMA, GHC_OPTION and LANGUAGE, so there is no need to return those completions here again.

Unless I'm mistaken how the mechanism works.

geraldus: @gracjan I tested this, and if first function in hook returns `nil` there will no completions.  I tested this creating two dummy completion functions and added both of them to `completion-at-point-functions` hook in elisp mode.  All time only first function was used to return completion candidates.  I don't know why it works this way though.

geraldus:
As for `haskell-interactive-mode-completion-at-point-function`, to remove it we have to support REPL command completions.  `haskell-completions-sync-repl-completions` can give right candidates if we take care of `:` when grabbing prefix in REPL buffer.

fice-t:
Should the hardcoded completions have extra information for support for `eldoc-mode`? That is, in `haskell-doc-mode`,  [there are a few huge hardcoded alists](https://github.com/haskell/haskell-mode/blob/master/haskell-doc.el#L465) that are currently outdated.

I was thinking that you make new giant alists so that `eldoc` could use both keys and values, where the completion would just use the keys.

geraldus:
@fice-t hmm, I didn't catch your idea, if we do have such list already why we need another one?

gracjan:
I wonder how useful the hints by haskell-doc-mode are. Old timers know basic language constructs by heart, can newbies sort out keyword usage from the very short descriptions provided by haskell-doc-mode?

geraldus:
I like doc messages for top-level forms, but keyword doc strings seem for me quite confusing :D

geraldus:
By the way, @gracjan in Emacs 25 I have colored message in mini-buffer from `eldoc`, should we follow this (I remember in past you suggested not to fontify haskell-doc messages)?

fice-t:
The lists are old as they're for Haskell-98. According to git they haven't been updated in over a decade.

geraldus:
@fice-t I agree, I saw that before, I just do not understand what benefits of extra information in hard-coded lists?

fice-t:
The extra information would be used in `eldoc` hints. For example:

`("filter" . "(a -> Bool) -> [a] -> [a]")`
The completion could just use `filter`, but `eldoc` could use both to create `filter :: (a-> Bool) -> [a] -> [a]`. The same for the keywords (though that might not be as useful).

So it's just using the same list for different purposes.

geraldus:
@fice-t, pal, I'm still confused.  Look, `haskell-completions--pragma-names` and `haskell-completions--keywords` lists are used only to provide completion candidates, that is either when `complete-symbol` function is invoked, or completion functions from `haskell-completions.el`.  They have no relation to `haskell-doc.el`, it have separate hard-coded lists, and I see doc-strings in mini-buffer for keywords and top-level form when haskell session is present (and in my configuration `haskell-doc` mode is not enabled, `Global-Eldoc` somehow magically gets information from `haskell-doc` itself).  Do you facing some issues with `haskell-doc`?

haskell-completions.el
gracjan: I've repeated this experiment and it works like described in Emacs docs (and unlike you say). That means I'm able to complete "1" to "123" and "ab" to any of the options listed. Emacs seems to work correctly.

gracjan:
I still think the static and the interactive completion functions should be completely separate.

For the interactive case there should be only `haskell-completions-grab-identifier-prefix` called, this should return `nil` if inside of a comment.

For the static case there should be only `haskell-completions-grab-pragma-prefix` called, this should return `nil` if outside of a pragma comment.

Emacs will call the hooks one by one, stopping with the one that returns non-nil value. This seems perfectly reasonable to me.

Am I missing something?

geraldus:
@gracjan sounds reasonable, but
1. you forgot about keywords in static case

> Emacs will call the hooks one by one

2. can you please test this (as I mentioned before in my tests second completion function was not triggered in case when first returned `nil`, I updated my comment previous comment with code I've used to tests).

fice-t:
>  Look, haskell-completions--pragma-names and haskell-completions--keywords lists are used only to provide completion candidates, that is either when complete-symbol function is invoked, or completion functions from haskell-completions.el.

That's why I'm saying this. I'm saying (I believe) there could be a single list which both completion and hints could use in some cases, such as keywords and basic/Prelude functions. That might be beneficial as one could maintain just a single list (and periodically generate a separate completion list if one was to speed completion up).

> Global-Eldoc somehow magically gets information from haskell-doc itself)

Yeah, `eldoc-documentation-function` is bound to a function that calls another function that checks the hardcoded `haskell-doc` lists.

geraldus:
@fice-t ok, now I got it :D 
First of all, both lists are incomplete in our case according to wiki page, they do not include operator like things like `<-`, `->` and friends.  While it feel sensible to add this stuff to `haskell-doc.el`, this is not a case for completions.  Secondly I added to completion list special entries such `data family`, `import qualified`. Why?  For example type `im`, in case of `company-mode` this will give you both `import` and `import qualified` candidates in list, in case of vanilla Emacs hitting <kbd>TAB</kbd> or <kbd>C-M-i</kbd> first expand it to `import`, consequential command invocation will pop-up a Completion List buffer where both candidates are listed.  In we'll use keywords from `haskell-doc` we loose this feature, though it's possible to have a tiny list with such improved candidates and append it haskell-doc list.

geraldus:
@gracjan this is my thought about your proposal.

1. I suppose that keywords should be completed in both cases
2. there is `:exclusive` property allowing to let Emacs run next completion function in case when it can't find matching candidate in list returned by current function.  Initially I though this could solve our puzzle, however, if static function will return keywords list with `:exclusive no` property user won't get other interactive completions if thing to be expanded matches some keyword, that is
`ca` with give sole completion `case`, but it is possible, that interactive function could return much more candidates, only when user will type `car` Emacs will call interactive completions.
3. if we follow you proposal we have to make sure that interactive completion function should be preceded by static, this will require first remove static function from hook and then add both in desired order.  Current implementation requires reversed order, e.g. interactive function should succeed static one and this is done out-of-box.

gracjan:
@geraldus: I think you convinced me that Emacs default handling of `competion-at-point-functions` is not appropriate and haskell-mode should handle completions on its own.

In this case you should implement it as you have done. Ping me when the PR is ready to merge.

PR #1174, opened 2016-02-22
[WIP] Use single session per stack project

corngood:
I have a stack project with a few local packages that I work on together.  I found that haskell-mode would create a `stack ghci` process/buffer for each project, and each one would load all the projects, as stack ghci does.

This change defaults each module under the stack project to the same ghci instance.

What do you think?  I haven't had any problems with it, and it saves a ton of memory for ghci processes, as well as simplifying work on multiple packages.

WIP because the buffer name is inconsistent with cabal, but using the package name on its own seems kind of fragile too...  Maybe it should be the full path to the stack.yaml / .cabal with some sort of prefix?

TODO: share the file logic with (haskell-process-type)

### DIFF
--- a/haskell-session.el
+++ b/haskell-session.el
@@ -80,11 +80,13 @@
 
 (defun haskell-session-default-name ()
   "Generate a default project name for the new project prompt."
-  (let ((file (haskell-cabal-find-file)))
-    (or (when file
-          (downcase (file-name-sans-extension
-                     (file-name-nondirectory file))))
-        "haskell")))
+  (if (eq 'stack-ghci (haskell-process-type))
+      (locate-dominating-file default-directory "stack.yaml")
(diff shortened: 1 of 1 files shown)


corngood:
TODO: The directory for the ghci buffer ends up being set to the project where it is first created.  It doesn't seem to matter to stack ghci, but having it run in the root of the stack project would be cleaner.

gracjan:
Looks good to me. Ping us when you finish.

corngood:
It's working for my purposes, but it may be complicated by certain things assuming there is a cabal-dir per session (e.g. tag generation).  I'll dig into it a bit more, but it's not as close to complete as I hoped.

I'll let you know.

purcell:
See also #1547.

webdevred:
I support the idea behind this idea, would be great.

Issue #1133, opened 2016-02-04
Implement electric comma, semicolon, etc


gracjan:
`haskell-indentation` reacts to a couple of characters, most notably comma and semicolonbut also other chars.

Implement function that can be used in `post-command-hook` that will:
- [ ] check if there is exactly one character on the line
- [ ] check if there is only one indentation position returned from `haskell-indentation-find-indentations`
- [ ] do the indentation if both are true
- [ ] connect this function to `,;` and `)}]`

More info:
http://www.gnu.org/software/emacs/manual/html_node/ccmode/Electric-Keys.html

bch29:
This function does the job, but unfortunately I'm not familiar enough with emacs to know how to hook it into electirc keys:

```elisp
(defun haskell-try-auto-indent ()
  (if (and
       (= 1 (length (s-trim (current-line))))
       (= 1 (length (haskell-indentation-find-indentations))))
      (haskell-indentation-indent-line)))
```

gracjan:
`cc-mode` knows how to do electric keys. Can you look there?

gracjan:
This is great find: https://github.com/haskell/haskell-mode/pull/1175#issuecomment-187692908 .

fice-t:
I don't believe that we should put this in `post-command-hook` (as far as I can tell, `cc-mode` doesn't use that approach either). I'm unsure about using `electric-indent-mode` but it does make the code shorter. The following should work:

```haskell
(defun haskell-electric-keys (char)
  "Indent after typing CHAR when:
1) There is no other character on the current line.
2) There is only one possible indentation position.
3) The variable `haskell-use-electric-keys' is either t or, if it's a list, CHAR is in it."
  (and (if (listp haskell-use-electric-keys)
           (memq char haskell-use-electric-keys)
         haskell-use-electric-keys)
       (string-empty-p (string-trim-left (thing-at-point 'line t)))
       (= 1 (length (save-excursion
                      (move-to-column (haskell-indentation-current-indentation))
                      (haskell-indentation-find-indentations))))))

(defcustom haskell-use-electric-keys t
  "If t, various special keys will be 'electric' when `electric-indent-mode` is turned on. 
If a list, only the special characters present in the list will be 'electric'."
  :group 'haskell-indentation
  :type '(choice boolean (repeat character)))

(setq-local electric-indent-chars (string-to-list ",;=|)}]"))
(setq-local electric-indent-functions #'haskell-electric-keys)
```
[...]

PR #1175, opened 2016-02-22
Update haskell-indentation-indent-line

### LINKED ISSUE #896: Redefining evil-open-above for Haskell (spacemacs)
Yet another electric-indent problem:

https://github.com/syl20bnr/spacemacs/issues/3162

fice-t:
Two changes:

1. Some Evil commands are now treated differently to achieve expected
behaviour.

2. Pressing TAB at the final indent location now starts to move to the
previous indent position instead of relying on the reindent function to
start this process.

Fixes #896

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -243,9 +243,9 @@ indentation points to the right, we switch going to the left."
   ;; try to repeat
   (when (not (haskell-indentation-indent-line-repeat))
     (setq haskell-indentation-dyn-last-direction nil)
-    ;; parse error is intentionally not cought here, it may come from
+    ;; parse error is intentionally not caught here, it may come from
     ;; `haskell-indentation-find-indentations', but escapes the scope
-    ;; and aborts the opertaion before any moving happens
+    ;; and aborts the operation before any moving happens
     (let* ((cc (current-column))
            (ci (haskell-indentation-current-indentation))
            (inds (save-excursion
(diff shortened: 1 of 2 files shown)


gracjan:
Hmm, I have bad feelings about this. I'm unable to follow the logic neither in the code nor in the test.

1. Hardcoding evil commands is super fragile. Can't we have logic that works properly without  this?
2. Can the test be more declarative? I would like to see more tests written in a more declarative manner, for example as a list of tuples: (COMMAND-TO-EXECUTE WHERE-SHOULD-INDENT-END), like this:

```
    '((indent-right 7)
      (indent-left 12) 
       ...)
```

That way it would be easier to create more than one scenario.

fice-t:
The cycle logic is that `haskell-indentation-indent-line`, on the last tab indent position, moves to the 2nd last indent position (if it exists) on the very first invocation. Before, it only moved on the 2nd call as then it passes control to `haskell-indentation-indent-line-repeat`.

1. Certainly it's fragile (the `this-command` part), but I don't really see an obvious better way. We could just set `indent-line-function` to a new function that does it how Evil commands expect and then bind TAB directly to the old `haskell-indentation-indent-line`. Might not be the best thing to do though.

2. I suppose. I tried to make the test so that it can be made general with a nice macro or something. Would a macro wrapper be alright?

geraldus:
Will `electric-indent-functions` hook be helpful in our case?  Looks like `haskell-mode` totally ignores this.

> electric-indent-functions is a variable defined in ‘electric.el’.
> Its value is nil

>  This variable may be risky if used as a file-local variable.

> Documentation:
Special hook run to decide whether to auto-indent.
Each function is called with one argument (the inserted char), with
point right after that char, and it should return t to cause indentation,
‘no-indent’ to prevent indentation or nil to let other functions decide.

----

> electric-indent-mode is an interactive compiled Lisp function in ‘electric.el’.

> (electric-indent-mode &optional ARG)

> Toggle on-the-fly reindentation (Electric Indent mode).
With a prefix argument ARG, enable Electric Indent mode if ARG is
positive, and disable it otherwise.  If called from Lisp, enable
the mode if ARG is omitted or nil.

> When enabled, this reindents whenever the hook ‘electric-indent-functions’
returns non-nil, or if you insert a character from ‘electric-indent-chars’.

> This is a global minor mode.  To toggle the mode in a single buffer,
use ‘electric-indent-local-mode’.

gracjan:
O! Nice find!

Maybe we will be able to pacify `electric-indent-mode` so it works when <kbd>RET</kbd> is pressed (or `evil-open-below`).

gracjan:
@geraldus: Your comment is priceless for #1133.

geraldus:
Glad to be helpful! :sunny:

gracjan:
I'm closing this without merging because I think that #1265 is fundamentally better idea.

Thanks for trying, this pr really helped to get the thinking going.

PR #1190, opened 2016-03-05
Add a test for indentation broken on a quasiquote

neongreen:


gracjan:
Nice. Can you write `29c*` instead of `29c` so that we know this is a failing test case?

neongreen:
Done.

gracjan:
Nice! Thanks.

Now you can try to fix it :) Should not be that hard, what you need is fix around `haskell-indentation-peek-token` and `haskell-indentation-skip-token` in https://github.com/haskell/haskell-mode/blob/master/haskell-indentation.el#L1129

Issue #1182, opened 2016-02-24
[Discussion] Deprecate haskell-mode-contextual-space


geraldus:
This is from online haskell-mode manual

> 4.2.5 Entering imports
>
> If you have Haskell Interactive Mode enabled and a session open, you can write
>
> import
> 
> And then hit SPC (running M-x haskell-mode-contextual-space) which will prompt a list of all modules, either installed modules or local project directory files. Then it’ll insert it and then format that module section.

`haskell-mode-contextual-space`<sup>[[1]]</sup> seems not to be used at all these days and is not bound to any key sequence<sup>[[2]]</sup>.  It provides following functionality:
+ prompt user to select module name when pressed <kbd>SPC</kbd> after `import` keyword (using `haskell-complete-module-read`)
  + it does not provide qualified import support
  + modules list comes from session
+ echoes identifier info to mini-buffer (using `haskell-process-do-try-info`)
+ it requires interactive haskell mode to be enabled and session to be present, otherwise it does nothing

Module import is simplified by auto completions.  Identifier info in mini-buffer is questionable.  I'd rather see identifier type signature.  But since it is not used in any other function, not bound to any key sequence, and do not provide any unique feature should we ~~drop it completely~~ deprecate it along with `haskell-mode-contextual-import-completion`<sup>[[3]]</sup>?

[1]:https://github.com/haskell/haskell-mode/blob/bcfe134622003fe840e7dd8127beb83c0418cc5e/haskell.el#L337
[2]:https://github.com/haskell/haskell-mode/search?utf8=%E2%9C%93&q=haskell-mode-contextual-space
[3]:https://github.com/haskell/haskell-mode/blob/bcfe134622003fe840e7dd8127beb83c0418cc5e/haskell-mode.el#L659

gracjan:
Agreed. Last time a guy tried to use it he was burned in #572. Just remove it, no deprecation needed.

geraldus:
Will do in few hours

geraldus:
@gracjan by the way how long we keep obsolete code in?

gracjan:
We do not have specific rules about deprecation.

PR #1191, opened 2016-03-05
Remove obsolete code

geraldus:
Related #1182


gracjan:
This is too fast. Can we have a pull request that removes just contextual space and nothing else?

geraldus:
:D surely

geraldus:
@gracjan done

Issue #1189, opened 2016-03-05
Indentation is broken for quasiquotation with quote symbols


neongreen:
~~~
foo = do
  let bar = [text|"some text"|]
  button "Cancel" $ do
                _
~~~

After pressing Enter after `do`, the cursor is at the position marked with `_`, which is rather wrong.

This doesn't happen when:

  * quasiquotation isn't present
  * or `"Cancel"` is removed
  * or quotes around `"some text"` are removed

gracjan:
Thanks for the report. Can you add a test case in `tests/haskell-indentation-tests.el`?

neongreen:
Sure.

neongreen:
@gracjan: Could you help me a bit?

`haskell-indentation-skip-token` has this code:

~~~ elisp
            ;; QuasiQuotes, with help of propertize buffer and string delimeters
            (looking-at "\\s\"\\S\"*\\s\"")
~~~

When the point is here, `\s"` treats `|` as a quote character:

~~~
     v
[text|foo "bar"|]
~~~

~~~
> (looking-at "\\s\"")
t
~~~

And when it's inside the quote, it doesn't treat `|` as a quote character anymore but it treats `"` as a quote character:

~~~
          v
[text|foo "bar"|]
~~~

~~~
> (looking-at "\\s\"")
t
~~~

(Moreover, `|` isn't treated as a quote-character outside of the quasiquotation context.)

So my guess is that thanks to this `haskell-indentation-skip-token` treats the string between `|` and `"` as a single token, which is incorrect. I have several ideas about how this could be fixed, but it's hard for me to choose because I don't understand how the meaning of `\s"` can change dynamically:

  * If meaning of `\s"` doesn't change *while* regexp consumes the string, I could make it so that `\s"` matches `|` but doesn't match `"` (when the point is after `[text`) and so `\s"\S"*\s"` wouldn't stop at the `"` in the quasiquote. (Where can it be changed, tho?)

  * If meaning of `\s"` can change while regexp consumes the string, I could write something like `\(\s"\) ???* \1` (I don't know how to say “any character that isn't the same as the quote character matched previously”, tho).

  * Can we somehow check whether we're at the beginning of a quasiquote and then write a regexp that would match `|`s explicitly without the `\s"` shenanigans? (But then I don't know how to write such a regexp, because it needs to detect `|]` and Emacs regexps don't have lookahead.)

  * Maybe `\s$` (paired delimiters) could help?

[...]

PR #1196, opened 2016-03-06
Treat insides of quasiquotes as strings

neongreen:
See #1189. Since the insides of quasiquotes are regarded as ordinary characters now, `\s"` no longer stumbles on `"`s inside quasiquotes.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -587,7 +587,13 @@ executable found in PATH.")
                 (put-text-property (match-beginning 3) (1+ (match-end 3)) 'syntax-table (string-to-syntax "|")))))
            ((equal token-kind 'template-haskell-quasi-quote)
             (put-text-property (match-beginning 2) (match-end 2) 'syntax-table (string-to-syntax "\""))
-            (put-text-property (match-beginning 4) (match-end 4) 'syntax-table (string-to-syntax "\""))))
+            (put-text-property (match-beginning 4) (match-end 4) 'syntax-table (string-to-syntax "\""))
+            (save-excursion
+              (goto-char (match-beginning 3))
+              (let ((limit (match-end 3)))
+                (save-match-data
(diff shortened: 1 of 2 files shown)


neongreen:
By the way, it also gives 2 possible indentations in this situation (previously it was just position 0, now it's positions 0 and 11):

~~~
(hindent-test "29d unfinished quasiquote" "
foo = [text|some
"
              (2 0 11))
~~~

This doesn't seem bad to me, tho.

gracjan:
It is good so.

haskell-mode.el
```diff
@@ -587,7 +587,8 @@ executable found in PATH.")
                 (put-text-property (match-beginning 3) (1+ (match-end 3)) 'syntax-table (string-to-syntax "|")))))
            ((equal token-kind 'template-haskell-quasi-quote)
             (put-text-property (match-beginning 2) (match-end 2) 'syntax-table (string-to-syntax "\""))
-            (put-text-property (match-beginning 4) (match-end 4) 'syntax-table (string-to-syntax "\""))))
+            (put-text-property (match-beginning 4) (match-end 4) 'syntax-table (string-to-syntax "\""))
+            (put-text-property (match-beginning 3) (match-end 3) 'syntax-table (string-to-syntax "."))))
```
gracjan: This is a cannon used on a fly and will break some other functionality, like navigation word-by-word inside quasiquote.

Better: loop using re-search-forward to find all `"` and put `.` syntax only on those.

gracjan:
@neongreen, I'd love to merge this PR, but I need you to fix what I mentioned. Can you get around to fixing this issue soon?

neongreen:
@gracjan Ouch, sorry! I didn't have time before and then I forgot about it. I have fixed it now.

gracjan:
Perfect, thanks!

PR #1199, opened 2016-03-07
Add c2hs mode

sergv:
This is a wrapper around haskell-mode, that mostly adds fontification for {#...#} blocks.

### DIFF
--- a/haskell-c2hs.el
+++ b/haskell-c2hs.el
@@ -1,0 +1,200 @@
+;; haskell-c2hs.el --- -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2016 Sergey Vinokurov
+;;
+;; Author: Sergey Vinokurov <serg.foo@gmail.com>
+;; Created: Monday,  7 March 2016
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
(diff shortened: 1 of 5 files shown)


haskell-c2hs.el
```diff
@@ -0,0 +1,203 @@
...
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; This mode is mostly indended for highlighting {#...#} hooks.
```
geraldus: `indended` — typo? (:

sergv:
Fixed typo in the description.

geraldus:
Would you like to add some tests?

sergv:
I think I'll add some font-lock tests

gracjan:
Yes, please add some tests. These are required, otherwise we will not be able to take responsibility of this module.

And thanks for great contribution!

sergv:
I added tests for all hooks currently colorized and rebased against latest master

gracjan:
Nice. Lets wait 2 days before merging so that people have time to look and test the code.

gracjan:
Okay, so this merge wasn't the cleanest one and we had issue reports that melpa compiled version did not load.

1. All symbols, including `defface` should start with `haskell-`.
2. `autoloads.el` should actually load, it did not in this case.
3. Overuse of macros is not a good sign.

@sergv: Can you inline the `haskell--let-rx` macro? It overcomplicates things. Please also add `haskell-` prefix to `defface`s. Thanks.

gracjan:
The `autoloads` issue will never happen again thanks to #1211.

sergv:
@gracjan I'll inline the macro and add prefixes, stay tuned!

sergv:
@gracjan why are macros not a good sing? The haskell--let-rx was introduced to make writing regexps via standard rx macro easier and make them both expand at compile time. It's easy to inline, but the advantage of compile time execution will be lost.

gracjan:
Macros are another layer of abstraction that a potential contributor needs to understand. As such it poses a possible mental roadblock which we would rather like to avoid.

sergv:
Fair enough

PR #1203, opened 2016-03-10
Small improvements of completion functions

geraldus:
As a user of company mode I see errors in mini-buffer very often:

> Company: An error occurred in auto-begin
> Wrong type argument: stringp, nil

Small investigation lead me to proposed changes

### DIFF
--- a/haskell-completions.el
+++ b/haskell-completions.el
@@ -286,17 +286,18 @@ GHC's options, and language extensions.
 PREFIX should be a list such one returned by
 `haskell-completions-grab-identifier-prefix'."
   (cl-destructuring-bind (beg end _pfx typ) prefix
-    (let ((candidates
-           (cl-case typ
-             ('haskell-completions-pragma-name-prefix
-              haskell-completions--pragma-names)
-             ('haskell-completions-ghc-option-prefix
-              haskell-ghc-supported-options)
-             ('haskell-completions-language-extension-prefix
-              haskell-ghc-supported-extensions)
-             (otherwise
-              haskell-completions--keywords))))
-      (list beg end candidates))))
+    (when (not (eql typ 'haskell-completions-general-prefix))
+      (let ((candidates
+             (cl-case typ
+               ('haskell-completions-pragma-name-prefix
+                haskell-completions--pragma-names)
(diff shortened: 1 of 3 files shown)


gracjan:
Is there a way to trigger this `nil` is some kind of test?

geraldus:
@gracjan didn't understood your question well.  I've faced the situation, when GHCi fails to start:

> The following modules are present in multiple packages:
 * Internal.API (in solitaire-ghcjs-simple, ghcjs-perch-canvas)
 * Internal.FFI (in solitaire-ghcjs-simple, ghcjs-perch-canvas)
 * Internal.Type (in solitaire-ghcjs-simple, ghcjs-perch-canvas)

In this case REPL process is alive, and not marked as restarting and all requests to REPL returns `nil`.  As a consequence, response parser passes this `nil` value to `split-string` function which leads to error.  I have no strong opinion what to do with `nil`-ish responses.

geraldus:
Oops.  I've found another one bug, force-pushed new changes.

TLDR: first fixes for static completion function added condition case for general type of prefix with nil value associated as result.  This was wrong, because instead of just `nil` value it returned a list of prefix boundaries positions and `nil`, making impossible other backends to proceed.

purcell:
+1 for these fixes: `completion-at-point` in company is very broken without them. I was about to submit something like 4c6b3ffa6b295d3d9ae7cbacfef6a68812d021c7 myself.

geraldus:
@purcell by the way, do you have some special setup for company mode and haskell?

purcell:
Not really -- I mostly rely on `company-ghci` and `company-capf` in `haskell-mode`.

geraldus:
@purcell I have found useful to override `company-backends` in haskell buffers, e.g.
```el
(set (make-local-variable 'company-backends)
       (append '((company-capf company-dabbrev-code) company-dabbrev)
               company-backends))
```
This combines completions from haskell mode with dabbrev-code completions outside strings and comments (useful for local identifiers defined in where and let blocks) and provides dabbrev completions otherwise.  I want to put it in docs.

purcell:
@geraldus Good tip, thanks.

PR #1204, opened 2016-03-13
Invoke hasktags more directly

geraldus:
#964

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -1079,52 +1079,45 @@ buffer not visiting a file returns nil."
       (when buffer-file-name
         (file-name-directory buffer-file-name))))
 
-(defun haskell-cabal--compose-hasktags-command (&optional cabal-dir)
-  "Prepare command to execute hasktags for current file.
+(defun haskell-cabal--compose-hasktags-command (dir)
+  "Prepare command to execute `hasktags` command in DIR folder.
 By default following parameters are passed to Hasktags
 executable:
 -e - generate ETAGS file
 -x - generate additional information in CTAGS file.
 
-Tries to find cabal file location, give optional CABAL-DIR parameter to
-override it.  If cabal file not found uses current file
-directory if current buffer is visiting a file.  Otherwise returns nil.
-
 This function takes into account user's operation system: in case
-of Windows it generates simple command like
+of Windows it generates simple command, relying on Hasktags
+itself to find source files:
 
 hasktags --output=DIR\TAGS -x -e DIR
-
-relying on Hasktags itself to find source files;
 
 In other cases it uses `find` command to find all source files
 recursively avoiding visiting unnecessary heavy directories like
 .git, .svn, _darcs and build directories created by
-cabal-install, stack, etc."
-  (let ((dir (or cabal-dir (haskell-cabal--find-tags-dir))))
-    (when dir
-      (if (eq system-type 'windows-nt)
-          (format "hasktags --output=\"%s\\TAGS\" -x -e \"%s\"" dir dir)
-        (format "cd %s && %s | %s"
-                dir
-                (concat "find . "
-                        "-type d \\( "
-                        "-path ./.git "
-                        "-o -path ./.svn "
-                        "-o -path ./_darcs "
-                        "-o -path ./.stack-work "
-                        "-o -path ./dist "
-                        "-o -path ./.cabal-sandbox "
-                        "\\) -prune "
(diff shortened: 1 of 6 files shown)


haskell.el
```diff
@@ -354,8 +354,11 @@ If `haskell-process-load-or-reload-prompt' is nil, accept `default'."
 (defun haskell-mode-after-save-handler ()
   "Function that will be called after buffer's saving."
   (when haskell-tags-on-save
-    (ignore-errors (when (and (boundp 'haskell-session) haskell-session)
-                     (haskell-process-generate-tags))))
+    (ignore-errors
+      (if (and (boundp 'haskell-session)
```
gracjan: Is there a need for both functions?

geraldus: Well, yes, at least for now, look at this
https://github.com/haskell/haskell-mode/blob/master/haskell.el#L351

geraldus: But we can rewrite jump-to-tag also.

geraldus:
> But we can rewrite jump-to-tag also.

Ok, I'll do that now

haskell.el
gracjan: Jump to tag generates tags? That is surprising.

geraldus: Well, this was my not idea :D
But I think it's not so bad at the end, otherwise user have to generate tags manually.  Do you think we should remove tags generation from jump-to-tag?  _(I suppose there are more surprising things awaiting :D)_

geraldus:
@gracjan done

gracjan:
Looks great!

I wonder if we could have a testing framework for all these external programs that we invoke here and there. We could have some integration tests in `.travis.yml`, a category of tests that is completely absent from our test suite right now.

There are a couple of ways around that issue:

1. Use `el-mock` and stop right before the external program is executed.
2. Create substitute programs as shell scripts, for example `hasktags-mock.sh` and setup `haskell-mode` to invoke those.
3. Install real programs and run those.

I would hate if `hasktags` suddenly stopped working due to some change that we make in good faith.

gracjan:
I've filled in #1206 to track the integration testing stuff.

geraldus:
@gracjan what about this one?

gracjan:
I've done code review and code is good. I did not have time to actually run the code though. Is there anybody else beside you @geraldus that has checked if it really works?

haskell-cabal.el
```diff
@@ -1052,6 +1052,62 @@ source-section."
...
+(defun haskell-cabal--find-tags-dir ()
+  "Return a directory where TAGS file will be generated.
+Tries to find cabal file first and if succeeds uses its location.
+If cabal file not found uses current file directory.  If current
+buffer not visiting a file returns nil."
+  (or (haskell-cabal-find-dir)
+      (when buffer-file-name
+        (file-name-directory buffer-file-name))))
+
+(defun haskell-cabal--compose-hasktags-command (&optional cabal-dir)
```
gracjan: Please make `cabal-dir` non-optional, always require this to be a proper directory.

haskell-cabal.el
```diff
@@ -1052,6 +1052,62 @@ source-section."
...
+
+hasktags --output=DIR\TAGS -x -e DIR
+
+relying on Hasktags itself to find source files;
+
+In other cases it uses `find` command to find all source files
+recursively avoiding visiting unnecessary heavy directories like
+.git, .svn, _darcs and build directories created by
+cabal-install, stack, etc."
+  (let ((dir (or cabal-dir (haskell-cabal--find-tags-dir))))
```
gracjan: This logic can go away when argument `cabal-dir` is required. One less failure scenario :)

haskell-cabal.el
```diff
@@ -1052,6 +1052,62 @@ source-section."
...
+    (when dir
+      (if (eq system-type 'windows-nt)
+          (format "hasktags --output=\"%s\\TAGS\" -x -e \"%s\"" dir dir)
+        (format "cd %s && %s | %s"
+                dir
+                (concat "find . "
```
gracjan: Cool stuff!

gracjan:
@geraldus, after you fix the one thing I found then we can merge this. Rest looks great.

haskell-cabal.el
geraldus: Well, I just tuned this a bit.

geraldus:
:tea: All is green now :kissing:

gracjan:
Nice!

Issue #1209, opened 2016-03-14
Error: (void-function haskell--rx-let)


PierreR:
I have just updated `haskell-mode` to `20160314.416` from elpa  and I have got the following error:
```
Error: (void-function haskell--rx-let)
```
It is probably due to a recent commit because if I use `20160309.237` the error disappears and I can use `haskell-mode` again.

PS: using `spacemacs with NixOS 15.09`

nomicflux:
Same issue, using `spacemacs` on `Kali 2.0` (which I believe uses packages from `Debian "jessie"`), as well as on `Windows 10`.

gracjan:
Already fixes with 86b6ffd148451e313b7cacaafd614d1a49ee3c98

nomicflux:
Made the same change in haskell-c2hs.el, removing `;;;###autoload` in front of `c2hs-font-lock-keywords `; still getting `(void-function haskell--rx-let)` when loading Emacs, rendering haskell-mode unusable.

gracjan:
You needs also run `rm haskell-mode-autoloads.el` and `make haskell-mode-autoloads.el`

nomicflux:
Perfect - that worked.

PR #1211, opened 2016-03-14
Makefile dependency tracking improvements

gracjan:
Related #1209.


gracjan:
Is Travis broken? Why doesn't it pick up these changes?

PR #1212, opened 2016-03-15
C2hs fixes

sergv:
Followup for #1199


gracjan:
Thanks.

I know this is uglier than the version before, but it should be easier to modify for new contributors.

PR #1224, opened 2016-03-20
Handle operators in haskell-ident-pos-at-point too

geraldus:
Let's see will it pass the tests.  I think I can add some tests for operators too.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -599,7 +599,7 @@ executable found in PATH.")
             (goto-char end)))))))
 
 (defun haskell-ident-at-point ()
-  "Return the identifier under point, or nil if none found.
+  "Return the identifier near point going backward or nil if none found.
 May return a qualified name."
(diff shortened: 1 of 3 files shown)


geraldus:
And it turns that we can drop `haskell-completions-general-prefix` due to prefix grabber no longer handles comments and strings (this will be next PR).

geraldus:
Don't merge this yet, this implementation is buggy :(

gracjan:
I think this function should be compatible with `thing-at-point`. That one seems to be standard.

geraldus:
@gracjan I agree, but not this time, ok? (:

geraldus:
@gracjan please overview

gracjan:
Looks good to me. Lets wait the usual two days so that other people may look at the code.

geraldus:
@gracjan no one going to test this :relieved:

gracjan:
Merged.

Issue #1184, opened 2016-02-25
Extend non-interactive completions for identifiers from TAGS


geraldus:
+ [x] make `hasktags` to work in non-interactive mode too
+ [x] combine TAGS completions with haskell-completions in non-interactive mode
+ [x] bring some light about this in docs
+ [x] interactive mode should override completion functions hook (as it does now)

Look like we can improve static completions to provide identifier and module name completions too making use of tags.

I need some time to do a basic research on this topic.

_Moved documentation notes to #702_

geraldus:
The best way I can think of is to steal this code from company mode
https://github.com/company-mode/company-mode/blob/250ca1c395c6f05b733abce69a2d928b6cd31618/company-etags.el#L62-L84

geraldus:
Ok, let's not complicate things!  I have a better solution.  Default value of `completion-at-point-functions` is `(tags-completion-at-point-function)`, so let's
+ in `haskell-mode` set `completion-at-point-functions` to 
```el
(haskell-completions-completion-at-point tags-completion-at-point-function)
```
+ in `haskell-interactive-mode` set `completion-at-point-functions` to
```el
(haskell-completions-sync-repl-completion-at-point)
```
This should give tags completions in static mode without even without any completion backend.  `company-mode` automatically re-uses `haskell-completions-completion-at-point` by default.  For interactive completions with company I would mention making `company-backends` list buffer-local for haskell buffers and prepending it with following grouped backend:
```el
(company-capf company-dabbrev-code)
```
This adds dabbrev completion candidates to REPL completions, which is quite handy.

gracjan:
Good idea.

geraldus:
> Good idea.

Ahh… Tags are generated only when session is present, so basically this won't work well: user must to generate tags manually and visit it manually.  Some features of Haskell Mode require minimal project information, and I often think about such functionality.  I think we should discuss this.

gracjan:
See #964.

geraldus:
@gracjan aha.  I though about some very simple `.cabal` file lookup in static mode (just for tags generation, enabled extensions and modules related to current file); though it looks very simple I guess the complexity is hidden underneath, for example cashing issues and etc.

geraldus:
I'm closing this

PR #1228, opened 2016-03-22
Extend static completions with candidates from tags table

geraldus:
Related #1184 

I thought to add `tags-completion-at-point-function` to `completion-at-point-functions` hook at first, but this approach is not good enough because in this case completion functions are run sequentially and candidates are not merged.  So, if haskell completion function will go first and it is able to complete thing at point there is no candidates from tags and vice versa.  And it is not clear what function should go first.

Fortunately, I've discovered `tags-completion-table` function, with it now it's possible to merge tags completions with keyword completions.

One question: should we revisit tags table automatically when it is regenerated (in `haskell-mode-generate-tags`)?

### DIFF
--- a/haskell-completions.el
+++ b/haskell-completions.el
@@ -296,7 +296,8 @@ PREFIX should be a list such one returned by
                ('haskell-completions-language-extension-prefix
                 haskell-ghc-supported-extensions)
                (otherwise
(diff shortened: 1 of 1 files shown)


gracjan:
Nice.

PR #1237, opened 2016-03-28
Font lock types in their own color

### LINKED ISSUE #899: Font-lock types differently
Types and constructors share namespace in Haskell and only context tell how a token starting with uppercase letter should be interpreted. Emacs' font lock isn't best at context but can be coerced to use one. The plan:

1. Trigger type font lock on `::` token.
2. Parse what comes after that as a type expression.
3. Mark where the type expression starts and ends.
4. Font lock tokens in marked region as types instead of constructors.

Tricky situations:

```haskell
putStrLn (123 :: Int)
x :: String <- return ""
x :: String = "xx"
x :: (HasField "field" r) => r
x + y :: Int where y = 45
data T = C
```

And everything related to `DataKinds`.

gracjan:
Fixes #899.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -820,7 +820,7 @@ Minor modes that work well with `haskell-mode':
           . haskell-syntactic-face-function)
          ;; Get help from font-lock-syntactic-keywords.
          (parse-sexp-lookup-properties . t)
-         (font-lock-extra-managed-props . (composition))))
+         (font-lock-extra-managed-props . (composition haskell-type))))
   ;; Haskell's layout rules mean that TABs have to be handled with extra care.
   ;; The safer option is to avoid TABs.  The second best is to make sure
   ;; TABs stops are 8 chars apart, as mandated by the Haskell Report.  --Stef
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -124,6 +124,12 @@ font faces assigned as if respective mode was enabled."
   :group 'haskell-appearance)
 
 ;;;###autoload
+(defface haskell-type-face
+  '((t :inherit font-lock-function-name-face))
+  "Face used to highlight Haskell types"
+  :group 'haskell-appearance)
+
+;;;###autoload
 (defface haskell-constructor-face
   '((t :inherit font-lock-type-face))
   "Face used to highlight Haskell constructors."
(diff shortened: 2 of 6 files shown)


gracjan:
So it currently looks like this:

font-lock-types

gracjan:
Todo:

- [ ] multiline cases do not refresh properly, possibly use `font-lock-extend-region-functions`
- [ ] handle `class` and `instance` with functional deps
- [ ] handle `data` ... `where` (looks easy)
- [ ] handle normal `data X1 = X2 Int String` (this is hard, has to know that X2 is constructor but Int and String are types )
- [ ] write lots of unit cases
- [ ] handle `module (...)` header (very hard)
- [ ] patterns? views?
- [ ] handle `type MyInt = Int` (hard, has to cross equal sign)
- [ ] handle `data family` and `type family`

gracjan:
In case of `data X = X1 Int | X2 String` everything is a type except uppercase constructors directly following equals or vertical bar.

geraldus:
Nice solution (:

gracjan:
So both patterns and views work only with constructors.

haskell-mode.el
```diff
@@ -501,6 +501,63 @@ executable found in PATH.")
...
+      (cond
+       ((null token)
+        (setq cont nil))
+       ((member token '(newline))
+        (setq last-token-was-newline (not last-token-was-operator))
+        (goto-char (match-end 0)))
+       ((and (member (match-string-no-properties 0)
+                     '("<-" "=" "<-" "←"  "," ";"
+                       ")" "]" "}" "|"
+                       "case" "class" "data" "default" "deriving" "do"
```
geraldus: can `haskell-font-lock--reverved-ids` be re-used here? (:

gracjan: Sorta maybe. I mean I'm to tired to make sure that all keywords end types. Is `forall` a keyword? is there other keyword that can happen in types?

gracjan: And, btw, thanks for reading this. It wasn't easy to write so I'm grateful for all the help.

gracjan: Fixed.

gracjan:
Mostly done. Lets give it a couple of days lying here and then we merge.

haskell-mode.el
geraldus: I always read everything, PRs, issues, and code (: But, unfortunately most of the time I have nothing to add ):

Issue #1206, opened 2016-03-14
Invent a way to test external programs integration


gracjan:
`haskell-mode` invokes many external programs. There is a need to have integration tests for such interactions so that we can ensure smooth cooperation. Integration tests is a category of tests that is completely absent from our test suite right now.

Requirements:

1. Integration tests help with development. It should be easier to write an integration test than to do the testing manually.
2. Integration tests should ensure working with multiple versions of external program.
3. Integration tests should run separately from unit tests so that unit testing development does not suffer in speed or issues.
4. Integration tests should run fast.
5. Integration tests should be deterministic and not suffer from environment misconfiguration.

Ideas:

1. Use el-mock and stop right before the external program is executed.
2. Create substitute programs as shell scripts, for example hasktags-mock.sh and setup haskell-mode to invoke those.
3. Install real programs and run those.

Discussion:

Item 1. cuts a bit too short although it might be useful in some scenarios. Item 2. is a great one because it does not depend on what happens to be installed in the `$PATH` and ensures that old versions of external executable will still work. Item 3. on the other hand is great to ensure that the newest executable will still work with `haskell-mode`.

I guess we will use each method depending on circumstances.

PR #1238, opened 2016-03-29
Run hs2hs on .hsc-files before loading

### LINKED ISSUE #1012: hsc2hs support
On doing `haskell-process-load-file` on a file that ends in .hsc, would it be possible to run `hsc2hs` on it first and load the resulting .hs (instead of the .hsc)?

unhammer:
Disable by setting haskell-process-do-hsc2hs to nil.

Tests only work interactively, TODO.

Fixes #1012.


haskell.el
```diff
@@ -398,18 +398,57 @@ Give optional NEXT-P parameter to override value of
...
+
+(defun haskell--file-name-to-load-string (file-name)
+  "Create a GHCi repl load statement from FILE-NAME."
+  (format "load \"%s\"" (replace-regexp-in-string
+                         "\""
+                         "\\\\\""
+                         file-name)))
+
+
+(defvar haskell-process-do-hsc2hs t
```
gracjan: We would not want to push the burden on users with this variable due to cost described in https://github.com/haskell/haskell-mode/wiki/Customizable-variables

tests/haskell-hsc2hs-tests.el
```diff
@@ -0,0 +1,65 @@
+;; haskell-hsc2hs-tests.el --- -*- lexical-binding: t; -*-
+
+(require 'ert)
+(require 'haskell)
+(require 'haskell-test-utils)
+
+
+(defvar default-hsc "{-# LANGUAGE CPP                      #-}
```
gracjan: Great test!

gracjan:
Note that on our build server there is no hsc installed and in general we would love to not depend on hsc. Lets do this:

1. Add a variable that is hsc-exec-path (with more proper name).
2. In tests customize this variable to a shell script that pretends to be a hsc compiler.

So basically we mock hsc on shell level. What do you think?

haskell.el
unhammer: > We would not want to push the burden on users with this variable due
> to cost described in
> https://github.com/haskell/haskell-mode/wiki/Customizable-variables

OK; PR updated.

unhammer:
> 1 Add a variable that is hsc-exec-path (with more proper name). 
> 2 In tests customize this variable to a shell script that pretends to
>   be a hsc compiler. 
>
> So basically we mock hsc on shell level. What do you think?

Sounds good to me. Do we customise test variables in haskell-hs2hs-tests.el, or is there a central file?

gracjan:
In `haskell-hs2hs-tests.el`. Actually you could use dynamic binding so that customization is done per test.

unhammer:
Added a fake script now; how does it look? I still get `Wrong type argument: stringp, ("ghci" "-ferror-spans")`  as in https://github.com/haskell/haskell-mode/issues/1012#issuecomment-202780797 both in my terminal and from travis though.

tests/haskell-hsc2hs-tests.el
```diff
@@ -0,0 +1,75 @@
...
+
+newtype NUMBERS = NUMBERS { unNUMBERS :: CInt }
+    deriving (Eq,Show)
+
+#{enum NUMBERS, NUMBERS
+ , rand_max = RAND_MAX
+ }
+")
+
+(defmacro with-hsc2hs (contents &rest body)
```
gracjan: This should *always* use the `hsc2hs.sh` script. I do not want reports from people that accidentally have something strange in their paths and therefore tests accidentally pass or fail.

tests/haskell-hsc2hs-tests.el
```diff
@@ -0,0 +1,75 @@
...
+        (should (looking-at-p "A_TYPO. undeclared"))))))
+
+(ert-deftest hsc2hs-compile-and-load ()
+  (with-hsc2hs default-hsc
+    (with-current-buffer "*haskell*" ; TODO: Where is this defined?
+      (goto-char (point-max))
+      (insert ":t unNUMBERS rand_max")
+      (goto-char (point-max))
+      (haskell-interactive-handle-expr)
+      (sit-for 1.0) ; TODO: can we wait until the prompt appears, with a timeout?
```
gracjan: There used to be an idea to have a callback from `haskell-interactive` when it detects that GHCi has loaded. As far as I know it is not implemented.

tests/hsc2hs.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+# Very stupid fake hsc2hs specific to our tests
+
+awk -v hs="${1%c}" '
```
gracjan: Nice! Is there a way to have this as pure awk script, i.e. not going through `/bin/sh`?

haskell.el
```diff
@@ -398,18 +398,58 @@ Give optional NEXT-P parameter to override value of
...
 
+
+(defun haskell--file-name-to-load-string (file-name)
+  "Create a GHCi repl load statement from FILE-NAME."
+  (format "load \"%s\"" (replace-regexp-in-string
+                         "\""
+                         "\\\\\""
+                         file-name)))
+
+(defvar haskell-process-path-hsc2hs "hsc2hs"
```
gracjan: This should be `defcustom` with `string` type.

gracjan:
This `Wrong type argument: stringp, ("ghci" "-ferror-spans")` indicates a couple of things:

1. Something in there should compose cmd line invocation even given list of strings. See #975.
2. On Travis we do not want ghci so there is a need for another awk script `fake-ghci.sh` just like the one you did for `hsc2hs`.
3. In general when executables aren't found the information should be much more user friendly.

gracjan:
And also note that this whole external program execution testing is tracked by issue #1206.

tests/haskell-hsc2hs-tests.el
unhammer: done

gracjan:
Nice!

gracjan:
How about inlining the contents of `hsc2hs.sh` in a function that will create this script on the fly, add executable bit and setup the path, I mean:

```
     (with-script-path ' haskell-process-path-hsc2hs
      "#!/bin/awk
      /END/ { the script here }"
      (body-of-function))
```

So this will create a new file with contents given, add executable bit, and set the given variable to the path of what was created. What do you think?

unhammer:
That'd work, but the macro can't clean up the file because it's running async now. Does
https://github.com/unhammer/haskell-mode/compare/hsc2hs-1012...unhammer:hsc2hs-1012-tempscript?expand=1 seem OK?

gracjan:
Yes.

No cleanup is okay, we might fix that later as low priority.

gracjan:
You need to create a script that plays the role of GHCi.

unhammer:
I haven't been able to get that to work; the process seems to just die before it sends any output at all. 

Does haskell-repl send any signals or such? Should the output be different from running ghci on the command line?

gracjan:
I do not think there are any signals involved. If there is an async communication isn't that the tests ends up before fake-ghc?

unhammer:
But why would the process die? When I run the test interactively with the real ghci, the process keeps running even when the test is done. In batch, I always get nil from get-buffer-process

gracjan:
Interesting. Communication is somehow preprocessed, maybe something is eating your output?

gracjan:
Ping? Anything happening here?

I find this idea really good, so would be lovely to see this pushed forward!

unhammer:
Hey; I never figured out how to fix this :( (also, I haven't actually had use for hsc2hs in some months months). 

I'd love to get it working though.  Could you explain what happen (or is supposed to happen) between the haskell-process-log and the REPL? I see haskell-mode at least puts the lambda character in the prompt, presumably some other things happen there as well. Since the log shows the expected output, I'm guessing whatever is parsing my fake-ghci is not seeing what it expects.

gracjan:
Well, I do know as much as you do at this point.

A `fake-ghci` would be great to test other things in `haskell-interactive-mode`.

gracjan:
Anyway I plan to steal the part with `awk`, make it more general and use it for some other tests.

unhammer:
Cool; it'd be great if something useful comes of this :)

gracjan:
I've remade the testing parts in #1331, no awk, just pure elisp :)

If you happen to need hsc again then please use my testing functionality.

Thanks for inspiration!

PR #1244, opened 2016-04-03
Move content from wiki to manual

gracjan:
Fixes #1244.


gracjan:
Some fixups need to wait for tomorrow or when time permits.

PR #1247, opened 2016-04-04
Disable interactive operator completions

geraldus:
GHCi prior to version 8.0.1 have bug in `:complete` command: when completing operators it returns a list of all imported identifiers (see Track [ticket](https://ghc.haskell.org/trac/ghc/ticket/10576)).  This leads to significant Emacs slowdown.  To aviod slowdown operator completions are disabled for now.


geraldus:
Whoops, it's too early.  I have to check not first char but the last.

geraldus:
Ok, done.

PR #1259, opened 2016-04-05
Add performance framework

### LINKED ISSUE #1181: Devise a performance framework
We would like to add more and more semantic features to haskell-mode and we risk performance degradation. We would not like to be surprised by such basic issues so we need to monitor performance of haskell-mode.

Tasks:

1. Find a performance metric in Emacs that is stable and deterministic.
2. Select a set of functions that are performance sensitive.
3. Attach performance tests that will make sure performance budgets aren't suddenly exceeded.

gracjan:
Fixes #1181.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,31 +14,37 @@ matrix:
   allow_failures:
     - env: TARGET=emacs-snapshot
 
-install:
-  - if [ "$TARGET" = "emacs24" ]; then
+install: |
+    case "$TARGET" in
+      emacs24)
         sudo apt-get update &&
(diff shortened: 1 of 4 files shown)


geraldus:
We are going to another rabbit hole :D

gracjan:
I would hate to accidentally cause quadratic complexity somewhere. Such performance tests are supposed to protect from this.

gracjan:
I thought that performance framework would be helpful in situations like #1360 but it does not seem so. I do not think the complexity required is worth it. Closing.

PR #1263, opened 2016-04-06
Expand font lock chapter

gracjan:
Promote font lock section to chapter, promote completions to chapter, add two nice animated gifs.


geraldus:
Cool!

PR #1266, opened 2016-04-06
Stylish-haskell runs on buffer instead of file on disk

WraithM:
So, I made stylish-haskell run on the current buffer. Now, when there's an error, stderr gets dumped to the buffer at the top of the file... This is probably not ideal, but it's the best I could manage with `call-process-region`.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -806,47 +806,33 @@ inferior GHCi process."
 Use buffer as input and replace the whole buffer with the
 output.  If CMD fails the buffer remains unchanged."
   (set-buffer-modified-p t)
-  (let* ((chomp (lambda (str)
-                  (while (string-match "\\`\n+\\|^\\s-+\\|\\s-+$\\|\n+\\'" str)
-                    (setq str (replace-match "" t t str)))
-                  str))
-         (_errout (lambda (fmt &rest args)
-		    (let* ((warning-fill-prefix "    "))
-		      (display-warning cmd (apply 'format fmt args) :warning))))
-         (filename (buffer-file-name (current-buffer)))
-         (cmd-prefix (replace-regexp-in-string " .*" "" cmd))
-         (tmp-file (make-temp-file cmd-prefix))
-         (err-file (make-temp-file cmd-prefix))
+  (let* ((cmd-prefix (replace-regexp-in-string " .*" "" cmd))
+         (tmp-buf (generate-new-buffer cmd-prefix))
          (default-directory (if (and (boundp 'haskell-session)
                                      haskell-session)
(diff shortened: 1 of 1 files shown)


WraithM:
Another way to possibly do this would be make a temporary file, save the current buffer to that temp file, then run stylish-haskell on that temp file with `call-process` (instead of call-process-region). You can get a separate stdout and stderr file with `call-process`, you can't get that with `call-process-region`.

WraithM:
Yeah, this solution is a hack. I don't think we should go with this solution.

gracjan:
Docs say:

```
(call-process-region START END PROGRAM &optional DELETE BUFFER DISPLAY
&rest ARGS)

Send text from START to END to a synchronous process running PROGRAM.
The remaining arguments are optional.
Delete the text if fourth arg DELETE is non-nil.

Insert output in BUFFER before point; t means current buffer; nil for
 BUFFER means discard it; 0 means discard and don't wait; and `(:file
 FILE)', where FILE is a file name string, means that it should be
 written to that file (if the file already exists it is overwritten).
BUFFER can also have the form (REAL-BUFFER STDERR-FILE); in that case,
REAL-BUFFER says what to do with standard output, as above,
while STDERR-FILE says what to do with standard error in the child.
STDERR-FILE may be nil (discard standard error output),
t (mix it with ordinary output), or a file name string.
```

Note that `BUFFER` may separate STDERR from STDOUT.

WraithM:
I made a new PR where it actually solved it. https://github.com/haskell/haskell-mode/pull/1268

haskell-commands.el
```diff
@@ -806,47 +806,33 @@ inferior GHCi process."
...
-                  str))
-         (_errout (lambda (fmt &rest args)
-		    (let* ((warning-fill-prefix "    "))
-		      (display-warning cmd (apply 'format fmt args) :warning))))
-         (filename (buffer-file-name (current-buffer)))
-         (cmd-prefix (replace-regexp-in-string " .*" "" cmd))
-         (tmp-file (make-temp-file cmd-prefix))
-         (err-file (make-temp-file cmd-prefix))
+  (let* ((cmd-prefix (replace-regexp-in-string " .*" "" cmd))
+         (tmp-buf (generate-new-buffer cmd-prefix))
```
gracjan: This temp buffer can be the same buffer as the one created with `with-temp-buffer` below, I think.

gracjan:
Does look pretty close to good solution to me...

gracjan:
I see. Do you know that in github you can just add commits to already open PR or even do a force push? No need to open new PRs :)

PR #1268, opened 2016-04-06
Stylish haskell after save

WraithM:
Okay! This actually works. It handles errors appropriately.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -806,47 +806,33 @@ inferior GHCi process."
 Use buffer as input and replace the whole buffer with the
 output.  If CMD fails the buffer remains unchanged."
   (set-buffer-modified-p t)
-  (let* ((chomp (lambda (str)
-                  (while (string-match "\\`\n+\\|^\\s-+\\|\\s-+$\\|\n+\\'" str)
-                    (setq str (replace-match "" t t str)))
-                  str))
-         (_errout (lambda (fmt &rest args)
-		    (let* ((warning-fill-prefix "    "))
-		      (display-warning cmd (apply 'format fmt args) :warning))))
-         (filename (buffer-file-name (current-buffer)))
-         (cmd-prefix (replace-regexp-in-string " .*" "" cmd))
-         (tmp-file (make-temp-file cmd-prefix))
-         (err-file (make-temp-file cmd-prefix))
+  (let* ((cmd-prefix (replace-regexp-in-string " .*" "" cmd))
+         (tmp-buf (generate-new-buffer cmd-prefix))
          (default-directory (if (and (boundp 'haskell-session)
                                      haskell-session)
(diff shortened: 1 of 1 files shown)


haskell-commands.el
```diff
@@ -806,31 +806,21 @@ inferior GHCi process."
...
+         (err-buf (generate-new-buffer " error"))
          (default-directory (if (and (boundp 'haskell-session)
                                      haskell-session)
                                 (haskell-session-cabal-dir haskell-session)
                               default-directory))
-         (_errcode (with-temp-file tmp-file
-		     (call-process cmd filename
-				   (list (current-buffer) err-file) nil)))
+         (_errcode
+            (call-process-region (point-min) (point-max) cmd nil (list (buffer-name tmp-buf) (buffer-name err-buf)) nil))
```
gracjan: AFAIK can't have buffer for STDERR, has to be a file.

WraithM: This worked as expected. I tested it. I could do a few more tests or something to be absolutely certain.

haskell-commands.el
```diff
@@ -806,31 +806,21 @@ inferior GHCi process."
...
-                    (setq str (replace-match "" t t str)))
-                  str))
-         (_errout (lambda (fmt &rest args)
-		    (let* ((warning-fill-prefix "    "))
-		      (display-warning cmd (apply 'format fmt args) :warning))))
-         (filename (buffer-file-name (current-buffer)))
-         (cmd-prefix (replace-regexp-in-string " .*" "" cmd))
-         (tmp-file (make-temp-file cmd-prefix))
-         (err-file (make-temp-file cmd-prefix))
+  (let* ((tmp-buf (generate-new-buffer " output"))
```
gracjan: Note that this has to be protected by `unwind-protect`.

haskell-commands.el
gracjan: Interesting. It should be more like:

```elisp
(call-process-region (point-min) (point-max) cmd nil
          (list tmp-buf (make-temp-name "tempxx")) nil))
```

gracjan: And again, temp file has to be deleted using `unwind-protect`.

gracjan:
Can you rebase on top of current master?

WraithM:
wasamasa on freenode #emacs helped me out a bit: https://gist.github.com/wasamasa/b89a54f56f3ef86d20cd6b8d85e2a5d8/961428e28279cb38955640211c6cbe3b3cc0fa6b

gracjan:
Need to use `unwind-protect`. Need to use `insert-buffer-substring`.

gracjan:
@WraithM: This is a lovely contribution, I would love to get this in. Can you rebease on mater and add `unwind-protect` where it is needed?

Issue #952, opened 2015-10-21
Unify haskell-indentation defcustoms into one variable


gracjan:
All of those should be replaced by a single defcustom:

    '(haskell-indentation-ifte-offset 4)
    '(haskell-indentation-layout-offset 4)
    '(haskell-indentation-left-offset 4)
    '(haskell-indentation-starter-offset 4)
    '(haskell-indentation-where-post-offset 4)
    '(haskell-indentation-where-pre-offset 4)

That defcustom should inherit value from `c-basic-offset`.

emmanueltouzery:
I think there are people who have a `do` offset at 4 and a `where` offset at 2. I would not merge those values.

Shimuuar:
Indeed. For example Johan Tibell: https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md

gracjan:
Tibell style has only where as exception. This is way less than 6 special
undocumented cases that we have now.

W dniu czwartek, 29 października 2015 Aleksey Khudyakov <
notifications@github.com> napisał(a):

bgamari:
It's actually unclear to how one is supposed to replicate @tibbe's style with `haskell-indentation` currently. I currently use,
```lisp
 '(haskell-indentation-ifte-offset 4)
 '(haskell-indentation-layout-offset 4)
 '(haskell-indentation-left-offset 4)
```
as recommended in [`haskell-style.el`](https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.el) and yet bindings in `where` clauses are indented two spaces too far,
```hs
hello hi = do
    doThings aBinding
  where
      aBinding = hi
```
What am I missing here?

gracjan:
@bgamari: looks like a bug to me.

bgamari:
Note that the indentation issue that I mention above came up again in #1042.

PR #1269, opened 2016-04-07
Simplify haskell-indentation-phrase-rest

### LINKED ISSUE #1084: No indentation cycling for case alternatives with haskell-indentation
Say I have:
```haskell
foo = case undefined ofX
```
where `X` is the cursor position. Pressing Enter indents new line by two spaces:
```haskell
foo = case undefined of
  X
```
That's not bad, but I think most haskell-mode's users over the years got used to behaviour of haskell-indent, which indented case alternatives relative to a `case` expression:
```haskell
foo = case undefined of
        X
```
While the default behaviour might be disputable I would certainly expect that pressing Tab will cycle me between these two alternatives. Currently it doesn't, which I think should be fixed.

### LINKED ISSUE #1197: Refactor `haskell-indentation-phrase-rest`
`haskell-indentation-phrase-rest` has special cases for `if-then-else`, `let-in` and something for `->`. This is not common code for these cases so split into specific functions and refactor.

### LINKED ISSUE #1270: Get rid of `haskell-indentation-indent-leftmost`
Variable `haskell-indentation-indent-leftmost`can be `t` or `nil` or `both`, and it applies in this case:

```haskell
func = case a of
  ^      ^
```

When `t` indentation offers just the left position, when `nil` offers just the rightmost, when `both` offers both.

There are some general truths about this situation:

1. **Truth #_1**: Nobody changes the defaults.
2. **Truth #_2**: People are upset when their favorite indentation point is not on the list.
3. **Truth #_3**: People are kind of okay with clicking <kbd>TAB</kbd> until it falls in place as they want.
4. **Truth #_4**: Cohort of people wanting rightmost indentation is progressively falling of the right edge of the screen.
5. **Truth #_5**: Cohort of people that care about indentation is growing smaller. People want it to just work.
6. **Truth #_6**: Determinism (as in: single option) is a virtue that could help with #1265.

Current default is `t` and I do not think there is a single person on this planet that changed it. There were some that wanted rightmost so this set is not empty.

Note: the fundamental reason to remove this defcustom is that it changes the algorithm. We do not test non `t` versions, if we keep this defcustom we have to start testing it. This is contrary to other numerical settings that just change the offsets.

gracjan:
Fixes #1197, fixes #1270, fixes #1084.

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -64,11 +64,6 @@
   :type 'integer
   :group 'haskell-indentation)
 
-(defcustom  haskell-indentation-ifte-offset 2
-  "Extra indentation after the keywords \"if\", \"then\", or \"else\"."
-  :type 'integer
-  :group 'haskell-indentation)
-
 (defcustom haskell-indentation-where-pre-offset 2
   "Extra indentation before the keyword \"where\"."
   :type 'integer
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -520,7 +520,6 @@ fixes up only indentation."
     ("type"    . haskell-indentation-data)
     ("newtype" . haskell-indentation-data)
     ("if"      . haskell-indentation-if)
-    ("case"    . haskell-indentation-case)
     ("let"     .
      ,(apply-partially 'haskell-indentation-phrase
                        '(haskell-indentation-declaration-layout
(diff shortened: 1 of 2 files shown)


geraldus:
Nice

gracjan:
Hmm, we must have `both` in `haskell-indentation-indent-leftmost` as per #1270 for this to work.

gracjan:
And we managed to make `haskell-indentation-ifte-offset` unused, nice. Related #952.

gracjan:
Some regularity gets in, some complexity gets out of indentation. Nice.

gracjan:
Sadly some determinism gets out of the window... sad.

PR #1277, opened 2016-04-09
Officially support only Emacs 24.3 and later

### LINKED ISSUE #1218: Require at least Emacs 24.3 or properly support Emacs 24.1 and Emacs 24.2
Currently `haskell-mode` claims compatibility with Emacs 24.1, this is not true.

Note that our test suite runs properly on both Emacs 24.1 and Emacs 24.2.

Support for Emacs 24.1/2 faces compilation warnings/errors:
- `eval-after-load` signals `"Error: Attempt to modify read-only object"`
- there are issues with `cl` package: function 'cl-find-if' from cl package called at runtime
- `setq-local` is absent

Fixable, but requires work.

So we have to decide and commit to one of the following:
1. Officially support Emacs 24.3 and greater.
2. Officially support Emacs 24.1/2 and add Travis tests for those.

Release dates for Emacs versions:
- Emacs 24.1: June 10, 2012
- Emacs 24.2: August 27, 2012
- Emacs 24.3: March 11, 2013
- Emacs 24.4: October 20, 2014
- Emacs 24.5: April 10, 2015

Ubuntu LTS 14.04 has Emacs 24.3 installed.

fice-t:
Closes #1218 

Perhaps this should be merged after a new stable release has been created. Then in the readme there can be a link to the stable release saying "This is the last version that sort-of supports 24.1/24.2".


gracjan:
Well, current master does not work with 24.2 so it would be a problem to claim it does.

fice-t:
I suppose so. I was thinking that there should be a recent stable release that does support 24.1/24.2. Perhaps you could edit an older tag to include that info?

Actually, it turns out that haskell-mode v13.18 also has the error I mentioned [here](https://github.com/haskell/haskell-mode/issues/1218#issuecomment-205082337) in 24.1.

Turns out that it's because `locate-dominating-file` doesn't accept a predicate as its 2nd argument before 24.3. This also means that 13.18 doesn't fully support 24.1, 24.2, or even 23. 
13.16 seems to fine.

gracjan:
We can include this information in the manual. We have one paragraph about Emacs 23, we can have more.

gracjan:
So, can you:

- [ ] add compatibility info in manual, there is a paragraph about this so next to it.
- [ ] change all instances of `(defvar x ...) (make-variable-buffer-local 'x)` into `defvar-local`
- [ ] change all instances of `(set (make-local-variable ..) ..)` into `setq-local`

?

This would clean up most glaring syntax problems.

fice-t:
What should it say? That 13.16 is the last version to support <24.3 or 13.18? As it stands 13.18 doesn't.

gracjan:
13.16

fice-t:
Now the manual just points to 13.16 as the latest version for 23/24.1/24.2 support.

The setq-local/defvar-local stuff was done with a query-replace-regex, but it should be fine.

fice-t:
@gracjan The [release descriptions](https://github.com/haskell/haskell-mode/releases)  for 13.16 and 13.18 should probably get updated to match the manual.

fice-t:
@gracjan Ah, also, are there any plans for the NEWS file? It hasn't been updated for a while.

gracjan:
I don't have energy for that. If you want to fix it then there is 15 months of Haskell Mode reports as a source material.

PR #1299, opened 2016-04-26
Add Unicode input method

u-combinator:
Add a function to auto insert module template with `Unicode input method` for the newly buffer.


u-combinator:
Add this function for someone like use `Unicode input method` when write code.

gracjan:
@u-combinator, thanks for the PR. Sadly, this is super specialized and we aren't in position to maintain such specialized functionality.

Have you considered using YASnippets? AFAIK it has functionality to add snippets like that by just a few keystrokes.

PR #1302, opened 2016-04-29
Implement electric characters

### LINKED ISSUE #1133: Implement electric comma, semicolon, etc

### LINKED ISSUE #1294: Unused lexical variables (Lexical Binding)
*Is this support for Emacs 24.5 version?*   
*Has small problem with some varable.*  

```
Compiling file /haskell-mode-20160419.752/haskell-indentation.el at Thu Apr 21 20:58:35 2016
haskell-indentation.el:657:1:Warning: Unused lexical variable `starter-indent-inside'
haskell-indentation.el:1024:1:Warning: Unused lexical variable `starter-line'

Compiling file /haskell-mode-20160419.752/haskell-load.el at Thu Apr 21 20:58:36 2016
haskell-load.el:93:1:Warning: Unused lexical argument `e'

Compiling file /haskell-mode-20160419.752/haskell-process.el at Thu Apr 21 20:58:36 2016
haskell-process.el:488:1:Warning: Unused lexical argument `e'
```

fice-t:
This does it like in `cc-mode'.

Also, I removed a few unnecessary things in the minor mode definition and removed an unused variable (Fixes #1294).

Closes #1133

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -74,10 +74,23 @@
   :type 'integer
   :group 'haskell-indentation)
 
-(defconst haskell-indentation-mode-map
+(defcustom haskell-indentation-electric-flag nil
+  "Non-nil means insertion of some characters may auto reindent the line.
+If the variable `electric-indent-mode' is non-nil then this variable is
+overridden."
+  :type 'symbol
+  :group 'haskell-indentation)
+(make-variable-buffer-local 'haskell-indentation-electric-flag)
+
+(defvar haskell-indentation-mode-map
   (let ((map (make-sparse-keymap)))
-    (define-key map (kbd "RET") 'haskell-indentation-newline-and-indent)
-    (define-key map (kbd "<backtab>") 'haskell-indentation-indent-backwards)
+    (define-key map (kbd "RET") #'haskell-indentation-newline-and-indent)
+    (define-key map (kbd "<backtab>") #'haskell-indentation-indent-backwards)
+    (define-key map (kbd ",") #'haskell-indentation-common-electric-command)
+    (define-key map (kbd ";") #'haskell-indentation-common-electric-command)
+    (define-key map (kbd ")") #'haskell-indentation-common-electric-command)
+    (define-key map (kbd "}") #'haskell-indentation-common-electric-command)
+    (define-key map (kbd "]") #'haskell-indentation-common-electric-command)
     map)
   "Keymap for `haskell-indentation-mode'.")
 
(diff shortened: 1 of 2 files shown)


haskell-indentation.el
```diff
@@ -323,7 +340,42 @@ fixes up only indentation."
...
+2) There is only one possible indentation position.
+3) The variable `haskell-use-electric-keys' is non-nil.
+4) The point is not in a comment, string, or quasiquote."
+  (interactive "*p")
+  (let* ((col (haskell-indentation-current-indentation))
+         (at-indent? (= col (current-column)))
+         ind)
+    (self-insert-command arg)
+    (when (and at-indent?
+               haskell-indentation-electric-flag
```
gracjan: Change this to:

```haskell
(or (equal 'always haskell-indentation-electric-flag)
    (bound-and-true-p 'electric-indent-mode))
```

Then the mode changing hook will not be necessary.

fice-t: Thanks, I don't know why I didn't think of that. At that point `haskell-indentation-electric-flag` might as well just be used as a boolean, so I changed it accordingly.

haskell-indentation.el
```diff
@@ -323,7 +331,31 @@ fixes up only indentation."
...
+  "Call `self-insert-command' to insert the character typed ARG times
+and indent when all of the following are true:
+
+1) The character is the first non-whitespace character on the line.
+2) There is only one possible indentation position.
+3) The variable `electric-indent-mode' or `haskell-indentation-electric-flag'
+   is non-nil.
+4) The point is not in a comment, string, or quasiquote."
+  (interactive "*p")
+  (let* ((col (haskell-indentation-current-indentation))
```
gracjan: Also: please check the `(or haskell-indentation-electric-flag electric-indent-mode)` in the beginning of the function. The point is that `haskell-indentation-current-indentation` may possibly be buggy, take lots of time or error-out. Therefore it should be invoked only after flag is enabled.

haskell-indentation.el
```diff
@@ -87,18 +100,13 @@
 It rebinds RET, DEL and BACKSPACE, so that indentations can be
 set and deleted as if they were real tabs."
   :keymap haskell-indentation-mode-map
-  (kill-local-variable 'indent-line-function)
```
gracjan: This `kill-local-variable`s should stay here. Reason: minor modes may be switched on and off, this code path is for switching off (although it is executed always).

fice-t: Right, I was thinking that someone would always want either haskell-indent or haskell-indentation. I guess someone might want neither.

gracjan:
Code looks good, I plan to run it locally for a couple of days to see how it 'feels'.

gracjan:
We should have a nice documentation page in manual. @fice-t, can you write it in `doc/haskell-mode.texi`? It is pretty easy.

fice-t:
Should that do for documentation?

PR #1305, opened 2016-04-30
[WIP] Add analytics

### LINKED ISSUE #1220: Consider event tracking
At this point we are in the dark about haskell-mode usage and about which parts of haskell-mode are used and in what ways. It is hard to make decisions about priority of optimizations, feature implementation or feature removal.

Insight by automatic event tracking could help a lot knowing what features are actually used and how.

Rough plan:

1. Use [Google analytics event tracking](https://developers.google.com/analytics/devguides/collection/analyticsjs/events) to count how many times each function is invoked.
2. Gather cumulative stats of usage while program is working and contact google only every couple minutes.
3. Send also information about other modes enabled along with haskell-mode.
4. Send also keybindings to know how people bind commands.
5. Never send any personal data.

There is prior art because [Aquamacs](https://github.com/davidswelt/aquamacs-emacs/blob/69699f60547c27c6b896c7f3ae3d26c818280462/aquamacs/src/site-lisp/check-for-updates.el#L288) sends general usage info to their servers while checking for updates.

Tasks:

1. Implement low level protocol to contact google analytics.
2. Implement function instrumentation and activate it only in our files.
3. See how make google analytics show useful graphs using given information.

gracjan:
Fixes #1220.

### DIFF
--- a/doc/haskell-mode.texi
+++ b/doc/haskell-mode.texi
@@ -74,6 +74,7 @@ interpreter (e.g. GHCi).
 * Spell checking strings and comments::  Using @code{flyspell-prog-mode}
 * Aligning code::                        Aligning code using @code{align-regexp}
 * Rectangular commands::                 Manage indentation manually
+* Telemetry support::                    Help out by doing nothing (almost)
 * Getting Help and Reporting Bugs::      How to improve Haskell Mode
 * Concept index::                        Index of Haskell Mode concepts
 * Function index::                       Index of commands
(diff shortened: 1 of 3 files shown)


david-christiansen:
I would personally vastly prefer that the analytics feature were a separate Lisp package distributed outside of `haskell-mode`, with `haskell-mode` limiting itself to a little nag on first run, perhaps. I really don't like the idea of having the code to watch me being present, even if I can turn it off in customize.

doc/haskell-mode.texi
```diff
@@ -2547,6 +2548,36 @@ This will insert the contents of the last killed rectangle.
...
 out what keys are bound beginning with the @kbd{C-x r} prefix.
 
+@node Telemetry support
+@chapter Telemetry support
+
+Most of Haskell Mode users have little time to help with the
+development. There is a way to help while doing almost nothing.
+
+Usage information could help us with time and effort allocation so that
+havilly used parts of Haskell Mode receive proper attention.
```
omefire: 'havilly' => 'heavily'

doc/haskell-mode.texi
```diff
@@ -2547,6 +2548,36 @@ This will insert the contents of the last killed rectangle.
...
+
+Haskell Mode optionally can sample usage statistics and report them back
+to central server. Samples track function call counts, variable access
```
omefire: 'central server' => 'a central server'

gracjan: Thanks.

gracjan:
Somebody would have to sit and understand these metrics. I do not think we have enough work-power to do this.

PR #1326, opened 2016-05-15
Improve tags behavior

bergey:
This PR fixes problems with tags that  I run into when using `haskell-mode` without an `interactive-haskell-mode` session.

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -1097,26 +1097,7 @@ recursively avoiding visiting unnecessary heavy directories like
 cabal-install, stack, etc and passes list of found files to Hasktags."
   (if (eq system-type 'windows-nt)
       (format "hasktags --output=\"%s\\TAGS\" -x -e \"%s\"" dir dir)
-    (format "cd %s && %s | %s"
-            dir
-            (concat "find . "
-                    "-type d \\( "
-                    "-path ./.git "
-                    "-o -path ./.svn "
-                    "-o -path ./_darcs "
-                    "-o -path ./.stack-work "
-                    "-o -path ./dist "
-                    "-o -path ./.cabal-sandbox "
-                    "\\) -prune "
-                    "-o -type f \\( "
-                    "-name '*.hs' "
-                    "-or -name '*.lhs' "
-                    "-or -name '*.hsc' "
-                    "\\) -not \\( "
-                    "-name '#*' "
-                    "-or -name '.*' "
-                    "\\) -print0")
-            "xargs -0 hasktags -e -x")))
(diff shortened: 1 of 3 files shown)


gracjan:
We had a PR recently that introduced exactly this `find` code. What was wrong with it?

gracjan:
Reference #964, #1204.

bergey:
Thanks for pointing out those related commits.  I think #1204 is necessary but not sufficient to using tags without a haskell session.

`haskell-mode-generate-tags` seems to assume that `haskell-mode-find-def` will return `nil` if there is no haskell-session.  The actual behavior is `(error "No session associated with this buffer. Try M-x haskell-session-change or report this as a bug.")`.  1919b7d26e819fcbc4d6e5285f08f4f3ba02916f only calls `haskell-mode-find-def` if the session exists.

I don't think the `find` code was added in #1204.  It was reformatted onto multiple lines, and moved into a different module.  The main problem is that when invoked this way, the tags table sometimes sends emacs to the wrong definition.  I mentioned this in [a comment](https://github.com/haskell/haskell-mode/issues/964#issuecomment-159612814) on #1204 - it's been true for a while.  I don't understand the TAGS file format, or exactly what goes wrong.

It seems the main point of #1204 was to call `hasktags` from emacs without going through ghci.  It also finds the .cabal directory (where the TAGS should go) without using ghci.  These are also improvements!

If you want some but not all three of these commits I can split them up.

gracjan:
The pruning of irrelevant directories seems to be important: https://github.com/haskell/haskell-mode/pull/1204#commitcomment-17386485, we need the `find` command in place.

Redefining <kbd>M-.</kbd> does not look like a good idea to me. There are other keys that are also in the same group: <kbd>C-x 4 .</kbd>, <kbd>C-x 5 .</kbd> and <kbd>M-C-.</kbd>. We do not want to redefine too much.

Note that I do not understand why we at all need functions like `haskell-mode-tag-find` or `haskell-mode-jump-to-tag` when pure Emacs functionality seems to be doing the trick just fine.

bergey:
Would you like me to make a new PR with just https://github.com/haskell/haskell-mode/pull/1326/commits/1919b7d26e819fcbc4d6e5285f08f4f3ba02916f?  Or shall I just close this entirely?  I don't have time to figure out the find situation, and I don't feel strongly about the binding.

gracjan:
Yes, lets go with 1919b7d first. I have slightly negative feelings above the two others, I hope we can find better ways.

bergey:
I expect we can find better ways.  Closed in favor of https://github.com/haskell/haskell-mode/pull/1330.

PR #1328, opened 2016-05-16
Re-introduce local (lexical) bindings to fix #1327.

### LINKED ISSUE #1327: haskell-doc messages obscured by "async" messages
With current haskell-mode the output of haskell-doc minor mode (as well as other output in the emacs echo area) is obscured by useless (at least to me) "async" messages when interactive-haskell-mode is started.  That is: the string "async" is printed to the echo area, thereby overwriting the actualy useful information ie. the output from doc mode...

To reproduce:
```
1. enable haskell-doc-mode
2. in any haskell file (with active haskell-mode) move the cursor on a known funtion (e.g. tail)
   → you should see the type of the function in the echo area (e.g. tail :: [a] -> [a])
3. now start interactive-haskell by C-c C-z
4. repeat step from 2.
   → this time the echo area shows "async" instead of the functions type
```

By bisecting I was able to track down the problem to commit 31426d720b2a40e607c6c95207b238af72158233 
reverting the changes made to haskell-doc.el there (that is: re-inserting the lexical-let) in the current master fixes the problem.  I currently don't see why, but maybe someone who understands the special magic taking place there can fix it without reintroducing the use of cl...

swilde:
Fixes #1327 by reintroducing the lexical bindings using emacs native lexical bind
(the file local lexical-binding is already set in the file header), which were removed by removing
the lexical-let in commit 31426d720b2a40e607c6c95207b238af72158233

### DIFF
--- a/haskell-doc.el
+++ b/haskell-doc.el
@@ -1535,14 +1535,18 @@ If SYNC is non-nil, make the call synchronously instead."
       (if sync
           (let ((response (haskell-process-queue-sync-request process ghci-command)))
             (funcall callback (funcall process-response response)))
-        (haskell-process-queue-command
-         process
-         (make-haskell-command
-          :go (lambda (_) (haskell-process-send-string process ghci-command))
-          :complete
-          (lambda (_ response)
-            (funcall callback (funcall process-response response))))))
-      'async)))
+        (let ((process process)
+              (callback callback)
+              (ghci-command ghci-command)
+              (process-response process-response))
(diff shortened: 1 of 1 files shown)


fice-t:
See [here](https://github.com/haskell/haskell-mode/issues/1327#issuecomment-219570989).

gracjan:
So as I understand this PR is not useful.

But thanks for digging into this issue!

PR #1333, opened 2016-05-20
Improve stylish buffer invocation

gracjan:


WraithM:
Thanks for writing the tests @gracjan!

Issue #624, opened 2015-05-02
beginning-of-defun-function and end-of-defun-function go too far


gracjan:
`beginning-of-defun-function` should put cursor in column 0 where the definition begins (currently it puts it at the end of line the line before).

`end-of-defun-function` should put cursor in column 0 after last line of the definition (currently it can go many empty lines below it).

gracjan:
And those two functions are active only when `haskell-decl-scan-mode` is active so referencing #623 seems in order. Those should work in plain `haskell-mode`.

gracjan:
Related: #1122.

gracjan:
`beginning-of-defun-function` should be same functionality as in `haskell-indentation-mode`, so it is part of common functionality.

gracjan:
Spec:

- `beginning-of-defun-function` should move up to the beginning of line that: 1) starts in the first column (for normal haskell) 2) starts in the 3rd column after bird (in literate) 3) is not in string 4) is not in comment 5) is not in quasiquote 6) is not in any parenthesized expression
- `end-of-defun-function` should move down to the beginning of line that: 1) is right after newline past last line with non-comment tokens 2) is not in string, comment, quasiquote or any parenthesized expression

gracjan:
`beginning-of-defun` is implemented in `haskell-indentation-mode` as `haskell-indentation-goto-least-indentation`. This should be put up higher level, presumably in `haskell-mode`.

gracjan:
And take those two functions out of `haskell-decl-scan`.

czipperz:
I believe this is already working?

czipperz:
On a semi-related issue, I made `haskell-ds-backward-decl` and `haskell-ds-forward-decl` work correctly with operators.  Is it OK to bind these to C-M-a and C-M-e, as they implement that functionality correctly, but the beginning-of-defun and end-of-defun functions don't work correctly with it.

gracjan:
The point is to make those functions available without enabling `haskell-decl-scan-mode`.

PR #1335, opened 2016-05-21
Going forward defun should go to end of defun

czipperz:
In other emacs modes, `C-M-e` will go just past the end of the current function rather than to the beginning of the next function.  Here is a patch that makes this work correctly for `haskell-decl-scan-mode`.

### DIFF
--- a/tests/haskell-decl-scan-tests.el
+++ b/tests/haskell-decl-scan-tests.el
@@ -1,0 +1,97 @@
+;;; haskell-decl-scan-tests.el -*- lexical-binding: t -*-
+
+;; Copyright © 2016 Chris Gregory. All rights reserved.
+
+;; This file is part of haskell-mode package.
+;; You can contact with authors using GitHub issue tracker:
+;; https://github.com/haskell/haskell-mode/issues
+
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+
+;; This file is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


gracjan:
Reference #624.

@czipperz: Can you add some tests for this behavior? Thanks.

czipperz:
I wrote some tests and fixed some bugs.  I refactored out this out to `haskell-ds-forward-decl` as I think it should only affect when `(end-of-defun)` is used.  The only problem with this new code is that if the cursor is on a line before the first non-blank line (ie file is "\nmodule File where\n..."), then it has the old behavior.

czipperz:
@gracjan Done

gracjan:
Instead of hardcoded position I would prefer checks based on `looking-at-p` after the move.

Can you also squash and force-push to this pr? Thanks.

czipperz:
That's smart will do tomorrow

czipperz:
I put you as the copyright holder for this code. Is this required for contribution or should I put myself down?

gracjan:
Copyright: Haskell mode Contributors
Author: You

czipperz:
Fixed build error if `(< (emacs-version) 24.4)`

tests/haskell-decl-scan-tests.el
```diff
@@ -0,0 +1,97 @@
+;;; haskell-decl-scan-tests.el -*- lexical-binding t -*-
```
gracjan: Add a colon between `lexical-binding` and `t`.

czipperz:
Did that.  What does that do?

czipperz:
Read it on Wikipedia nvm

PR #1336, opened 2016-05-21
Use PWD instead of CURDIR in Makefile

thumphries:
Produces correct autoloads when the source tree has been symlinked. If the source tree is a symlink, `CURDIR` will always be the dereferenced path, while `PWD` is usually what we expect. Using `CURDIR` can lead to some incorrect paths in `haskell-mode-autoloads.el`.

#### Context

I noticed this when trying to build a haskell-mode source tree that had been symlinked from another directory (I use GNU Stow to symlink things from a git repo):

```
[tim@axiom:~] % ls -alr ~/.emacs.d/haskell-mode
lrwxr-xr-x  1 tim  staff  40 21 May 23:31 /Users/tim/.emacs.d/haskell-mode -> ../src/emacs/emacs/.emacs.d/haskell-mode
```

During `cd ~/.emacs.d/haskell-mode && make`, the following are true:

```
CURDIR=/Users/tim/src/emacs/emacs/.emacs.d/haskell-mode
PWD=/Users/tim/.emacs.d/haskell-mode
```

... and this leads to broken relative paths in `haskell-mode-autoloads.el`:

```
Generating autoloads for ../../../../../.emacs.d/haskell-mode/highlight-uses-mode.el...
Generating autoloads for ../../../../../.emacs.d/haskell-mode/highlight-uses-mode.el...done
Generating autoloads for ../../../../../.emacs.d/haskell-mode/inf-haskell.el...
Generating autoloads for ../../../../../.emacs.d/haskell-mode/inf-haskell.el...done
Generating autoloads for ../../../../../.emacs.d/haskell-mode/w3m-haddock.el...
Generating autoloads for ../../../../../.emacs.d/haskell-mode/w3m-haddock.el...done
```

This fix should not break the build for the regular unsymlinked case. The Makefile shelled out to `pwd` before commit e4418c3 without incident.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ deploy-manual : doc/html
 $(AUTOLOADS): $(ELFILES)
 	$(BATCH) \
 		--eval '(setq make-backup-files nil)' \
-		--eval '(setq generated-autoload-file "$(CURDIR)/$@")' \
+		--eval "(setq generated-autoload-file \"$${PWD}/$@\")" \
 		-f batch-update-autoloads "."
 	# check if autoloads will really load
 	$(BATCH) -l "$@"


thumphries:
Hmm, `$PWD` might be a bashism. shelling out might be the way to go.

gracjan:
The `$(CURDIR)` part is interpreted by `make` (notice single dollar).

We might get away with relative path, that is a simple `$@` should do.

thumphries:
`$@` is a relative path, but it seems the code expects something absolute. emacs batch mode seems to drop into `/tmp` somewhere, so the relative path is meaningless:

```
usr/local/bin/emacs --eval "(add-to-list 'load-path (expand-file-name \"tests/compat\") 'append)" --eval "(when (boundp 'load-prefer-newer) (setq load-prefer-newer t))" --batch -Q -L . \
		--eval '(setq make-backup-files nil)' \
		--eval '(setq generated-autoload-file "haskell-mode-autoloads.el")' \
		-f batch-update-autoloads "."
Opening output file: no such file or directory, /private/tmp/emacs-W76dlA/emacs-24.4/lisp/haskell-mode-autoloads.el
make: *** [haskell-mode-autoloads.el] Error 255
```

On my box `$PWD` is defined and does the right thing for symlinks in csh, ksh, bash and zsh, not sure if it's a portability problem or not.

gracjan:
So the line in question should look something like this:

    --eval "(setq generated-autoload-file \"$${PWD}/$@\")"

Note changed quotes, double dollars and shell kind of variable dereference.

Can you test if something like this works for you? If so please force push new commit to this pull request. Thanks.

thumphries:
Still works as expected with your change. I've rebased and pushed.

Thanks for your patience, Gracjan, and thanks for maintaining haskell-mode.

gracjan:
Note: I've changed this place for Windows compatibility in #1373.

thumphries:
Thanks for the heads up. I've tested it locally and everything behaves correctly 🎉

gracjan:
Great! Thanks!

Issue #1313, opened 2016-05-03
Pattern synonyms are highlighted as definitions


taktoa:
It seems to me that every instance of the token `pattern` in the attached code sample should be highlighted with the keyword face (i.e.: the same face as `module`, `class`, `where`, etc.).

The main objection to this kind of fix is that people could want to name a function `pattern` when the extension is disabled. However, the `proc` keyword has similar problems but is currently being syntax-highlighted regardless of the presence of the relevant language pragma. I guess `proc` *is* technically in the Haskell 2010 Report so it's Official Haskell whereas -XPatternSynonyms is merely GHC Haskell, but I think that doesn't hold much water given the popularity of GHC extensions.

```haskell
{-# LANGUAGE GADTs           #-}
{-# LANGUAGE PatternSynonyms #-}
{-# LANGUAGE UnicodeSyntax   #-}
{-# LANGUAGE ViewPatterns    #-}

-- Pattern synonyms can be exported
module Main (pattern A1) where

data Foo = Foo
data Bar where Bar :: (Show a) => a -> Bar

-- Basic usage of unidirectional and bidirectional pattern synonyms
pattern A1 <- Foo
pattern A2 = Foo

-- Patterns can have type signatures
-- These type signatures have up to two contexts
-- (these are called CProv and CReq for proven and required constraints resp.)
pattern B1 :: (Show a) => (Show a) => a -> Bar
pattern B1 x <- Bar x
pattern B2 :: (Show a) => (Show a) => a -> Bar
pattern B2 x =  Bar x

-- If CProv is empty, it can be omitted.
pattern C1 :: (Show a) => a -> Bar
pattern C1 x <- Bar x
pattern C2 :: (Show a) => a -> Bar
pattern C2 x =  Bar x

-- If CProv is not empty but CReq is, it's denoted like so:
pattern D1 :: (Show a) => () => a -> Bar
pattern D1 x <- Bar x
pattern D2 :: (Show a) => () => a -> Bar
pattern D2 x =  Bar x

-- Explicitly-bidirectional pattern synonym
pattern Succ n <- (subtract 1 -> n) where Succ n = n + 1

-- With -XUnicodeSyntax, ← can be used for a unidirectional synonym definition
pattern Baz ← Foo

-- For comparison, a function definition:
foo :: Int -> Int
foo x = x + 5

main :: IO ()
main = return ()
```

gracjan:
[...]

PR #1337, opened 2016-05-22
Added font lock tests for pattern synonyms (for #1313)

taktoa:
I wrote tests for the font lock issues with the `pattern` keyword, as discussed in #1313.

Please check if the way I used `check-properties` is correct, since:
* It wasn't entirely clear but it seemed like the triplets in the second argument of `check-properties` each describe the fact that a token should be fontified a particular way in all the test strings given in the first argument (the alternative interpretation is that the triplets describe a series of fontifications for each token in order, and since none of the existing tests duplicate the first element of a triple, it's ambiguous which it is).
* I gave `t` as the second element of each triple, which I am not sure about either. This element is the "syntax" of the token, but I don't know enough about Emacs fontification to know what that means.

The `haskell-no-pattern-1` and `haskell-no-pattern-2` tests currently pass, so I commented out the `:expected-result :failed` so as to prevent an error.

cc: @gracjan 

EDIT: fun fact, this is pull request number 1337 on haskell-mode :D

### DIFF
--- a/tests/haskell-font-lock-tests.el
+++ b/tests/haskell-font-lock-tests.el
@@ -883,22 +883,20 @@
      "pattern A :: (C a) => () => a -> B"
      "pattern A :: (C a) => (C a) => a -> B"
      "pattern A :: (C a) => (C a) => a -> B")
-   '(("pattern"  t haskell-keyword-face)
-     ("where"    t haskell-keyword-face)
-     ("module"   t haskell-keyword-face)
-     ("A"        t haskell-constructor-face)
-     ("B"        t haskell-type-face)
-     ("C"        t haskell-type-face)
-     ("<-"       t haskell-operator-face)
-     ("←"        t haskell-operator-face)
-     ("=>"       t haskell-operator-face)
-     ("::"       t haskell-operator-face)
-     ("+"        t nil)
-     ("1"        t nil)
-     ("("        t nil)
-     (")"        t nil)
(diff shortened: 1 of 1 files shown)


tests/haskell-font-lock-tests.el
```diff
@@ -866,3 +866,55 @@
...
+     "pattern A n <- (subtract 1 -> n) where A n = n + 1"
+     "module Main (pattern A) where"
+     "pattern A :: a -> B"
+     "pattern A :: (C a) => a -> B"
+     "pattern A :: (C a) => a -> B"
+     "pattern A :: (C a) => () => a -> B"
+     "pattern A :: (C a) => () => a -> B"
+     "pattern A :: (C a) => (C a) => a -> B"
+     "pattern A :: (C a) => (C a) => a -> B")
+   '(("pattern"  t haskell-keyword-face)
```
gracjan: This checks only this first `pattern`. You need to write as many `patter` rules as there are `pattern` keywords in the example code. It is okay to leave out other tokens, `check-properties` skips over fragments that aren't listed.

gracjan:
Thanks.

gracjan:
Looks like I was wrong and when `PatternSynonyms` are enabled then `pattern` is a keyword everywhere it appears. For example this is a syntax error:
```
foo :: (a -> pattern) -> a -> pattern
```

Issue #1332, opened 2016-05-20
haskell-process-get-repl-completions completely hangs Emacs if REPL is inside IO action


cobalamin:
I noticed this by using `company-mode` with the `company-ghc` backend. It's an easy way to reproduce this problem.

I have a hook to enable `company-mode` inside `haskell-interactive-mode`. It works great usually, but recently I've noticed that my entire Emacs sometimes freezes up. After some investigation, I noticed that this happens only when I'm inside an IO action and either waiting (`company-mode` by default triggers after a certain amount of time) or triggering the completion manually. A simple `getLine` is enough, but a `forever $ putStrLn ...` loop also works.

I don't think `company-mode` nor `company-ghc` are to blame here, since they seem to simply use `haskell-process-get-repl-completions`. I killed my GHC process and had a look at the log:

```
-> getLine
<- ghc: signal: 15

-> :complete repl "\nasd"
<- 0 0 "\n"

<- ghc: signal: 15
ghc: signal: 15

<- Leaving GHCi.
Leaving GHCi.

Event: "finished
"

Process reset.
```

and noticed that this `:complete repl ...` call always occurs before REPL+Emacs start hanging. So I looked in my `.emacs.d` for files which send such a command, and `haskell-process.el` in `haskell-mode` showed up, specifically the `haskell-process-get-repl-completions`. And sure enough, when I insert a `(debug)` statement at the top of the function, I can see that this function is called when my problem would occur, and Emacs doesn't start hanging since it's not executing the rest.

This is possibly in some way related to https://github.com/haskell/haskell-mode/issues/871. However, I'm not in multiline mode. To make sure, I explicitly executed `:unset +m` before doing the same thing, and the problem still occurred.

Further info:
* `stack ghc -- --version`: `The Glorious Glasgow Haskell Compilation System, version 7.10.3`
* haskell-mode 20160518.945
* `M-x version`: `GNU Emacs 24.5.1 (x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21) of 2015-04-10 on builder10-9.porkrind.org`
* `M-x eval-expression haskell-process-type`: `stack-ghci`
* `stack --version`: `Version 1.0.4 x86_64`

geraldus:
[...]

PR #1341, opened 2016-05-27
prompt2 support and multi-line

### LINKED ISSUE #915: GHCi multiline prompt (`:set prompt2`) support
GHCi has multiline input mode which can enabled by `:set +m` or `:{` and `:}`.
For example:

```haskell
ghci> :set +m
ghci> let a = 12
ghci|     b = 42
ghci| 
ghci>
```

`:{` version:
```haskell
ghci> :{
ghci> let a = 12
ghci|     b = 42
ghci| :}
ghci>
```

Prompt for multiline input (here, `ghci| `) can be set by `:set prompt2 "ghci| "`, but we don't have any configuration item to set `prompt2` in `haskell-mode`.
This results in the following not-so-useful-for-indenting looking:

```haskell
λ> let a = 12
ghci|     b = 42
ghci| 
λ> :{
λ> let a = 12
ghci|     b = 42
ghci| :}
λ> 
```

So we need the way to `:set prompt2`.

### LINKED ISSUE #1278: Fix indentation in haskell-interactive-mode-multi-line
There is this magic: https://github.com/haskell/haskell-mode/blob/be9044b2e633b9342a68f762d07e82fe9ababf5f/haskell-interactive-mode.el#L254

This means that by using <kbd>C-j</kbd> one can have multiline command like this:

```
λ> let x = 1
       y = 3
```

it will be wrapped in `:{` and `}:` and send to GHCi like this:

```
:{
let x = 1
       y = 3
:}
```

Note that this is not documented and does not really work due to misaligned indentation.

Tasks:

- [ ] make it respect indentation
- [ ] document that by using <kbd>C-j</kbd> one can get multilne working

fice-t:
There's a couple oddities still left, but I think this should improve the state of multi-line support in the REPL.

* Adds a custom variable for prompt2 (multi-line prompt)
* Highlights prompt2 when inside a multi-line block or in `:set +m`

* haskell-interactive-mode-multi-line properly indents the multi-line expression (strips off whitespace in later lines)
* `C-j` indents a bit more like you'd expect

Fixes #915 
Fixes #1278

### DIFF
--- a/doc/haskell-mode.texi
+++ b/doc/haskell-mode.texi
@@ -1754,6 +1754,42 @@ To evaluate expressions, simply type one out and hit `RET`.
     123
 @end example
 
+@subsection Evaluating multiline expressions
+
+GHCi features two ways to evaluate multiline expressions. You can use
+@code{:set +m} to
+enable @uref{https://www.haskell.org/ghc/docs/latest/html/users_guide/ghci.html#multiline-input,
+multiline input} for all expressions, or you can wrap your expression in
+@code{:@{} and @code{:@}} (they have to be on their own lines).
+
+The prompt will change to indicate that you're inputting a multiline
+expression:
+
+@example
+λ> :@{
+λ| let a = 10
+λ|     b = 20
+λ|     c = 30
+λ| :@}
+@end example
+
+You can also simulate multiline mode by having your input contain
+newline characters. You can input a literal newline character with
+@kbd{C-q C-j}, or you can use:
+
+@example
+    M-x haskell-interactive-mode-newline-indent
+@end example
+
+which is bound to @kbd{C-j}. This command indents after the newline. You
+can simulate the above example like so:
+
+@example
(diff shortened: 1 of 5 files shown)


gracjan:
Looks nice. Will haskell-mode know if GHCi entered multiline mode? That could be useful to suspend for example company completions.

fice-t:
I couldn't find a direct way to do that, but one could perhaps check to see if the prompt2 was at the beginning of the line and use that information. Instead of just setting prompt2's font-lock property, one could also set an extra text property to make this easier.

That could go in a different PR I suppose.

fice-t:
@gracjan Though on second thought, I don't see much point in that. That doesn't help in cases like #1332 where it's not multi-line, but just in any IO action.

There's `(haskell-process-evaluating-p (haskell-process))` that detects both that case and the multi-line case. It seems to be fast enough.

PR #1347, opened 2016-06-01
Use :set +c in GHCi

### LINKED ISSUE #1200: Automatic ghci-ng/ghci-8.0 support for :set +c
Support for `:complete` command requires `:set +c` in the beginning of a ghci session. We should support this out of the box:

1. Issue `:set +c` unconditionally for every GHCi session started.
2. Look for response, if there was an error mark session as incapable of `:complete`.
3. Use this marker to either issue `:complete` commands of not.
4. Remove from documentation mentions of `:set +c` in `.ghci` file. Those are rude :)

fice-t:
Fixes #1200

### DIFF
--- a/doc/haskell-mode.texi
+++ b/doc/haskell-mode.texi
@@ -1226,10 +1226,9 @@ is bound to @code{haskell-goto-prev-error}
 is bound to @code{haskell-goto-first-error}
 @end table
 
-@section Using GHCi-ng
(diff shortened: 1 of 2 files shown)


gracjan:
What does this do to old versions of GHC?

fice-t:
Nothing as far as I can tell. I tested on 7.8.4 and there's just an extra message in the process log.

PR #1354, opened 2016-06-04
Remove a couple warnings and a byte-compile error

fice-t:


gracjan:
Good fix but what worries me more is why wasn't this caught by our travis build system? It is supposedly setup to detect such issues early. What is wrong with our builds?

PR #1356, opened 2016-06-05
Fix completion after operator

### LINKED ISSUE #1349: Completion and "."
In interactive-haskell-mode, when i try to complete something after a dot, it won't work. For example if i try to complete something like `Ssh.`, it won't show any completions. If i try to complete `Ssh`, it will show the expected candidates, e.g. `Ssh.noPasswords`. If it try to complete `Ssh.n`, it will also complete.

akermu:
Fix #1349

### DIFF
--- a/doc/haskell-mode.texi
+++ b/doc/haskell-mode.texi
@@ -445,6 +445,11 @@ example:
                          company-backends))))
 @end lisp
 
+If you use a GHCi version prior to 8.0.1 you might want to set
+@code{haskell-completions-complete-operators} to @code{nil}, if you
+experience major slowdown while trying to complete after an Haskell
+operator (See @url{https://ghc.haskell.org/trac/ghc/ticket/10576, GHC-Bug 10576}).
+
 @node Unicode support
 @chapter Unicode support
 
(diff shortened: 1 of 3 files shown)


gracjan:
Nice, thanks for putting this together. We must have some tests for this, at least 2x2=4 for nil/t for custom variable and two kinds of symbols (identifiers/operators).

akermu:
Ok, I'll try to add them in the next few days.

akermu:
I have a question: How should I design the tests? Would it be better to mock `haskell-completions-sync-repl-completion-at-point' or should I try to start a GHCi-session to test the functionality?

I'm in favor of the first option, as it has overall lower overhead.

gracjan:
Yes, first option. There is no need to run ghci for this test.

geraldus:
I believe first option is the only way to go (if I'm not mistaken we do not have GHC executable on Travis).

akermu:
I have added some tests, which finally work now on travis.

gracjan:
Great! Thanks!

geraldus:
Thanks

PR #1357, opened 2016-06-06
[WIP] Support internal prompts

### LINKED ISSUE #1353: Interactive-Haskell multiline support
By default, `Interactive-Haskell` change `prompt`, but not `prompt2`, so it looks strange, when try to use multiline input:

```haskell
λ> :set +m
λ> let
*Main FunnyPrint Main| 
λ> 
```

fice-t:
Some people do fancy stuff with the prompt (e.g. multiline, colouring).
This offers some support for it.

The haskell-interactive-mode-prompt-previous/next used the prompt regex
to search for the prompt, but this doesn't work with variable
prompts (i.e. containing module names). Now they use text property
search.

Closes #1353


fice-t:
@Pitometsu Can you see if this works for you? You also need to evaluate this:

```elisp
(setq haskell-interactive-use-interactive-prompt nil)
(setq haskell-interactive-prompt "FunnyPrint.prompt \"λ \" \"%s\" \" ¬\nλ > ")
```

haskell-commands.el
```diff
@@ -104,9 +104,9 @@ You can create new session using function `haskell-session-make'."
                                                     ":set -v1"
                                                     ":set +c") ; :type-at in GHC 8+
                                                   "\n"))
-          (haskell-process-send-string process ":set prompt \"\\4\"")
           (haskell-process-send-string process (format ":set prompt2 \"%s\""
```
gracjan: Please use `haskell-string-literal-encode` here, it can handle escapes the Haskell way.

haskell-commands.el
```diff
@@ -134,8 +134,20 @@ If I break, you can:
...
+  4. Hide these tips:   C-h v haskell-process-show-debug-tips")))
+                (unless haskell-interactive-use-interactive-prompt
+                  (with-current-buffer (haskell-session-interactive-buffer
+                                        (haskell-process-session process))
+                    (setq-local haskell-interactive-mode-prompt-start (point-max-marker)))
+                  ;; Now it's safe to set the prompt
+                  ;; Make sure to double escape any newlines
+                  (haskell-interactive-mode-run-expr
+                   (format ":set prompt \"%s\\4\""
+                           (replace-regexp-in-string "\n"
```
gracjan: Please use `haskell-string-literal-encode` here to handle escapes the Haskell way.

haskell-customize.el
```diff
@@ -324,6 +324,17 @@ The default is `haskell-interactive-prompt' with the last > replaced with |."
   :type 'string
   :group 'haskell-interactive)
 
+(defcustom haskell-interactive-use-interactive-prompt t
```
gracjan: I really dislike forcing users to know about some magical defcustoms. Is there a way to get the same functionality to Just Work without any fuss like setting something to some value?

gracjan:
We need to have automated tests for all of this functionality. People get offended very much if their colors stop working. There is some infrastructure for external interaction testing in https://github.com/haskell/haskell-mode/blob/master/tests/haskell-test-utils.el#L134 and with examples in https://github.com/haskell/haskell-mode/blob/master/tests/haskell-mode-tests.el#L577 .

haskell-customize.el
Pitometsu: @gracjan https://github.com/haskell/haskell-mode/issues/1353#issuecomment-224023785
https://github.com/haskell/haskell-mode/issues/1353#issuecomment-224010812

fice-t: @gracjan  @Pitometsu 
I've been trying to get another version that uses `haskell-interactive-prompt` only (checks if it's `nil`), but it's uglier. Also, `haskell-interactive-mode-multi-line` doesn't work without knowing the length of the last part of the prompt. (That is, mult-line with `set +m` and `:{ :}` works, but not with `C-j`).

If you don't mind both of these things, then I can replace this version.

Oh, I also don't see `haskell-doc`/`eldoc` working with it as well.

gracjan:
@fice-t, I understand next to nothing out of this, I just hope you can get this right :)

fice-t:
I guess that depends on what you call 'right'. I think this way of having an extra `defcustom`, while not ideal, is better than the alternative solution (that doesn't use an extra defcustom) I have on my local branch for a few reasons:

1) My other code uses [this](https://mail.haskell.org/pipermail/haskell-cafe/2007-September/032260.html) method to work around GHCi's lack of ability to redirect `:show` output. This code has to be formatted in a highly unreadable way so that GHCi parses it right.

2) As I said above, it would break a couple features.

@gracjan Would you object to a manual parsing of their GHCi init file on the Elisp side? I was hoping to avoid that method, but it might be better than the alternatives.

gracjan:
@fice-t, finding the `.ghci` might be a problem. Remember that GHCi is started from cabal or stack and it is hard to know what those are doing.

fice-t:
Even if it's started from cabal or stack doesn't it still use, by default, the same init file? So there should only be 4 places to check according to the [manual](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/ghci.html#ghci-dot-files). And if people need a different one we could just make it a defcustom.

Pitometsu:
With `stack` it really may be in other place. For example, I use it as
```yaml
# stack.yaml
ghc-options:
  "*": -ghci-script .ghci_conf
```
Because otherwise `stack` override some `.ghciconf` values.

fice-t:
@Pitometsu Doesn't `-ghci-script` append to the default GHCi init file rather than replacing it? This would mean that nothing special would have to be done on the Emacs side.

Does putting your `prompt` and `prompt2` info in your regular GHCi init file not work with `stack` with your other settings?

Pitometsu:
@fice-t well, it depend. `let` and `:def!` bindings from usual general `.ghciconf` overriden. Also if I call something like `:setupmyprompt` at the end of usual `.ghciconf` it overriden by `stack` as well. So I forced to call custom `.ghci_conf` explicit specially for `stack`.

Pitometsu:
But if you intend to set prompt interactively after `ghci` loaded, it doesn't matter.

Pitometsu:
@fice-t So in general `.ghciconf` user can create function via `:def` for prompt, but for stack it can be enable either interactively or at custom script.

Pitometsu:
@fice-t sorry, have no ability to check your PR for some time. And now could help with testing it. Please, point me, if I can check something.

Pitometsu:
Any updates here?

Pitometsu:
@fice-t any updates? :)

PR #1362, opened 2016-06-21
Make going forward defun work correctly with comments

czipperz:
I don't know how to write a test for this because it requires the buffer to be fontified, and I don't know how to do that in batch mode.

```
module Test where

{- documentation here -}
fun = 13

{- document this function -}
f = 23
```

This patch makes going forward defuns (`C-M-e`) correctly go over comments.
`M-< C-M-e` gets you to line 2, `C-M-e` again gets you to line 5, `C-M-e` again gets you to line 8.

Before it would take you to lines 4, 7, and 8, respectively.

### DIFF
--- a/haskell-decl-scan.el
+++ b/haskell-decl-scan.el
@@ -351,6 +351,43 @@ there."
   (interactive)
   (haskell-ds-move-to-decl nil (haskell-ds-bird-p) nil))
 
+(defun haskell-ds-comment-p
+    (&optional
+     pt)
+  "Test if the cursor is on whitespace or a comment.
+
+`PT' defaults to `(point)'"
+  ;; ensure result is `t' or `nil' instead of just truthy
+  (if (or
+       ;; is cursor on whitespace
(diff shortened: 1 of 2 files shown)


geraldus:
> it requires the buffer to be fontified, and I don't know how to do that in batch mode

If I'm not mistaken this should work for you:
```el
(ert-deftest haskell-mode-test ()
  (should (with-temp-buffer
            (haskell-mode) ;; enabling Haskell mode should fontify temporal buffer
            (eq t whatever))))
```

czipperz:
Tried that, doesnt work

czipperz:
Maybe I have to run `syntax-ppss-flush-cache` after adding the text... lol

gracjan:
`(font-lock-fontify-buffer)` is what you are looking for.

czipperz:
That works thanks!

czipperz:
Force pushed squashed some tests in

Issue #608, opened 2015-04-27
Add an interactive function to send the top-level definition the point is currently in to the REPL


ocharles:
Working in the Haskell REPL with only the ability to reload a file is extremely painful, because reloading a file throws away all local bindings. I propose that `haskell-mode` learns a new interactive function that sends just the current top-level definition to the REPL, shadowing the existing binding with that name.

As a motivating example, consider building a parser for a CSV file. First of all, I load the CSV contents into a string in the REPL with `csvStr <- readFile "data.csv"`. Next, I work on my `parse :: String -> [Record]` function. Whenever I make changes I send this function to the REPL, and then I can test `parse csvStr`. I don't have to reload the CSV file every time I change that function.

This becomes more important when the data I'm testing against comes from a database query. With the current behavior I have to reconnect to PostgreSQL, issue my query, and then I can test my changes. For all purposes, this makes the REPL significantly less productive than it could be.

We probably don't have the ability to send *all* top-level definitions to the REPL - for example, I'm not sure it's possible to redefine type class instances. However, just redefining functions would be a good start.

gracjan:
Good idea!

Note that it has to support also type signatures and multiple cases:


fact :: Int -> Int
fact 0 = ...
fact n = ...

ivan-m:
Not to mention that the type signature might be nowhere near the actual function definition, or there might not _be_ a type signature.

Also, if the `foo` function you reload is used by another function `bar`, I'm not sure if it's possible to have `bar` use the new definition.

Otherwise, it might be possible to do a region-based copy and then in ghci do a big `let` definition, possibly over multiple lines.

gracjan:
There are `beginning-of-defun-function` and `end-of-defun-function` that we
need to make work properly and:

1. Use `mark-defun` (that uses the above internally).
2. Send the region to eval.

Parts of this mechanism already exists, it is a matter of checking and
connecting the dots.

gracjan:
Note that `Ctrl-Alt-H` (`mark-defun`) works if `haskell-decl-scan-mode` is enabled. It should work always.

gracjan:
...and it should work better. It mostly marks what it should, but not always and not exactly.

guaraqe:
+1
This is one of the functionalities that can make working with Haskell closer to what working with Python or R is.

gracjan:
Related #1122.

PR #1366, opened 2016-06-22
allow pasting into multiline expressions

unhammer:
If user pastes something into a multiline expr, the lines won't have
the initial prompt2, so try to avoid removing them in that case

(I was looking at https://github.com/haskell/haskell-mode/issues/608 and wondering if I could make a quick hack by simply sending defun-at-point as a multiline expression, but it seems to require quite a lot of fiddling with indentation, seems like something that should be built-in to ghci rather …)

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -249,10 +249,10 @@ do the
 :}"
   (if (not (string-match-p "\n" expr))
       expr
-    (let ((len (length haskell-interactive-prompt))
+    (let ((pre (format "^%s" (regexp-quote haskell-interactive-prompt)))
           (lines (split-string expr "\n")))
       (cl-loop for elt on (cdr lines) do
-               (setcar elt (substring (car elt) len)))
+               (setcar elt (replace-regexp-in-string pre "" (car elt))))
       ;; Temporarily set prompt2 to be empty to avoid unwanted output
       (concat ":set prompt2 \"\"\n"
               ":{\n"


fice-t:
I disagree with this PR. Specifically, it breaks that function's purpose: strip extra multiline (not prompt2!) whitespace from the beginning of each non-prompt line.

For instance, after this was merged:

```haskell
λ> let a = 5
       b = 10
```

results in ` parse error on input ‘=’`.

@unhammer I'm not sure why you introduced the regex replace. For most cases I can think of (what that function was intended to handle), your solution is equivalent to simply removing that `cl-loop` from the function. That is, only the first line should start with a prompt.

If you wanted to paste multiline code without reindenting it, then perhaps a better option is to first enter a newline, and then paste your code. This approach should also look better as the indentation stays the same. 

Here's a solution:

```elisp
(defun haskell-interactive-mode-multi-line (expr)
  (if (not (string-match-p "\n" expr))
      expr
    (let ((len (length haskell-interactive-prompt))
          (lines (split-string expr "\n")))
      (if (string= (car lines) "")
          (setq lines (cdr lines))
        (cl-loop for elt on (cdr lines) do
                 (setcar elt (substring (car elt) len))))
      ;; Temporarily set prompt2 to be empty to avoid unwanted output
      (concat ":set prompt2 \"\"\n"
              ":{\n"
              (mapconcat #'identity lines "\n")
              "\n:}\n"
              (format ":set prompt2 \"%s\"" haskell-interactive-prompt2)))))
```

To test it, type `C-q C-j` to insert a literal newline, and then paste your multiline expression.

gracjan:
This looks like a top level bug that we should fix.

fice-t:
What do you mean by "top level"?

PR #1375, opened 2016-07-14
Improve guess module name

### LINKED ISSUE #1338: `haskell-guess-module-name` returns wrong results
When package name starts with a capital letter maybe, for example if package name is `JuicyPixels-extra` then in module `Codec.Picture.Extra` it returns: `JuicyPixels-extra.Codec.Picture.Extra`. This is incorrect, the result should be `Codec.Picture.Extra`. I can reproduce this with [`JuicyPixels-repa`](https://github.com/TomMD/JuicyPixels-repa) (just clone the repo, go to src file and call `haskell-guess-module-name`) which follows the same naming scheme.

gracjan:
Fixes #1338.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -1113,14 +1113,28 @@ successful, nil otherwise."
       (if (not (haskell-mode-try-insert-scc-at-point))
           (error "Could not insert or remove SCC"))))
 
-(defun haskell-guess-module-name ()
-  "Guess the current module name of the buffer."
-  (interactive)
-  (let ((components (cl-loop for part
-                             in (reverse (split-string (buffer-file-name) "/"))
+(defun haskell-guess-module-name-from-file-name (file-name)
+  "Guess the module name from FILE-NAME.
+
+Based on given FILE-NAME this function tries to find path
+components that look like module identifiers and composes full
+module path using this information. For example:
+
+    /Abc/Def/Xyz.lhs => Abc.Def.Xyz
+    /Ab-c/Def/Xyz.lhs => Def.Xyz
+    src/Abc/Def/Xyz.hs => Abc.Def.Xyz
+    c:\\src\\Abc\\Def\\Xyz.hs => Abc.Def.Xyz
+
+This function usually will be used with `buffer-file-name':
+
+    (haskell-guess-module-name-from-file-name (buffer-file-name))"
+
+  (let* ((file-name-sans-ext (file-name-sans-extension file-name))
+         (components (cl-loop for part
+                             in (reverse (split-string file-name-sans-ext "/"))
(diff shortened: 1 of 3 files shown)


mrBliss:
I understand that `haskell-guess-module-name-from-file-name`is a more accurate name than `haskell-guess-module-name`, but it is annoying for existing users of `haskell-guess-module-name` (like me), who have to deal with the breakage.

I'd like it better if gratuitous backwards-incompatible renaming were avoided or announced more clearly.

gracjan:
Hmm, it looked like a private function. How do you use it?

mrBliss:
I use it to show the module name instead of the file name in the mode line.

I have the following code in a `haskell-mode` hook:

```emacs
(when (buffer-file-name)
  (let ((module-name (haskell-guess-module-name)))
    (unless (string= "" module-name)
      (setq mode-line-buffer-identification
            (propertize module-name 'face 'mode-line-buffer-id)))))
```

How do you distinguish private from public functions in `haskell-mode`? To me this function looks public because:
* it has a docstring,
* it is marked interactive,
* it does not have `--` in its name (an often followed Elisp convention to mark function as internal/private).

gracjan:
That is a creative use of this function. Sad to hear that we broke your code. Sorry about that.

ivan-m:
My usage has also been broken:

```elisp
(require 'skeleton)
(require 'autoinsert)

;; Skeletons
(define-skeleton haskell-module-skeleton
  "Haskell hs file header"
  "Brief description (leave blank for default): "
  "{- \|\n"
  "   Module      : " (setq v1 (or (haskell-guess-module-name) "Main")) "\n"
  "   Description : " str | (concat "The \\\"" v1 "\\\" module") "\n"
  "   Copyright   : " (haskell-cabal-guess-setting "copyright") | (concat "(c) " user-full-name) "\n"
  "   License     : " (haskell-cabal-guess-setting "license") | "BSD-style (see the file LICENSE)" "\n"
  "   Maintainer  : " (haskell-cabal-guess-setting "maintainer") | user-mail-address "\n"
  "\n"
  "   " _ "\n"
  "\n"
  " -}\n"
  "module " v1 " where\n\n")

(add-to-list 'auto-insert-alist '("\\.hs\\'" . haskell-module-skeleton))
```

(And apparently `haskell-cabal-guess-setting` is deprecated, though I've never received a warning about it until I checked its documentation just now to make sure it wasn't a function I had defined.)

Is it possible to at least have the `haskell-guess-module-name` put back in as an alias for the rather often (even going by your own patch) `(haskell-guess-module-name-from-filename (buffer-file-name))`?  That's rather a mouthful :/

gracjan:
We can have obsolete alias. Who can do a PR?

PR #1377, opened 2016-07-16
Add an alias to fix the breakage of #1375

mrBliss:
In #1375, `haskell-guess-module-name` was renamed to
`haskell-guess-module-name-from-file-name` which caused breakage for
users of the original function name. Fix the breakage by defining an
alias that makes the original name point to the new function name.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -1137,6 +1137,11 @@ This function usually will be used with `buffer-file-name':
                              collect part)))
     (mapconcat 'identity (reverse components) ".")))
 
+(defun haskell-guess-module-name ()
+  "Guess the current module name of the buffer.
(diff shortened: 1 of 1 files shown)


ivan-m:
I think I would prefer a version that wraps up the call to `buffer-file-name`; I'm not sure if that's possible with `defalias` though.

mrBliss:
> I think I would prefer a version that wraps up the call to buffer-file-name; I'm not sure if that's possible with defalias though.

You're right, I forgot about that. I'll update the PR.

gracjan:
Thanks.

ivan-m:
This has different semantics if the module name can't be guessed: previously it would return `nil`, now it returns `""`.

PR #1380, opened 2016-07-19
Remove controversial keybindings

### LINKED ISSUE #856: Don't override M-n without permission/configuration
Making M-n something other than `next-error` is annoying and breaks everything else.
screenshot from 2015-09-08 15 53 28

gracjan:
Fixes #856.

Note that I did not introduce any new bindings as it is unclear what could those be.


purcell:
Thanks!

PR #1394, opened 2016-08-05
Implement italic and bold for haddock

gracjan:


geraldus:
Cool stuff!

gracjan:
There is plenty of other markup in haddock that could be implemented.

Issue #1409, opened 2016-08-19
Trailing garbage following expression: ’


aplanas:
Using emacs for git, and haskell-mode from git too:

    EMACS check-emacs-version
    Using EMACS = (correct emacs path), version = 25.1.50.1
    mkdir build-25.1.50.1
    EMACS build-25.1.50.1/haskell-menu.elc
    Trailing garbage following expression: ’
    make: *** [Makefile:69: build-25.1.50.1/haskell-menu.elc] Error 255

aplanas:
It is happening since aa1fc44925bc57aaa63246f449c11c7199ba9192

Maybe this single quote at the end? https://github.com/haskell/haskell-mode/commit/aa1fc44925bc57aaa63246f449c11c7199ba9192#diff-b67911656ef5d18c4ae36cb6741b7965R64

PR #1410, opened 2016-08-19
Makefile: Remove extra quote that breaks the build

aplanas:
Fix https://github.com/haskell/haskell-mode/issues/1409


gracjan:
Good catch. I was looking for this bug for a couple of weeks already.

PR #1412, opened 2016-08-26
Fix inferior-haskell warning font lock for GHC 8.0

alanz:
GHC 8.0 uses an error message of the form

```
/home/alanz/mysrc/github/alanz/ghc-exactprint/tests/Test.hs:99:3: warning: [-Wunused-matches]
    Defined but not used: ‘prettyRoundTripTests’
```

The word "warning" no longer has an initial capital letter.

### DIFF
--- a/inf-haskell.el
+++ b/inf-haskell.el
@@ -81,7 +81,7 @@ The command can include arguments."
   ;; The format of error messages used by Hugs.
   `(("^ERROR \"\\(.+?\\)\"\\(:\\| line \\)\\([0-9]+\\) - " 1 3)
     ;; Format of error messages used by GHCi.
-    ("^\\(.+?\\):\\([0-9]+\\):\\(\\([0-9]+\\):\\)?\\( \\|\n *\\)\\(Warning\\)?"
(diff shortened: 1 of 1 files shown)


inf-haskell.el
```diff
@@ -81,7 +81,7 @@ The command can include arguments."
   ;; The format of error messages used by Hugs.
   `(("^ERROR \"\\(.+?\\)\"\\(:\\| line \\)\\([0-9]+\\) - " 1 3)
     ;; Format of error messages used by GHCi.
-    ("^\\(.+?\\):\\([0-9]+\\):\\(\\([0-9]+\\):\\)?\\( \\|\n *\\)\\(Warning\\)?"
+    ("^\\(.+?\\):\\([0-9]+\\):\\(\\([0-9]+\\):\\)?\\( \\|\n *\\)\\(.arning\\)?"
```
gracjan: How about `[Ww]`?

alanz: That was my first idea, but there is so much escaping going on in the regex I was not sure if any would be needed.  Will update.

Issue #1438, opened 2016-10-14
Scroll to bottom option does not work


vshabanov:
I have `Haskell Interactive Mode Scroll To Bottom` option disabled but cursor still goes to the end of current output.

You could try code like this (and you could change `putStrLn` to `putStr`):
`forM_ [1..10] $ \ i -> threadDelay 1000000 >> putStrLn (show i)`
if you move cursor up or left it will jump to the end next second.

fice-t:
A couple functions were missing some `save-excursion`s. This should be fixed in #1441.

vshabanov:
Thank you. But now when this option is disabled it never scrolls to bottom, even if cursor was at the bottom.

Is it possible to restore old haskell-mode behavior -- when cursor automatically goes to bottom if it was at the bottom before new output has arrived and does not move when it's not at the bottom?

It was very convenient since I get cursor at the λ> prompt and see output progress by default but can navigate output while program is working if I want.

fice-t:
> Is it possible to restore old haskell-mode behavior -- when cursor automatically goes to bottom if it was at the bottom before new output has arrived and does not move when it's not at the bottom?

Was this really the old behaviour? Wasn't the absence of that behaviour the point of this issue?

Anyway, #1442 should do this for you.

vshabanov:
Wow. That's great! I even think that you can make `haskell-interactive-mode-scroll-to-bottom` set to `nil` by default again since current behavior is much more reasonable than always scroll to bottom no matter where cursor is.

Yes, it was the old behavior but I mean really old `inf-haskell` mode. (I've updated from haskell-mode v2.4 recently ;)

vshabanov:
Docs for `haskell-interactive-mode-scroll-to-bottom` says _Scroll to bottom in the REPL **always**_ so I thought -- OK, then with this option disabled it will scroll only when cursor is at the end of output.

But it scrolled in any case and then stopped to scroll in any case. And I just wanted old `inf-haskell` behavior. Here the confusion.

fice-t:
Ah, I never really used `inf-haskell`.

> I even think that you can make haskell-interactive-mode-scroll-to-bottom set to nil by default again since current behavior is much more reasonable than always scroll to bottom no matter where cursor is.

I agree that it's better. @gracjan would you agree to changing the default?

gracjan:
Can we get rid of the setting and just do the one single thing everybody wants?

fice-t:
[...]

PR #1442, opened 2016-10-21
Scroll in haskell-interactive-mode only when at correct positions

fice-t:
Adds to the fix for #1438.


gracjan:
One option less, great!

If somebody wants to have scroll-to-bottom then they can just put the cursor at the end of the buffer.

geraldus:
> If somebody wants to have scroll-to-bottom then they can just put the cursor at the end of the buffer.

Would it be great to mention this at top of the buffer?

gracjan:
This is default behavior in Visual Studio and probably other editors. We can mention it in documentation but we should spare screen real estate for program output.

PR #1443, opened 2016-10-24
Display "Collecting type info for N module(s) ..." message for GHC 8

vshabanov:


haskell-load.el
```diff
@@ -88,7 +88,14 @@ Restarts the SESSION's process if that is the case."
    ((haskell-process-consume process "\nBuilding \\(.+?\\)\\.\\.\\.")
     (let ((msg (format "Building: %s" (match-string 1 buffer))))
       (haskell-interactive-mode-echo (haskell-process-session process) msg)
-      (haskell-mode-message-line msg)))))
+      (haskell-mode-message-line msg)))
+   ((string-match "\\(Collecting type info for [[:digit:]]+ module(s) ...\\)"
```
gracjan: Please remove the `\\(` and `\\)` and use `string-match 0` instead of `string-match 1` later on.

gracjan:
Overall it is a shame that we do not have any automated tests for this kind of stuff. Next time a GHC 8.0.x or 9.x comes out it may change this string and functionality will silently fail.

geraldus:
This is second time we've beaten by changed GHC output.  You're right, we should have tests for all  GHC output parsing code pieces.

vshabanov:
I've made changes you've proposed.

PR #1444, opened 2016-10-26
Make most interactive text optionally writable

### LINKED ISSUE #1437: Do not make GHCi output read-only
In old haskell-mode I used to be able to clean GHCi output (completely, or just few lines). But when I'm trying to select whole buffer (or line) and remove it I get `Text is read-only`.

Sometimes there are really long lines of Haskell data (10-100k characters for example) in GHCi output. And they slowdowns Emacs a lot. Plus sometimes I'd like to see only latest "fixed" data and do not mess with previous output.

Is it possible to make GHCi output editable again?

fice-t:
Text that was explicitly documented as read-only is still
unconditionally read-only.

Fixes #1437.

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -343,6 +343,21 @@ printing compilation messages."
   :type 'boolean
   :group 'haskell-interactive)
 
+(defcustom haskell-interactive-mode-read-only
+  t
+  "Non-nil means most GHCi/haskell-interactive-mode output is read-only.
+This does not include the prompt.  Configure
+`haskell-interactive-prompt-read-only' to change the prompt's
+read-only property."
+  :type 'boolean
+  :group 'haskell-interactive)
+
+(defcustom haskell-interactive-prompt-read-only
+  haskell-interactive-mode-read-only
+  "Non-nil means the prompt (and prompt2) is read-only."
+  :type 'boolean
+  :group 'haskell-interactive)
+
 (defcustom haskell-import-mapping
   '()
   "Support a mapping from module to import lines.
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -298,7 +298,7 @@ SESSION, otherwise operate on the current buffer."
       (let ((prompt (propertize haskell-interactive-prompt
                                 'font-lock-face 'haskell-interactive-face-prompt
                                 'prompt t
-                                'read-only t
+                                'read-only haskell-interactive-prompt-read-only
                                 'rear-nonsticky t)))
         ;; At the time of writing, front-stickying the first char gives an error
         ;; Has unfortunate side-effect of being able to insert before the prompt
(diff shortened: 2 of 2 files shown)


gracjan:
We cannot burden our users with decisions about such minutae details. We should make this decision ourselves: should this buffer be read-only or should it be read-write?

fice-t:
> We cannot burden our users with decisions about such minutae details.

I don't think a couple variables are much of a burden.

The prompt being read-only should at least be a customization option. There's two precedents for it: `ielm-prompt-read-only` and `comint-prompt-read-only`.

>  should this buffer be read-only or should it be read-write?

Part of me likes it being read-only since it protects you from accidentally overwriting part of the output, forgetting about it, and then thinking that was the actual output.

Though that situation is rare enough, so either way is fine. I suppose it being read-write is more in-line with other Emacs REPLs.

gracjan:
How about we use `comint-prompt-read-only` directly?

fice-t:
Even `IELM`, which is `comint`-based, still has an option for it. Since `haskell-interactive-mode` isn't `comint`-based, I think there should be an option.

vshabanov:
I'm seeing such code in `comint-prompt-read-only` help
```elisp
   (add-hook 'comint-mode-hook
   	  (lambda ()
   	    (define-key comint-mode-map [remap kill-region] 'comint-kill-region)
   	    (define-key comint-mode-map [remap kill-whole-line]
   	      'comint-kill-whole-line)))
```

It may be a good combination to make prompt read only (do not editable by backspace/delete) but still be able to remove it by `C-k` or `C-w` or delete/backspace on selected region. As I understand comint makes this possible. Although I personally prefer the simplest solution -- just make all interaction buffer read-write.

geraldus:
> just make whole interaction buffer read-write.

And what will happen if you accidentally erase `𝝺` prompt?

gracjan:
I've merged as is. We will never know if anybody has found these two options and changed their value.

vshabanov:
Nothing will happen when `𝝺>` erased. I sometimes press enter to make it visible again and sometimes I just enter new command. It's possible to remap common editing and navigation operations to make buffer editable but preserving `𝝺>` prompts. But perhaps it's too much.

Current behaviour is more or less OK. We get read-only output by default but can change it via options.

fice-t:
Oh, I was going to slim it down to the one option, but if you do `(setq haskell-interactive-mode-read-only nil)` before loading `haskell-customize`, then it amounts to the same thing.

> It's possible to remap common editing and navigation operations to make buffer editable but preserving 𝝺> prompts. But perhaps it's too much.

Well, I'm not sure if it's really worth it to do it in this package. If you'd like, you can leave the prompt read-only and just put this in your init file:

```elisp
   (add-hook 'haskell-interactive-mode
      (lambda ()
        (require 'comint)
        (define-key haskell-interactive-mode-map [remap kill-region] 'comint-kill-region)
        (define-key haskell-interactive-mode-map [remap kill-whole-line]
          'comint-kill-whole-line)))

PR #1447, opened 2016-11-04
Recognize documentation comments as comments

czipperz:
This will fix edge cases in haskell-ds-forward-decl that would cause it to go forward to the beginning of the next decl when there are documentation comments, rather than the correct behavior of going to the end of the current declaration.

Behavior demonstrated (cursor is at POINTBEFORE at the beginning and should go to NEWPOINTAFTER, used to go to POINTAFTER).

    POINTBEFOREmain = x
    NEWPOINTAFTER
    -- | Documentation
    POINTAFTERx = return ()

### DIFF
--- a/haskell-decl-scan.el
+++ b/haskell-decl-scan.el
@@ -372,8 +372,9 @@ there."
            (setf fontfaces (list fontfaces)))
          (delq nil (mapcar
                     #'(lambda (f)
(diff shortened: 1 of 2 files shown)


czipperz:
I added tests for this behavior to the ones I wrote previously as well.

PR #1449, opened 2016-11-05
Use functions `bobp' and `member' instead of manual implementations.

czipperz:
This simplifies the code, making it consistent and readable.

### DIFF
--- a/haskell-decl-scan.el
+++ b/haskell-decl-scan.el
@@ -360,10 +360,7 @@ there."
   ;; ensure result is `t' or `nil' instead of just truthy
   (if (or
        ;; is cursor on whitespace
-       (let ((f (following-char)))
-         (or (= f ?\t)
-             (= f ?\n)
-             (= f ? )))
+       (member (following-char) '(?\t ?\n ?\ ))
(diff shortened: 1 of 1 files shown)


czipperz:
@geraldus Your renaming is incorrect.  Look at the first edit:

    -       (let ((f (following-char)))
    -         (or (= f ?\t)		
    -             (= f ?\n)		
    -             (= f ? )))
    +       (member (following-char) '(?\t ?\n ?\ ))

This uses `member` instead of a manual implementation of it.

geraldus:
@czipperz oh, yep, sorry for that!

gracjan:
Nice to see that haskell-decl-scan is getting some love. It could use much more hugging.

czipperz:
@gracjan Yea I fixed it up this summer.  Are there any more problems I could work on?

gracjan:
@czipperz: plenty. You can take any of the issues reported and just try to fix it. If you are talking about haskell-decl-scan then know that this module does not support half of useful haskell syntax extensions available: type operators, view patterns, type families, gadts. There is plenty to do to bring it up to current haskell language.

czipperz:
@gracjan With regards to the type operators, I can make it work if I replace the regular expressions with a function to recognize the beginning of functions.  I'm not sure if this will break anything, but I'll try it out.

PR #1454, opened 2016-11-10
Beginning/end of buffer should count as whitespace.

czipperz:


gracjan:
Have you considered using `looking-at-p "[[:space:]]"` or `(equal (string-to-syntax "_") (syntax-after (point)))`?

czipperz:
I'll change it to the second one.  Thanks @gracjan

PR #1483, opened 2017-02-09
Fix error in haskell-session-interactive-buffer

wlemuel:
haskell-session-interactive-buffer: Format specifier doesn’t match argument type


gracjan:
Thanks!

Issue #1457, opened 2016-11-16
haskell-process-wrapper-function is marked as risky as a file-local variable


TikhonJelvis:
The `haskell-process-wrapper-function` variable is marked as risky, which gives a warning popup if you set it as a directory local variable—as suggested by its documentation:

> See Info Node `(emacs)Directory Variables' for a way to set this option on
a per-project basis.

Is this desired behavior? Emacs marks variables ending in `-function` as risky by default (which is probably sensible), but do we want to do something different for a variable we *expect* to be set locally?

bezirg:
This fix seems to be gone in master. The variable is unsafe again , thus the issue is open

bezirg:
Another way is to advise inside the docstring that people may add to their config `(put 'haskell-process-wrapper-function 'safe-local-variable #'functionp)` if they want to trust this variable

PR #1488, opened 2017-02-23
Make haskell-process-wrapper-function safe for .dir-locals.

TikhonJelvis:
In its documentation, we recommend configuring `haskell-process-wrapper-function` per-directory in a `.dir-locals.el` file, but doing this creates an unnecessary warning because it is not marked as safe. This change suppresses that warning as long as the local value is a function, streamlining the experience.

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -85,7 +85,8 @@ a per-project basis."
   :group 'haskell-interactive
   :type '(choice
           (function-item :tag "None" :value identity)
-          (function :tag "Custom function")))
+          (function :tag "Custom function"))
+  :safe 'functionp)
 
 (defcustom haskell-ask-also-kill-buffers
   t


TikhonJelvis:
This addresses #1457

bezirg:
This fix seems to be gone in master. The variable is unsafe again , thus the issue is open

purcell:
This is actually an unsafe variable, though: the "fix" was reverted in #1581. If it were safe, then you might end up running arbitrary commands as a result of trying out someone's github project on your machine.

If you have a value for `haskell-process-wrapper-function` that you personally consider safe, you can mark it as such:

```
(add-to-list 'safe-local-variable-values 
  '(haskell-process-wrapper-function . my-process-wrapper))
```
or using customisation.

purcell:
For any future readers, see https://github.com/haskell/haskell-mode/pull/1488#issuecomment-557848374

PR #1493, opened 2017-03-08
Fix up installation instructions

### LINKED ISSUE #1490: Infinite "Contacting host: stable.melpa.org:80"
I'm trying to install this mode. I added the following code (as explained in the README)

```
(require 'package)
(custom-set-variables
 ;; custom-set-variables was added by Custom.
 ;; If you edit it by hand, you could mess it up, so be careful.
 ;; Your init file should contain only one such instance.
 ;; If there is more than one, they won't work right.
 '(package-archives
   (quote
    (("gnu" . "http://elpa.gnu.org/packages/")
     ("melpa-stable" . "http://stable.melpa.org/packages/")))))
```

to the file `~/emacs`. Then I re-open Emacs and I type `M-x package-refresh-contents`, and the following appears in a buffer:

```
Contacting host: elpa.gnu.org:80
Contacting host: stable.melpa.org:80
```

and it seems that it never terminates. 

I'm not very familiar in general with Emacs but I would like to try it with this mode.

ivan-m:
Closes #1490

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Make sure you have this in your [init file](http://www.gnu.org/software/emacs/ma
    (quote
     (("gnu" . "http://elpa.gnu.org/packages/")
      ("melpa-stable" . "http://stable.melpa.org/packages/")))))
+(package-initialize)
 ```
 
 Then run emacs, and evaluate:


ivan-m:
I think it would also be preferable to use `setq` directly rather than having a snippet of `custom-set-variables`.

gracjan:
I do not agree with `setq` remark. When using `customize` you get a section in `.emacs` that sets vars like this. Users will be able to recognize the region and edit accordingly.

Issue #1187, opened 2016-03-04
Choosing a test target using Stack


fatlazycat:
Hi, probably my own lack of knowledge but I use haskell-session-change-target to set the target to be the test of my cabal file within a stack project.

When I start the repl it fails to find modules defined within the test target.

What obvious stupid thing am I doing wrong ? :)

Thanks

geraldus:
@fatlazycat how do you start your REPL by the way?

geraldus:
`haskell-session-change-target` also fails for me when I select executable.  What is the purpose of target selection, usually I'm either reloading file in REPL <kbd>C-c C-l</kbd> or running build command <kbd>C-c C-c</kbd>.

fatlazycat:
In a test haskell file, i do haskell-session-change-target then start the repl ( I'm using spacemacs ) SPC m s b to load the current file into the repl which i believe is fn haskell-process-load-file

fatlazycat:
Is the there a difference given I would start a stack repl with 
```
stack repl 'project name':test:'test target'
```
?

fatlazycat:
Ah, can get it to work if I type that pattern when i do haskell-session-change-target so can work around it.

Guess the targets are just incorrect as it's just showing the cabal versions ?

geraldus:
@gracjan can you help with this?  I have no much experience personally.  I guess target selecting makes some sense when running <kbd>C-c C-c</kbd> `haskell-process-cabal-build`.  I'm not sure.

gracjan:
I tried to understand what you are talking about here but no avail. I'm lost. What is the problem again?

fatlazycat:
Sorry, when showing the targets for ```haskell-session-change-target``` for a stack based project I see the cabal entries for executables & test suites. 

But when attempting to use any, such as the test target, it does not start CHCI correctly.

I assume it's running the command ```stack ghci test-suite```

But as far as I understand for a stack project you want to invoke it as 

```stack ghci project-name:test:test-suite```

gracjan:
Ah, now it is clear as sun.

You are right that command line invocation is the same for cabal and stack projects, as I see it started to diverge. We have to split this functionality into two: cabal and stack.

Should not be that hard. @fatlazycat, are you able to look at this?

fatlazycat:
Sure will try and take a look.

gracjan:
Great!

fatlazycat:
[...]

PR #1494, opened 2017-04-02
Generate 'stack' tool style compilation targets.

csasarak:
This is an initial attempt at addressing issue #1187. Please let me know if there's a better way, this is my first outing into the haskell-mode code.

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -468,9 +468,10 @@ OTHER-WINDOW use `find-file-other-window'."
 (defun haskell-cabal-section-data-start-column (section)
   (plist-get section :data-start-column))
 
-(defun haskell-cabal-enum-targets ()
-  "Enumerate .cabal targets."
-  (let ((cabal-file (haskell-cabal-find-file)))
+(defun haskell-cabal-enum-targets (&optional process-type)
+  "Enumerate .cabal targets. PROCESS-TYPE determines the format of the returned target."
+  (let ((cabal-file (haskell-cabal-find-file))
+        (process-type (if process-type process-type 'ghci)))
     (when (and cabal-file (file-readable-p cabal-file))
       (with-temp-buffer
         (insert-file-contents cabal-file)
@@ -488,7 +489,10 @@ OTHER-WINDOW use `find-file-other-window'."
                           (string= val "{")
                           (not val))
                       (push projectName matches)
-                    (push val matches))))
+                    (push (if (eq 'stack-ghci process-type)
+                              (concat projectName ":" val)
+                            val)
+                          matches))))
             (haskell-cabal-next-section))
           (reverse matches))))))
 
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -468,6 +468,13 @@ OTHER-WINDOW use `find-file-other-window'."
 (defun haskell-cabal-section-data-start-column (section)
   (plist-get section :data-start-column))
 
+(defun haskell-cabal-map-component-type (component-type)
+  "Map from cabal file COMPONENT-TYPE to build command component-type."
+  (let ((component-type (downcase component-type)))
+    (cond ((equal component-type "executable") "exe")
+          ((equal component-type "test-suite") "test")
+          ((equal component-type "benchmark")  "bench"))))
+
 (defun haskell-cabal-enum-targets (&optional process-type)
   "Enumerate .cabal targets. PROCESS-TYPE determines the format of the returned target."
   (let ((cabal-file (haskell-cabal-find-file))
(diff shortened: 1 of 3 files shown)


gracjan:
Looks good to me. If nobody raises any voices agaist I'll merge this in a day or two.

hvr:
What are "stack tool style" compilation targets exactly btw? Is this any different from cabal's compilation target syntax? For reference (as I couldn't find a place where this is documented), Cabal uses the following (fully-qualified) syntax to identify components within a project containing one or more packages:

    <packagename>:<component-type>:<component-name>

where `<component-type> := {exe,lib,bench,test}`

When it's unambiguous, it's allowed to use less qualified forms, e.g.

 - `mypackage:exe:myprog1`, `exe:myprog1` or `myprog1`
 - `mypackage`
 - `mypackage:lib` or `mypackage:lib:mypackage` (this denotes the primary library of a package - cabal 2.0 supports multiple libraries per package)

haskell-cabal.el
```diff
@@ -488,7 +489,10 @@ OTHER-WINDOW use `find-file-other-window'."
                           (string= val "{")
                           (not val))
                       (push projectName matches)
-                    (push val matches))))
+                    (push (if (eq 'stack-ghci process-type)
+                              (concat projectName ":" val)
```
hvr: what is a `projectName`? Does Stack name projects?

csasarak: `projectName` is whatever appears in the 'name:' field at the top of the cabal file, it's in one of the hidden sections in the diff, but `projectName` is defined on line 481 with `(haskell-cabal--get-field "name")`. Stack doesn't have a separate notion of a project to what cabal provides as far as I can tell.

csasarak:
I hadn't realized that cabal also uses that syntax since as you said it isn't documented anywhere. Stack uses a very similar syntax, but it's required: https://docs.haskellstack.org/en/stable/build_command/#target-syntax
The issue I was having that I would try to switch to a test-suite which had different dependencies than the library package. Using `haskell-session-change-target` and then building would fail because it would try to use a target like 'project-test' when stack requires 'project:project-test', or possibly ':project-test'.

gracjan:
Is cabal syntax same as stack syntax?

If so we should just be using fully qualified target specs for all cases. This is autocompletion after all, so it makes sense to list fully qualified target names.

@hvr: good catch, I did not know cabal has extended syntax. Since when is it supported?

csasarak:
@gracjan Comparing the description and the documentation in Stack's documentation, the one case where it seems to be different is that Stack allows something like `:project-test` as a target but it doesn't seem that cabal does per @hvr's comment.  If cabal is okay with that though, it would be simpler to just always do the extended syntax.

gracjan:
We should do fully qualified syntax always. @csasarak: can you fix that? Thanks.

csasarak:
@gracjan I will try to take care of that tonight.

haskell-cabal.el
hvr: oh I see, then clearly `projectName` is a misnomer (but your PR didn't introduce it... it was already in the codebase before that... :-) )

csasarak: I can switch it to `package-name` which might be more appropriate while I'm making edits tonight.

hvr:
@gracjan I think the 3-part syntax is only supported in `new-build`; whereas the 2-part `{lib,bench,test,exe}:<name>` syntax was already supported by older cabal versions; but I'm not sure how far back.

csasarak:
@hvr Is it okay then to assume that current cabal users will use `new-build`? So I can unconditionally generate the 3-value target? The 2-value syntax used by stack is `<package-name>:<name>` but the syntax you describe is `<component-type>:<name>`.

ivan-m:
@casarak: I'm not using `new-build` yet (I looked at it earlier this year but it didn't seem to have all the functionality yet, at least not documented well enough for me to work out how to use it)

csasarak:
@ivan-m Good to know, I'm playing with stack and cabal tonight (I'm most familiar with stack) so I can take a more educated stab at this.

csasarak:
So unfortunately the acceptable syntax for targets are different between stack and cabal. 
`stack build package:package-test` works but `cabal build package:package-test` does not while `cabal build test:package-test` works but `stack build test:package-test` does not.  Unfortunately the fully qualified names that stack accepts (like `stack build package:exe:package-exe`) aren't accepted by cabal unless you do `cabal new-build`. 

I have a partial solution that I want to test a bit more, but if we anticipate stack command syntax deviating from cabal more in the future (or vice versa) it would probably make more sense to forgo my solution and have some abstraction so that stack command and cabal command strings aren't being generated in the same place.

hvr:
@csasarak I think it's reasonable to assume that Stack won't coordinate their syntax w/ Cabal upstream. So anticipating syntax deviation and abstracting seems reasonable.

ivan-m:
Ebal has a very nice interface for this.

csasarak:
@ivan-m  Ebal does look promising.
@hvr Okay, it'll take me some time to get more acquainted with how things are done. Unless it makes more sense to write tests for what I have already and then refactor the commands as a separate task/issue?

gracjan:
For uninitiated: https://github.com/mrkkrp/ebal

gracjan:
In this case we should split code paths inside haskell-mode for ghci, cabal and stack properly. This is a more involved change, but needed very much. Would be great to have a more principled approach to the issue.

csasarak:
I've added some tests to my original implementation for the stack/cabal fully-qualified targets. I'd need some time to look at the code in order to make a decent decision about how to split things up and will be traveling for the next week and a half or so, so I won't make any progress during that time. Still, if you have any advice for where to start (`haskell-interactive-mode.el`?) I can look into it when I return.

gracjan:
We can merge it for now and refactor later. Is it ready to be merged in current state?

csasarak:
To the best of my knowledge it's ready to be merged.

PR #1495, opened 2017-04-03
Add haskell-interactive-mode-flush-output

emmanueldenloye:
I thought it would be nice to have something like comint-delete-output for the haskell-interactive-mode. I decided to not to deal with multiline commands, but if anyone thinks that would be nice to have as well, I'll look into it.


gracjan:
Looks cool, but this function is not connected to any affordability. There should be either a keybinding, or menu item or at least texinfo entry describing existence of this function.

emmanueldenloye:
Ok, I'll add that to the pull request shortly.

Issue #1501, opened 2017-04-22
[discussion] unification of process and session


vasanthaganeshk:
Since `haskell-process` and `haskell-session` both have separate processes, but have some redundancies, it is desirable to unify both of them. I propose we do it incrementally.

## phase 1
+ The idea is we are going to keep the same architecture.
+ copy all forms from `haskell-process.el` to `haskell-session` (no functions or defcustoms are renamed), but we change the appropriate `require` statements. At this point nothing changes, so no new tests are required.

## phase 2
+ We retain the same architecture, we make `haskell-process` related forms as low level detail and we expose only the `haskell-session` related forms.
+ We depricate `haskell-process` related forms (that means they are still available but they will be removed in the future, so the old defcustoms will still work, but they can also be configured with their `haskell-session` cousins; same for other forms).
+ Docs!
+ Since there is renaming, we have to modify appropriate tests, dependent modules as well.
+ **divide phase two into a number of steps?**

## phase 3
 + now we can start the actual work, we can start with some simple refactoring

The reason for this is so that it can be easy to review merge requests.

This is the same as what was discussed in #799 and #808

vasanthaganeshk:
This is happening in a separate repository

PR #1502, opened 2017-04-23
phase 1 of #1501

vasanthaganeshk:
copied all forms from haskell-process.el to haskell-session, changed the appropriate require statements, updated docs.

phase 1 of #1501


gracjan:
This will be done in separate repo.

PR #1506, opened 2017-05-14
Make haskell-cabal-subsection-arrange-lines sort base before other packages in build-depends

Philonous:
Changes sorting of dependencies so "base" appears at the type of the list.

E.g. instead of 

```
build-depends: aeson
             , base >= 4.8 && < 5
             , text
```

we have

```
build-depends: base >= 4.8 && < 5
             , aeson
             , text
```

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -728,7 +728,7 @@ Respect the comma style."
      (haskell-cabal-goto-mark)
      (haskell-cabal-remove-mark)))
 
-(defun haskell-cabal-sort-lines-predicate (key1 key2)
+(defun haskell-cabal-sort-lines-depends-compare (key1 key2)
   (let* ((key1str (buffer-substring (car key1) (cdr key1)))
          (key2str (buffer-substring (car key2) (cdr key2)))
          (base-regex "^[ \t]*base\\($\\|[^[:alnum:]-]\\)"))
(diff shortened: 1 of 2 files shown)


gracjan:
How will the following be sorted [base32-bytestring](http://hackage.haskell.org/package/base32-bytestring), [base64-string](http://hackage.haskell.org/package/base64-string), [base91](http://hackage.haskell.org/package/base91)?

ivan-m:
Not to mention [base-compat](http://hackage.haskell.org/package/base-compat).

Philonous:
Good point. I fixed the regex to properly distinguish between "base" and "base*", made sure it only applies to "build-depends" sections and added a test.

``` 
Executable bin-1
    Main-Is:          TestParsing.hs
    Build-Depends: aeson
                 , text
                 , base >= 4.8 && < 5
                 , base64
                 , bytestring
                 , base-compat
```

will be sorted to

```
Executable bin-1
    Main-Is:          TestParsing.hs
    Build-Depends: base >= 4.8 && < 5
                 , aeson
                 , base-compat
                 , base64
                 , bytestring
                 , text
```

I think this is the right behaviour because only the "base" package is special.

ivan-m:
What about packages with multiple build -depends?

Though if an executable depends on the internal library I think it should come first, and then base.

Philonous:
@ivan-m The sorting only ever acts on the sub-section point is in when you call haskell-cabal-subsection-arrange-lines (C-c C-s by default). The presence of other sections doesn't affect it. 

Putting the internal library at the top is a nice idea, but out of scope for this PR. I just wanted to fix the "base"-case because it is arguably always present. The sorting breaks down in a few other situations too, e.g. for comments and conditionals, so you'll have to sort lines manually for now in those cases.

gracjan:
Nice! And even a unit test is there! Beautiful!

gracjan:
@ivan-m: people at Cabal were working on full read-cabal-file/change/output workflow that would allow structured edits to *.cabal files. We should depend on that functionality if we want to support Cabal files refactorings.

Issue #1433, opened 2016-10-11
Support `new-repl`


slpopejoy:
I'm using cabal 1.24 new-build and have gotten the interactive session to work by simply replacing `(list "repl")` with `(list "new-repl")` in `haskell-process-compute-process-log-and-command` in haskell-process.el, line 99.

new-build is still very new, but maybe we can introduce an customize option for the "repl" literal?

slpopejoy:
Related: https://github.com/flycheck/flycheck-haskell/issues/65 with discussion of cabal new-build.

codygman:
I'm trying Cabal's new-build functionality and had a need for this as well. I'm just downloading haskell-mode's source and doing the change OP did.

peti:
The fact that the info manual mentions `cabal-new-repl` as a supported process type is very confusing. Why is that option documented already when the code doesn't allow me to actually set it?

gracjan:
Is it documented? That is a proof of time travel right there!

On a more serious note somebody should sit down and clean up those pieces of code, it is a convoluted hack that nobody had time to properly untangle.

slpopejoy:
Presumably #1507 addressed this, and besides cabal is a whole new animal now, so I'll close as the original reporter

purcell:
Yep, thanks!

ukari:
when `defcustom haskell-process-type` is `auto`,  the `defun haskell-process-type` can detect cabal project correctly and returns `cabal-repl`. 

But in https://github.com/haskell/haskell-mode/blob/d771202bcb04c04cdfc6c102f99b917f31216390/inf-haskell.el#L65, the `"repl"` is hardcode, and `haskell-process-args-cabal-repl` has value `'("--ghc-option=-ferror-spans")`.

Should this code be changed to ```('cabal-repl `(,haskell-process-path-cabal ,@haskell-process-args-cabal-repl))``` and `(defcustom haskell-process-args-cabal-repl '("repl" "--ghc-option=-ferror-spans"))`, so that it can support `cabal new-repl` when `haskell-process-type` is `auto` by customize `haskell-process-args-cabal-repl`?

purcell:
@ukari I'm having a bit of trouble understanding the situation you're describing, sorry. Would you mind filing a new issue with steps and expected/actual results please?

PR #1507, opened 2017-05-16
make cabal-new-repl a legitimate process type

peti:
One step further towards fixing https://github.com/haskell/haskell-mode/issues/1433.


gracjan:
Is this enough? Where are the places that case-match on this?

peti:
Yes, this is enough. I'm using new-repl right now.

Issue #1436, opened 2016-10-14
Make ":set +c" optional


vshabanov:
`:set +c` is unconditionally executed at GHCi startup (for `:type-at` in GHC 8+, added in
https://github.com/haskell/haskell-mode/pull/1347) but it considerably slowdowns module loading in large projects.

`Collecting type info for N module(s) ...` takes long time (and GHCi eats a lot of memory). And (although I'm not sure how it's connected) saving of large files becomes painfully slow (I couldn't even move cursor in Emacs for 5-10 seconds).

I've just commented `:set +c` in `haskell-commands.el` and everything went smooth. Could you, please, add an option to disable `:set +c`?

gracjan:
I hate such tradeoffs with a passion. On one hand `:type-at` at friends are mostly useful in projects that you cannot keep in your head, that is larger ones. On the other hand it cannot be used in larger projects due to loading time and memory usage problems.

It basically renders the feature useless in conditions when it is most needed.

vshabanov:
It's a fresh feature that could be optimized in future. Plus it could be useful for novices with small projects.

But at the moment it take about 10 seconds after reloading module. And eats additional 2-4G of RAM after several reloads. I've posted a bug report https://ghc.haskell.org/trac/ghc/ticket/12706 for this. So it worth to add an option till it will be resolved.

vshabanov:
And maybe it worth to add `Collecting type info for N module(s) ...` message, so user could see where time is spent between `Compiling: MyModule` and `OK`. I initially thought that there was a performance bug in haskell-mode, not in GHCi.

gracjan:
Are you able to implement such option? I can review the code if you do.

vshabanov:
Seems that I just need to add `defcustom`. I'll try.

vshabanov:
But adding `Collecting type info for N module(s) ...` message could be too much for me. I have almost no experience with Emacs LISP.

fice-t:
@vshabanov Can you evaluate the following and test to see if the `Collecting type info...` message shows up?

```elisp
(defun haskell-process-live-build (process buffer echo-in-repl)
  "Show live updates for loading files."
  (cond
   ((haskell-process-consume
     process
     (concat "\\[[ ]*\\([0-9]+\\) of \\([0-9]+\\)\\]"
             " Compiling \\([^ ]+\\)[ ]+"
[...]

PR #1515, opened 2017-06-01
Add defcustom for :set +c

slpopejoy:
Adds `haskell-interactive-set-+c` defcustom to switch whether option is added to ghci invocation, and updates error msg on type-at failure. For #1436


gracjan:
Works. Thanks.

Issue #1520, opened 2017-06-20
make check-conventions fails


vasanthaganeshk:
`haskell-code-conventions.el` uses a function  `el-search--search-pattern` from `el-search` module. `el-search--search-pattern` is not available in recent versions of `el-search`. Click [here](https://travis-ci.org/haskell/haskell-mode/jobs/243732195#L1198) to see travis complaining.

gracjan:
This was a good idea but the value delivered wasn't as big as hoped. Basically check=conventions should be removed as it does not bring much to the table.

vasanthaganeshk:
It checks for the lexical binding thing, no tabs for indentation, no spaces or tabs at the end of the line, it could be useful.

I thought it should work with the older version of `el-search` and I copied the definitions of the missing functions from the older version and `make` stops complaining (the code is under GPL-V3, I just put FSF's copyright). But it will need some refactoring later on.

vasanthaganeshk:
Fixed in #1521

Issue #1518, opened 2017-06-10
haskell-collapse modifies the code


vasanthaganeshk:
Consider this code snippet

```
import           Data.List
import qualified Data.Set as Set

main = interact $ unlines . f . tail . map readInt . words

readInt :: String -> Integer
readInt = read

delsundsieve n = filter (<=n) [i+j+2*i*j| i <- [1..e1], let i' = fromIntegral i, j<- [1..(floor ((n' - i')/(2*i' + 1)))]]
  where n' = fromIntegral n
        e1 = floor (sqrt (n' / 2))

inv [] ys = []
inv xs [] = xs
inv a@(x:xs) b@(y:ys)
  | x < y = x:(inv xs b)
  | x > y = inv a ys
  | x == y= inv xs ys

prime n = filter (<= (10^6)) (2:(map (\j -> 2*j + 1) (inv [1..n] (sort (delsundsieve n)))))

f xs = map (\j -> if (Set.member j sv)… then "YES" else "NO") xs
  where sv = Set.fromDistinctAscList (map (\j -> j^2) (prime (5*10^5)))
```

If I select all the code and run `haskell-collapse`, it modifies the buffer.

It looks like this:

```
import           Data.List                                                                                                                                                      
import qualified Data.Set as Set                                                                                                                                                
                                                                                                                                                                                
main = interact $ unlines . f . tail . map readInt . words                                                                                                                      
                                                                                                                                                                                
readInt :: String -> Integer                                                                                                                                                    
readInt = read                                                                                                                                                                  
                                                                                                                                                                                
delsundsieve n = filter … …                                                                                                                                                     
  where n' = fromIntegral n                                                                                                                                                     
        e1 = floor …                                                                                                                                                            
                                                                                                                                                                                
inv … ys = …                                                                                                                                                                    
inv xs … = xs                                                                                                                                                                   
inv a@… b@…                                                                                                                                                                     
  | x < y = x:…                                                                                                                                                                 
  | x > y = inv a ys                                                                                                                                                            
  | x == y= inv xs ys                                                                                                                                                           
                                                                                                                                                                                
prime n = filter … …                                                                                                                                                            
                                                                                                                                                                                
f xs = map … xs                                                                                                                                                                 
  where sv = Set.fromDistinctAscList … 
```

What happens here is that it modifies the buffer (that is emacs asks if we want to save changes while exiting. If I save it, the code is gone!)

gracjan:
Nice catch!

[...]

PR #1521, opened 2017-06-21
Collapse

vasanthaganeshk:
# As of now what works?
+ haskell-collapse-toggle (with tests)
+ `collapse-all `
+ ~temporary fix for #1520~
+ permanent fix for #1520 (removed it)
+ Independent minor-mode
+ default key bindings inspired by [hs-minor-mode](https://www.gnu.org/software/emacs/manual/html_node/emacs/Hideshow.html)

+ docs

# what needs to be done?
+ ~Needs docs~
+ Has to work with braces (now only works for indentation)
+ ~`collapse-all ` optional~
+ `collapse-all-comments` optional
+ ~Make it as independent minor-mode?~
+ ~Add default key bindings~

It is a PR to solve #1518


tests/haskell-collapse-tests.el
```diff
@@ -0,0 +1,124 @@
...
+                 (end-of-line)
+                 (point)))
+        (end (save-excursion
+               (beginning-of-buffer)
+               (forward-line end)
+               (end-of-line)
+               (point))))
+    (cons start end)))
+
+(defun test-haskell-indented-block (source lines result)
```
gracjan: This is really cool!

tests/haskell-code-conventions.el
```diff
@@ -1,12 +1,20 @@
 ;;  -*- lexical-binding: t -*-
 
+;; FIXME
 ;; require 'el-search only on Emacs 25 or above
+;; This file NEEDS REFACTORING. The new el-search as of 20th June 2017
```
gracjan: Just remove it. This idea proved to be more problem that it was worth.

vasanthaganeshk:
done

PR #1525, opened 2017-07-12
[WIP] inf-haskell

vasanthaganeshk:
+ comint based repl mode (with test) with ability to detect project root directory, runs the correct process with `haskell-process-type` ~([little bit of docs](https://haskell-interactive-mode.readthedocs.io~/en/latest/inferior-haskell-interpreter.html))~
+ docs


gracjan:
This is really nice! I liked the documentation the most, great!

Setup for automated tests was also really good!

Please do not touch `inf-haskell.el`, it is a mess but it is out of scope for `haskell-interacive-mode` cleanup. It is too much work to cleanup both at the same time.

And we need way more tests!

PR #1528, opened 2017-07-30
Fix jumping to qualified names

### LINKED ISSUE #1523: Jumping to definition is broken for qualified names
I'm using `hasktags` to generate tags file. This works fine, but if name under point is qualified, `haskell-mode-jump-to-def-or-tag` is not clever enough to strip the prefix up to (and including) the last dot. For example, `T.strip` won't find `strip` even if it's in the tags file.

Maybe `haskell-ident-at-point` is the right place to drop the prefix, but I'm not sure.

mrkkrp:
Fixes #1523.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -32,6 +32,7 @@
 (require 'haskell-font-lock)
 (require 'haskell-interactive-mode)
 (require 'haskell-session)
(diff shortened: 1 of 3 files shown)


mrkkrp:
This works, let me know if something can be improved here.

PR #1534, opened 2017-08-08
[WIP] Everything is Comint

vasanthaganeshk:
+ Only one comint based ghci process per session.
+ Everything is based on comint (`M-.`, `completion-at-point-functions`, showing type of function and other docs at point, loading and reloading file to repl)
+ Docs for `haskell-doc.el` (which defcustoms to modify etc.)
+ Fixing prettification of type information (for example replacing `::` with its unicode alternative with defcustom) in `haskell-doc.el`.
+ `interactive-haskell` has feature duplication with doc mode. We can know the type of identifier at point (in doc mode) or type `C-c C-t` and find the type of identifier at point. Thus `C-c C-t` was removed.
+ loading the module only tries to load the module, reloads if already loaded (doesn't try to provide compilation facilities because this is was a feature duplication before).
+ Compilation revamped


haskell-utils.el
```diff
@@ -189,5 +189,27 @@ expression bounds."
 If given DISABLED argument sets variable value to nil, otherwise to t."
   (setq haskell-mode-interactive-prompt-state (not disabled)))
 
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Set functionalities
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defun make-set ()
```
gracjan: This has to be prefixed with `haskell-`

vasanthaganeshk: Fixed

gracjan:
I tried to checkout out this PR as haskell-mode and your https://github.com/vasanthaganeshk/haskell-interactive-mode as `haskell-interactive-mode` and I see that there is a lot of `haskell-mode` inside of `haskell-interactive-mode`. I thought that `haskell-interactive-mode` will depend on `haskell-mode`, am I right?

PR #1542, opened 2017-08-24
Improve sorting of imports

sergv:
I noticed that buffer sorting does not work for cases like this (`_|_` denoting the cursor):

```
module Foo where

import Bar
import Baz

-- Unrelated group of imports
import Mod3
_|_import Mod2
import Mod1
```

### DIFF
--- a/haskell-sort-imports.el
+++ b/haskell-sort-imports.el
@@ -89,11 +89,15 @@ within that region."
 
 (defun haskell-sort-imports-goto-group-start ()
   "Go to the start of the import group."
-  (or (and (search-backward "\n\n" nil t 1)
-           (goto-char (+ 2 (line-end-position))))
-      (when (search-backward-regexp "^module " nil t 1)
-        (goto-char (1+ (line-end-position))))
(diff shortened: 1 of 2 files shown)


vasanthaganeshk:
I'm just putting it here that there is a built-in way to do this in Emacs. Mark the import lines that you would like to sort and then `M-x sort-lines`.

sergv:
@vasanthaganeshk True, you can sort lines but the built-in function won't properly handle qualified imports. That is, if I have following in my buffer
```
import Foo
import Bar
import qualified Bar as B
```

I'd like to get this after sorting (and `haskell-sort-imports` produces this output)
```
import Bar
import qualified Bar as B
import Foo
```

But I will get this if I use `sort-lines`
```
import Bar
import Foo
import qualified Bar as B
```

sergv:
Bump?

sergv:
Bump^2? Can someone review this, please?

sergv:
Perhaps this project is just abandoned. Too bad.

PR #1545, opened 2017-08-27
Haskell compilaion

vasanthaganeshk:
# Please don't merge this before merging the #1543 PR.

+ New compilation mode

+ The good
 - Works very similarly to existing `haskell-interactive-mode`/`inf-haskell`.
 - It uses `haskell-process-type` for deducing the compilation command. (Different ones in the presence of `cabal.sandbox.config`, `stack.yaml`. Fully explained in the docs [here]().)
 - Standard key bindings from `compile.el` (this existed before, but it is documented now).
+ The bad
 - Breaking change
 - New `defcustoms`
 - Old `defcustoms` removed.


gracjan:
Okay, so this is not that simple:

1. There are projects that have both xxx.cabal and stack.yml and people want to be able to choose which one to use this time. Some projects have stack.yml as convenience only and their authors develop using xx.cabal. Also `find-dominating-file` may find `stack.yml` up the directory tree when author meant only a cabal file in one of the subdirectories.
2. The reason for both `haskell-compile-cabal-build-command` and `haskell-compile-cabal-build-command-alt` is the you may want to normally use a command like `cabal build` and from time to time do a different command like `cabal configure && cabal build`. I'm not sure who is using this kind of workflow, but I find it hard to change.

Overall it would be better to have more principled approach for all of this, based on a sensible definition of 'project' and what does it mean to 'select a project', what is 'project time' and which project does a source file belong to. There is too much magic happening around this.

vasanthaganeshk:
@gracjan I was thinking why not provide a minimal interface like the one provided by `M-x compile` for `haskell-compile`. It is simple and clear. So with this we can run whatever command we want to run and it will open in a new buffer with `HsCompilation` mode.

hvr:
@gracjan This is btw getting even more confusing with new-build, which throws an (optional) `cabal.project` and `cabal.project.local` files (the latter can btw exist even when the former doesn't) into the mix, increasing the combinatorics even more... (you may even end up with those files existing at different levels of your folder hierarchy... fun!) ;-)

gracjan:
There was a very crude project selection interaction going on in `haskell-get-session`. It was confusing and unclear what a project is.

I think we need to define what a project is now, especially because of the backslash that we got on reddit.

Anyway it looks like project management could be a very good selling point of haskell-mode.

PR #1547, opened 2017-08-29
Comint everywhere [WIP]

vasanthaganeshk:


cocreature:
Is there any chance of supporting multiple sessions per Emacs process with `inf-haskell`? Lot’s of people myself included use a workflow with Emacs where they only have one process running for everything and this new architecture seems to contradict that.

gregorycollins:
I have to vote no on this one as well -- it removes functionality that currently works well.

gracjan:
Looks like multiple sessions are important, so we will bring it back.

Note that there are people out there that think in opposite way: they think that managing multiple sessions is confusing and brings nothing to the table because the usual GHCi process is anyway about 4GB of resident RAM occupancy.

gracjan:
@vasanthaganeshk: Can you cut this somehow into even smaller parts? I mean something like this process:

1. Convert to `comint`, keep session management as it is. (this PR)
2. Fix session management properly. (some future PR)

vasanthaganeshk:
Emacs runs on a single thread. If you  are doing some task (opening a big file, starting some process, etc.),  Emacs is going to get stuck for a while. In the meanwhile you might want  to edit something in the other session, the only problem is that Emacs  won't respond to C-x b or anything for that matter. Emacs  was not designed to handle concurrent things. I think running two Emacs  process will be the solution if you want to work on things  simultaneously. Running multiple Emacs process is not taxing on a  modern computer. I think doing things without  leaving Emacs is a little overrated. I don't think multiple sessions is the correct solution to work on two different projects simultaneously.

lambdageek:
Multiple everything is the expected behavior in emacs, IMO.

I often work on multiple constituent subprojects (for example: explore a bug in some downstream project and then try out fixes or potential regression tests in an upstream library) as well as on unrelated projects (work and hobby, for example).

purcell:
Big +1 from me on using `comint` more directly and idiomatically: this is a nice piece of work, and is a welcome move towards simplifying some of `haskell-mode`, which is arguably becoming unwieldy.

I agree that there should be no assumption of only one session per Emacs instance, but that doesn't need to be complicated to achieve:

At the very least, I should be able to set a variable like 'inf-haskell-buffer-name` on a directory-local basis, and then as long as everything uses that name to find the repl associated with the current haskell file, everything should Just Work™.

Anything else on top of that (e.g. functions that automatically create a suitable repl buffer name based) is optional gravy, IMO, and perhaps even undesirable: some people want a single Emacs-wide REPL, while others would like a separate REPL per cabal target (or group of targets).

fommil:
@vasanthaganeshk I really like this PR. Are you still active in Haskell? I'd be interested in trying this out and improving it, maybe under some beta name like `haskell-neo-interactive` or something. There are some other improvements I'd like to make in this area, specifically around: per-buffer (per-target, really) management of interactive compilers, and replacing `haskell-doc-mode` with calls to `:type-of`

fakedrake:
Is this still alive?

purcell:
> Is this still alive?

No, this is long since conflicting.

For the sake of `haskell-mode`'s health I'd be quite keen to throw away all the existing REPL machinery and replace it with a `comint`-based solution like this, but I'm not working enough with Haskell right now to tackle it myself or to deal promptly with any fallout from changing existing behaviour.

fakedrake:
> For the sake of `haskell-mode`'s health I'd be quite keen to throw away all the existing REPL machinery and replace it with a `comint`-based solution like this [...]

I can't put in words how much I agree.

tonyday567:
I followed this PR in helping to cut haskell-mode and it worked out quite well. It was a great PR and thanks to the OA.

See https://github.com/tonyday567/haskell-mode/tree/removals

I'm going to shove this branch into doom, try it out for a few weeks and see how it runs.

fakedrake:
In case you want some extra inspiration troubleshooting or extending this I have switched to my [own](https://github.com/fakedrake/fd-haskell/blob/master/fd-haskell-comint.el)

tonyday567:
Ok, so that's haskell-repl done!

purcell:
Exciting to see you pick this up, @tonyday567. Happy to help land it here in due course.

rntz:
@tonyday567 any news on this?

tonyday567:
@rntz Hi Michael,

I spent a bit of time teasing apart the comint usage, but I couldn't get a handle on the complexity. In my opinion, it would be a lot of work with uncertain outcome, and little hope of acceptance at the end.

I subsequently switched to https://gitlab.com/magus/haskell-ng-mode and found it had everything I needed.

PR #1550, opened 2017-09-06
Format library target names properly

### LINKED ISSUE #1546: haskell-session-change-target fails when exe has same name as library
A [package](https://github.com/ivan-m/jbi) that I'm working on has a library and an application with the same name as the library.

When running `haskell-session-change-target`, I'm presented with an ido selection that says `{jbi | exe:jbi}`.

Selecting the first one (for the library) results in an error (from `*haskell-process-log*`) of:

```
cabal: Ambiguous build target 'jbi'. It could be:
   exe:jbi (component)
   lib:jbi (component)
```

I think we need to prepend `lib:` to the library portion of the available targets.

ivan-m:
Fixes #1546 

I've tested this locally using both cabal-install and stack and it seems to work, and solves my issue with having a package name matching the executable name.


ivan-m:
Thanks @gracjan!

PR #1551, opened 2017-09-06
Better targets

ivan-m:
Two main changes from this (apart from a little bit of code cleanup):

1. Can now get the list of targets from `stack ide targets`, which should help in multi-package projects

2. When starting a session, the prompt of available targets is presented as well (rather than having to remember them or using the default, the latter of which fails in the case of a package having the same name as an executable)

### DIFF
--- a/haskell-session.el
+++ b/haskell-session.el
@@ -27,6 +27,8 @@
 
 ;;; Code:
 
+(eval-when-compile (require 'cl))
+
 (require 'cl-lib)
 (require 'haskell-cabal)
 (require 'haskell-customize)
@@ -181,7 +183,7 @@ HISTORY provides the history to `completing-read'."
 (defun haskell-session-get-targets (process-type)
   "Return a list of available targets."
   (case process-type
-    ('stack-ghci
+    (stack-ghci
      (haskell-session-get-targets-command haskell-process-path-stack "ide" "targets"))
     (t
      (haskell-cabal-enum-targets (haskell-process-type)))))
--- a/haskell-session.el
+++ b/haskell-session.el
@@ -153,10 +153,49 @@ If `haskell-process-load-or-reload-prompt' is nil, accept `default'."
          (target (if maybe-target maybe-target
                    (let ((new-target
                           (if haskell-process-load-or-reload-prompt
-                              (read-string "build target (empty for default):")
+                              (haskell-session-choose-target "Build target (empty for default): " t)
                             "")))
                      (haskell-session-set-target s new-target)))))
     (if (not (string= target "")) target nil)))
+
+(defun haskell-session-choose-target (&optional prompt blank-default history)
+  "Ask the user which of the available targets they want to use.
+Optional arguments:
+
+PROMPT allows you to specify which prompt should be presented to the user.
+
+BLANK-DEFAULT will allow specifying a default blank argument.
+
+HISTORY provides the history to `completing-read'."
+  (let ((prompt (or prompt "Build Target: "))
+        (known-targets (haskell-session-get-targets (haskell-process-type)))
+        (default-target (when blank-default (list ""))))
+    (completing-read prompt
+                     (append default-target known-targets)
+                     nil
+                     nil
+                     nil
+                     history
+                     default-target)))
+
+(defun haskell-session-get-targets (process-type)
+  "Return a list of available targets."
+  (case process-type
+    ('stack-ghci
+     (haskell-session-get-targets-command haskell-process-path-stack "ide" "targets"))
+    (t
+     (haskell-cabal-enum-targets (haskell-process-type)))))
+
+(defun haskell-session-get-targets-command (command &rest args)
+  "Run an external command to obtain a list of available targets."
+  (with-temp-buffer
+    (case (apply #'process-file command nil (current-buffer) t args)
+      (0
+       (cl-remove-if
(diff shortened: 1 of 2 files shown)


gracjan:
```
haskell-session.el:181:37:Error: `(quote stack-ghci)' is a malformed function
make: *** [Makefile:69: build-24.5.1/haskell-session.elc] Error 1
```

ivan-m:
What should it be? It seems to work when I just evaluate the buffer...

Though there is one case where this behaviour might not be ideal: fresh Stack repo, build plan hasn't been downloaded yet, stack insists on downloading it first rather than just stating what the available targets are.

As such, this means it hangs waiting for it to finish.

gracjan:
Well, we need unit test to pass, otherwise, it is hard to merge.

ivan-m:
I think this fixes it, but may there should be an option or timeout on calling stack to get the targets (see commercialhaskell/stack#3423 ).

ivan-m:
I've just noticed that I get a warning about redundant patterns in `pcase`... I think the behaviour of `pcase` changed in 25 :/

ivan-m:
@gracjan the remaining failure with emacs-snapshot is not from my changes.

ivan-m:
In my testing on my work Macbook this worked fine; on my NixOS desktop this code noticeably lags compared to just parsing the `.cabal` file.

I don't know if this is because of the difference in hardware or due to the implicit nix-shell that gets launched (even though I was using nix-shell there as well).

gracjan:
@vasanthaganeshk: this commit goes much into the territory of session management. Can you have a look at it and merge or not merge depending on your feeling about it? Thanks.

vasanthaganeshk:
I'll look into it.

ivan-m:
@vasanthaganeshk have you had a chance to look at this yet?

vasanthaganeshk:
@ivan-m Not yet. Sorry.

ivan-m:
@vasanthaganeshk I'd like to either get this merged in or told otherwise so I can move forward with some other changes I had planned.  I've been using this manually for over a month now (having to remember to evaluate all of these each time I have to restart Emacs as I couldn't work out how to get it to use my fork of `haskell-mode`) without any problems.

The only issue I'm aware of is that `stack query targets` still has a slight delay (see the issue on their tracker above).  I would consider removing the extra empty case as in practice I've found it problematic, but I was duplicating existing behaviour.

vasanthaganeshk:
@ivan-m I'm sorry I can't review this. I should have told this earlier, but I thought could review it. I suggest that you ask someone else to review this patch. Again, I'm sorry.

ivan-m:
@gracjan would you be able to check it then please?

ivan-m:
I have come across one scenario where this may not work very well: if you're trying to use `cabal` and don't have any sandboxes yet.  I'll have to try and see if I can re-produce this (and if so fix it).

ivan-m:
@purcell could you have a look at this please?

haskell-session.el
```diff
@@ -27,6 +27,8 @@
 
 ;;; Code:
 
+(eval-when-compile (require 'cl))
```
purcell: Please use `cl-lib` instead, as elsewhere in this package. This `require` seems to be for `cl-remove-if` anyway, which is in `cl-lib`. (No `eval-when-compile` is needed/desirable here in that case.)

haskell-session.el
```diff
@@ -153,11 +155,50 @@ If `haskell-process-load-or-reload-prompt' is nil, accept `default'."
          (target (if maybe-target maybe-target
                    (let ((new-target
                           (if haskell-process-load-or-reload-prompt
-                              (read-string "build target (empty for default):")
+                              (haskell-session-choose-target "Build target (empty for default): " t)
```
purcell: Nice!

haskell-session.el
ivan-m: It's actually for the `case` statement; byte compilation kept throwing up semi-spurious errors when I didn't have this line.  I'll try it with `cl-lib` instead.

purcell: Ah. Then `cl-case` would be the one to use.

ivan-m: I've changed it to `cl-lib` now but it seems to fail on `HEAD`; should I change it to just `cl`?

purcell: That's probably because you've still wrapped it in `eval-when-compile`, which I mentioned above is unnecessary/undesirable.

ivan-m: Ahhh, right, just `(require 'cl)` instead of `(require 'cl-lib)`?

purcell: Nope, `(require 'cl-lib)` instead of  `(eval-when-compile (require 'cl-lib))`.

ivan-m: The file already has `(require 'cl-lib)`; but without that `eval-when-compile` statement it throws errors during the CI tests.

purcell: I don't think that's true. I made a branch with your changes minus that line, and it gives the [same CI results](https://travis-ci.org/haskell/haskell-mode/jobs/626977476) as [master](https://travis-ci.org/haskell/haskell-mode/jobs/614690412). I'll therefore go ahead and merge my branch.

ivan-m: Huh, that's weird... either way, thanks for reviewing + merging @purcell !

PR #1552, opened 2017-09-06
Compilation buffer keybindings

vasanthaganeshk:
Added key bindings for compilation mode. We could Just link to the [GNU documentation](https://www.gnu.org/software/emacs/manual/html_node/emacs/Compilation-Mode.html#Compilation-Mode) or Merge this PR.


gracjan:
I do not mind some duplication of docs.

PR #1554, opened 2017-09-13
Fix haskell-navigate-imports-go with Bird-style

cydparser:


purcell:
Thanks for this. Sorry it sat unmerged for a long time.

Issue #1496, opened 2017-04-05
Haskell process command errored with: (error "Unexpected response from haskell process.")


Abhiroop:
A Haskell project which builds fine with `stack build` runs into this issue.
```
Unexpected response from haskell process.
Haskell process command errored with: (error "Unexpected response from haskell process.")
```
No helpful error messages to debug too.

identicalsnowflake:
I just upgraded to GHC 8.2.1 and suddenly I'm getting this, too. My project builds fine (no stack involved), but emacs mode is broken - `C-c C-l` gives the error you posted.

gracjan:
This is related: https://github.com/haskell/haskell-mode/pull/1168

Abhiroop:
The issue was solved for me when I `:set -fobject-code` in GHCi. This issue can be closed now, if its solved for @identicalsnowflake

identicalsnowflake:
Where are you setting this? I tried entering `:set -fobject-code` in ghci directly as well as putting it in my `.ghci` file in my home directory, but I still get the `unexpected response` error.

Abhiroop:
This error message might be a common message for different class of errors(I am not sure about this).
For me I faced it while compiling https://github.com/typelead/eta
Setting object code in GHCi worked for me. And I was on GHC 7.10.

identicalsnowflake:
Hmm. In my case I don't think it's a package: I get it even on the default "Hello world" file generated by `cabal init`.

nrolland:
I am confronted to this message. How can we debug that to get more information on what's failing  ?

Abhiroop:
@nrolland This seems like a common catch all error clause whose occurrence entirely depends on the code that you are trying to load. Like mentioned above my issue was solved by setting `:set -fobject-code`, but it is not necessary that the same will work with you.

I faced the same warning when I was trying to load a program with unlifted types. So you would have to produce a snippet of the program you are trying to load to debug it.

nrolland:
@Abhiroop yeah, and since it's generic I imagined there was some "debug" facility to get more info to activate somewhere. 

There is an `haskell-process-log` emacs variable which creates a buffer but that seems to be OK, my module is loaded. 

How did you come up with your idea of setting `:set -fobject-code` ?

nrolland:
For potential future references, there seems to be a bunch of heuristics in `haskell-load.el` in `haskell-process-load-complete` on how to interpret the response.

Replacing the function which does the matching to 

```
[...]

PR #1555, opened 2017-09-16
option to disable ':set prompt2" for ghci sessions. ghci>=8.2 breaks

alios:
One of the issues after upgrading to ghc 8.2.1 I had (next to parsing the new default ghc(i) error log messages) was that when starting a new interactive session (stack ghci) it automaticaly issued an `:set prompt2 <prompt>` to the ghci which 8.2.1 no longer accepts and creates a warning.

For now I simply added a new config variable `haskell-interactive-set-multiline-prompt` to disable the `:set prompt2` output.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -49,6 +49,12 @@
   :group 'haskell-interactive
   :type 'boolean)
 
+(defcustom haskell-interactive-set-multiline-prompt
+  t
+  "Issue ':set prompt2' in interactive session. Not supported by ghci >= 8.2"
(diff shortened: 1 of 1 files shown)


zenspider:
bump

peti:
https://github.com/haskell/haskell-mode/issues/1496#issuecomment-412348006 suggests using `:set prompt-cont` instead of `:set prompt2`.

purcell:
I've addressed this by switching to `prompt-cont` throughout.

PR #1563, opened 2017-10-14
Fix haskell-process-load-complete for GHC 8.2.1

bjc:
At some point by GHC 8.2.1, the text for successful load of a module
into ghci changed. This patch allows for the new format, in addition
to the existing ones.


purcell:
Thanks for this - I made a further tweak in eac5567e2a47be1272ed73b19c43025a457fa416. Merging now.

PR #1565, opened 2017-10-21
Provide customization option `haskell-process-show-overlays'

purcell:
When using haskell-interactive-mode's REPL support in environments that already have in-buffer flycheck support (e.g. `intero', `dante'), the redundant overlays become problematic and confusing for users. This commit adds a customization option to suppress haskell-mode's overlays so that users with such set-ups can choose to see only their flycheck errors.

Submitting this as a PR for visibility, but will assume this is okay to merge in a day or two if no other maintainers have an objection, since it seems fairly uncontroversial.

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -200,6 +200,13 @@ See `haskell-process-do-cabal' for more details."
   :type 'boolean
   :group 'haskell-interactive)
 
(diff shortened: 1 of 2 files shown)


purcell:
Build failures appear spurious. :-/

Issue #1560, opened 2017-10-10
C-c C-l gives unhelpul error


monnier:
I guess this is related to #522 (but I couldn't figure out how to re-open it).
Now when I try <kbd>C-c C-l</kbd>, I get an error saying

    Run ‘C-h f haskell-mode‘ for instruction how to setup a Haskell interaction mode.

but then if I do `C-h f haskell-mode`, I only get told that there are:

    Interaction modes:

        ‘interactive-haskell-mode’
          Interact with per-project GHCi processes through a REPL and
          directory-aware sessions.

        ‘inf-haskell-mode’
          Interact with a GHCi process using comint-mode. Deprecated.

[ the last part has obviously a problem since I can;t find any `inf-haskell-mode`.  ]

The problem here is that nothing tells me how to do what I wanted with <kbd>C-c C-l</kbd>.  Looking at commit ca94d814b943afeea5bc418df3773f19be9ac8e3 is worse: it says "revamped" but it should just say "gutted", since the functionality that was bound to <kbd>C-c C-l</kbd> was simply thrown away.  That seems completely pointless: if you want to deprecate `inf-haskell.el`, then just let it rot, but throwing away some of its code while keeping a tiny mostly useless shell is just infuriating the user who gets the impression he might be able to find his old <kbd>C-c C-l</kbd> somewhere in there.

purcell:
@vasanthaganeshk - anything we can do to make this less frustrating?

vasanthaganeshk:
Removed the inconsistent docs. I think something along the lines of asking the user to start `interactive-haskell-mode` must be sufficient.

monnier:
Why not just enable that mode by default?  After all, the other mode is deprecated (and has been emptied of its contents on top of that), so the inf-haskell-mode is simply not a usable option any more.  The choice imposed on the user is a complete waste of time for her.

vasanthaganeshk:
It is my opinion that we don't make `interactive-haskell-mode` to be default because some people might want to use Intero and some people might want to use lsp based minor mode (some work seems to be going on in haskell ide engine). One more thing is that some people might have configured key bindings used in `interactive-haskell-mode` to do something else. Some people might not want to use `interactive-haskell-mode` at all because with larger code bases Emacs might get slow with loading the big code base into the GHCi and the queries to GHCi might take longer to respond.

monnier:
[...]

PR #1568, opened 2017-10-22
removed remanant docs about inf-haskell-mode, that was purged

vasanthaganeshk:
fix for #1560


purcell:
Thanks. I cherry-picked the two relevant commits here and dropped the `Dockerfile` one. Sorry it took so long for someone to respond.

PR #1574, opened 2017-11-23
limit ubuntu to zesty for emacs24

pvmart:
There isn't any emacs24 on later versions of ubuntu


purcell:
Thanks for this - sorry your PR languished for so long without even a comment!

Issue #651, opened 2015-05-15
stylish-haskell integration no longer reports errors


MichaelXavier:
Often in haskell development, the buffer I'm working on will be invalid for stylish haskell: missing MultiParamTypeClasses, syntax error, etc. If I recall, stylish-haskell used to drop the error it was getting into the minbuffer. Recently, is just says: "Error: stylish-haskell ended with errors, leaving buffer alone". In order to fix the error, I have to leave my editor and manually run `stylish-haskell file.hs | head` to get the error. Seems like the formatter should capture the stderr of the process and leave it somewhere.

gracjan:
This change was motivated by #557 and done in #557.

As we do not know if buffer is or is not valid Haskell it is an expected condition that stylish-haskell fails. if you wish it reported then we need to find a better way of handling the error.

@MichaelXavier: do you have a more concrete proposal?

LukeHoersten:
I would recommend using something like flycheck-haskell to check for errors on the fly since it is purpose built for that. Stylish-haskell is only reporting why it couldn't format your code and isn't intended for on-the-fly debugging like flycheck is.

MichaelXavier:
@LukeHoersten I'm not really trying to use stylish-haskell for debugging. There are cases if I recall  where the program will compile but stylish-haskell will refuse to parse it (I think to do with missing extensions like MultiParamTypeClasses). Sometimes you're concerned with formatting independently of errors in the code.

@gracjan Thanks for the heads up. That works for me! One thing I'd recommend though is maybe to recommend setting `warning-minimum-log-level` to debug rather than `warning-minimum-level`. That seems like it would satisfy more people because it won't pop up in your face but it won't suppress the actionable information from your warnings buffer either.

PR #1579, opened 2018-01-09
Log stderr into separate buffer

corpix:
I have come with similar use case as in #651

I just want to be able to see the `stylish-haskell` stderr because it could fail with errors like:

```
stylish-haskell: Language.Haskell.Stylish.Editor.applyChanges: refusing to make overlapping changes on lines 59-60, 59-60
CallStack (from HasCallStack):
  error, called at lib/Language/Haskell/Stylish/Editor.hs:46:28 in stylish-haskell-0.8.1.0-Dqoruy2bCD1Kn6ytiOdLeZ:Language.Haskell.Stylish.Editor
```

59-60 lines in my code:

> Declaration uses unicodeSyntax but it is not consistent: `... Integer -> IO ...`

``` haskell
getLimitedRange ∷ Integer → Maybe Integer → Maybe Integer -> IO (UTCTime, UTCTime)
getLimitedRange limit fromTime toTime = do ...
```

This kind of errors is not covered by flycheck and running `stylish-haskell File.hs` every time I see `Error: stylish-haskell ended with errors, leaving buffer alone` in minibuffer is just boring.

_I don't know much about elisp, sorry for any strange things in the code :)_

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -832,7 +832,11 @@ output.  If CMD fails the buffer remains unchanged."
                   (insert-file-contents out-file nil nil nil t))
               (progn
                 ;; non-null stderr, command must have failed
-                (message "Error: %s ended with errors, leaving buffer alone" cmd)
+                (with-current-buffer
+                    (get-buffer-create "*haskell-mode*")
+                  (insert-file-contents err-file)
+                  (buffer-string))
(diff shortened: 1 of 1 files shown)


corpix:
Travis build is broken https://travis-ci.org/haskell/haskell-mode/jobs/326853841#L815
Probably `emacs24-nox` package is not available in `ubuntu:rolling`

purcell:
Thanks for this - sorry your PR sat ignored for so long!

PR #1581, opened 2018-03-13
[security] Don't make functions safe file-local values

thblt:
Remove `:safe 'functionp` from the definition of the `haskell-process-wrapper-function` Customize variable.  

Marking values which validate `functionp` as "safe" meant that any file could potentially execute any code on the Emacs instance it's been open.  This commit removes this possibility.  

Thanks!


purcell:
Makes sense - thanks for this. Sorry you didn't get a faster response.

PR #1582, opened 2018-03-15
Add note on installing from git on macOS to the README.

alxp:


ivan-m:
Or just install the package from MELPA?

alxp:
The normal way of installing works fine, but I was wanting to test pull requests on the project and ran into the problems I noted. I updated the README change to make it more clear those instructions are specifically for installation via git. Hope this helps.

purcell:
Thanks!

PR #1587, opened 2018-04-08
Add display of SVG images in REPL.

fedeinthemix:
When working on SVG images, it's sometimes useful to see them directly as the REPL output within Emacs.


fakedrake:
I wish I had that too!

purcell:
This is really neat. Merging, thanks!

PR #1589, opened 2018-04-16
If the hasktags command fails `haskell-mode-generate-tags` throws an error

fakedrake:
_PS. This is partly me testing the waters while contributing infinitesimally._


purcell:
Closing because this is generating spurious new notifications.

Issue #1592, opened 2018-04-24
haskell-interactive-mode: provide jump to previous prompt 


andreas-roehler:
Functionality is known from shell as command `comint-show-output`. Can't see it so far.

andreas-roehler:
Okay, found haskell-interactive-mode-prompt-previous etc.

PR #1593, opened 2018-04-24
#1592, haskell-interactive-mode: provide jump to previous prompt

andreas-roehler:


andreas-roehler:
Found haskell-interactive-mode-prompt-previous, not need any more.

PR #1597, opened 2018-05-24
Add support for `signature`s and .hsig files

alexbiehl:
This adds font-lock, indentation and recognition for GHCs Backpack signatures. @purcell


purcell:
Makes sense to me - thanks.

PR #1599, opened 2018-06-01
Add haskell-doc-show-ghci that pings GHCI for type information.

czipperz:
This a last resort.  It is slow, but it works.  It is disabled by default because it can make navigating large files extremely slow.

On my computer it takes about 0.5 seconds to query a 700 line code base.


czipperz:
This is effectively equal to `haskell-process-do-type` done automatically.

PR #1603, opened 2018-06-28
Improve tags excludes

ntc2:
The `haskell-cabal--compose-hasktags-command` collects Haskell source
files and sends them to `hasktags`. Before it only excluded build
directories -- such as `.stack-work` (Stack) or `dist` (Cabal) -- at
the top level. Now it excludes them everywhere in the project, and
additionally excludes `dist-newstyle` (Cabal 2) dirs.

For a large project I'm working on that has many Git submodules, this
change eliminates 0.5MB of spurious `TAGS` entries in nested
`.stack-work` and `dist-newstyle` dirs.

Also, unrelated, I put a note in CONTRIBUTING.md about using `emacs -L <path to haskell-mode.git>` to test your changes to Haskell Mode. Took me awhile to figure that out the first time around.


purcell:
LGTM, thanks.

PR #1607, opened 2018-08-18
Fix error when haskell-process-path-ghci has arguments

kuribas:
If I give arguments to haskell-process-path-ghci run-haskell fails with an error.


purcell:
LGTM, thanks.

PR #1608, opened 2018-09-01
support ANSI colours in the compilation buffer

fommil:
For example, when tasty is called with `--color=always` // @feuerbach

Disabled by default to avoid any performance impact, but perhaps we can do a test cycle and if people like it maybe turn it on by default?

BTW, I think there is a bug in Emacs... I had to use `(point-max)` but the docs said `(point)` should work.

2018-09-02-004013_1918x1078_scrot

### DIFF
--- a/haskell-compile.el
+++ b/haskell-compile.el
@@ -28,6 +28,7 @@
 
 (require 'compile)
 (require 'haskell-cabal)
+(require 'ansi-color)
 
 ;;;###autoload
 (defgroup haskell-compile nil
@@ -59,6 +60,11 @@ The `%s' placeholder is replaced by the current buffer's filename."
 (defcustom haskell-compile-ghc-filter-linker-messages
   t
   "Filter out unremarkable \"Loading package...\" linker messages during compilation."
+  :group 'haskell-compile
+  :type 'boolean)
+
+(defcustom haskell-compile-color nil
+  "When non-nil will render ANSI color sequences correctly."
   :group 'haskell-compile
   :type 'boolean)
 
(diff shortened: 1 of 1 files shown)


fommil:
oh, and there is also a missing regex to catch failed tasty tests as warnings... but I'll do that as a follow up.

fommil:
The build failed for flakiness reasons unrelated to my proposed change:

```
The command '/bin/sh -c apt-get update     && apt-get install -y build-essential     && apt-get install --no-install-suggests --no-install-recommends -y emacs24-nox     && apt-get install -y ghc     && apt-get install -y haskell-stack' returned a non-zero code: 100
```

fommil:
@purcell the `#haskell-emacs` room said that you're the maintainer, is that correct? Could you please do a review of this and merge if you're happy? The CI seems to be broken. I wonder if perhaps moving to gitlab, or using a custom docker image, would make this more reliable.

purcell:
> #haskell-emacs room said that you're the maintainer, is that correct?

LOL, just the only person occasionally looking at this in recent months, afaict. I've no bandwidth to fix up the build process currently, but the change looks reasonable. I only wonder why we wouldn't just unconditionally handle ANSI colours: I'm not convinced we need a new flag var.

fommil:
@purcell hehe, I know the score.

Re new var, I just didn't want to break anybody's workflow. I plan on following up with a change to make it the default if I get say 10 reports of it working.

haskell-compile.el
```diff
@@ -96,7 +102,12 @@ This is a child of `compilation-mode-map'.")
     (delete-matching-lines "^ *Loading package [^ \t\r\n]+ [.]+ linking [.]+ done\\.$"
                            (save-excursion (goto-char compilation-filter-start)
                                            (line-beginning-position))
-                           (point))))
+                           (point)))
+
+  (when haskell-compile-color
+    (read-only-mode -1)
```
purcell: Here, I believe you can let-bind `inhibit-read-only` IIRC.

haskell-compile.el
```diff
@@ -62,6 +63,11 @@ The `%s' placeholder is replaced by the current buffer's filename."
   :group 'haskell-compile
   :type 'boolean)
 
+(defcustom haskell-compile-color nil
```
purcell: Since you're talking about a follow-up change later, let's just be bold and do this unconditionally.

fommil: ok, will do!

fommil:
@purcell changes applied and eyeballed that it works.

purcell:
Sweet, thanks!

PR #1609, opened 2018-09-01
failed tasty tests are errors

fommil:
follow up to #1608 (ansi colours included in this screenshot to show them playing well together)

Note that `next-error` (i.e. jump to error) works as expected... and everything you see that is underlined is clickable, if you are so inclined as to use something as clumsy as a mouse :stuck_out_tongue_closed_eyes: 

2018-09-02-005739_1918x1078_scrot

### DIFF
--- a/haskell-compile.el
+++ b/haskell-compile.el
@@ -78,8 +78,8 @@ The `%s' placeholder is replaced by the current buffer's filename."
      1 2 4 0) ;; info locus
 
     ;; failed tasty tests
-    ;; error, called at src/Course/State.hs:42:3 in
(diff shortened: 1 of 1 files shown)


fommil:
Unfortunately this doesn't catch this sort of failure

```
  FAIL
        test/Course/StateTest.hs:91:
        expected: (Full 'c',3)
         but got: (Full 'c',8)
      find 'i' in 'a'..'h':                                                          
```

fommil:
latest commit should address that

purcell:
Thanks!~

Issue #1611, opened 2018-09-05
Don't take C-a key


andreas-roehler:
haskell-interactive-mode.el sets

(define-key map (kbd "C-a") 'haskell-interactive-mode-beginning)

which is bad, as C-a is a very basic edit key.
Shadows beginning-of-line that way.

As C-r C-p works fine, don't see a need for this.

Thanks,
Andreas,
~/.emacs.d/elpa/haskell-mode-20180601.843/

bradyt:
Maybe C-a should be bound to a function that goes to beginning of line, unless prompt is on same line. The issue seems to be that the previous function was going to prompt even if you were on a different line.

purcell:
@bradyt Are you reporting an issue?

bradyt:
I don't understand the question.

purcell:
> @bradyt Are you reporting an issue?

To clarify, I think that since merging the fix for this (I forgot to close the issue), the behaviour is correct. But if not, please let us know.

null-a:
Having `C-a` move the cursor to after the prompt (for single line edits), seemed pretty natural to me. It's a shame that this is no longer the default behaviour. Just my 2 cents.

andreas-roehler:
Yeah, sorry, that suggestion went wrong. Please revert that commit/merge.
C-a should stop before prompt when on first line, but go to BOL in multiline-edit.

andreas-roehler:
However, C-a as  ‘haskell-interactive-mode-beginning’ should be called only, if
on first line with prompt.
Probably needs a haskell-mode-beginning-of-line.

bradyt:
The suggestion I heard when this came up on IRC recently, was something like the following.

If the prompt prefix has somehow been made read only, then after going to beginning of line, check if there is a first position where text switches from read-only to not. If that exists, put point there.

But that assumes that the haskell repl has gone with the idiom of a read-only prompt prefix.

andreas-roehler:
Proposed a fix: https://github.com/haskell/haskell-mode/issues/1666

Not perfect yet, but should provide some amendment.

PR #1612, opened 2018-09-05
#1611, Don't take C-a key fixed

andreas-roehler:


purcell:
Thanis. What's the rationale for this change?

purcell:
Belated thanks!

PR #1616, opened 2018-10-07
Fix haskell-process-load-complete for ghc 8.4+

drvink:
Newer GHC versions will emit messages containing English words for numbers in place of Latin script numerals, e.g. "Ok, five modules loaded." instead of "Ok, 5 modules loaded."; a regex meant to match such messages is updated in accordance. This fixes commands such as `haskell-process-load-file-complete`, which were failing due to no regex matching the emitted message.

### DIFF
--- a/haskell-load.el
+++ b/haskell-load.el
@@ -107,15 +107,11 @@ actual Emacs buffer of the module being loaded."
   (let* ((ok (cond
               ((haskell-process-consume
                 process
-                "Ok, \\(?:[0-9]+\\) modules? loaded\\.$")
-               t)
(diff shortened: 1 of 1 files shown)


drvink:
(The test "failure" here is due to `apt-get` not having been able to fetch some package lists when the CI ran, i.e. nothing to do with the fix in this PR.)

drvink:
Can someone please take a look at this? The problem and the fix are trivial.

gregorycollins:
Sorry I didn't see this pre-existing PR before I merged the very similar https://github.com/haskell/haskell-mode/pull/1625/files

PR #1619, opened 2018-10-19
haskell-compile supports stack / cabal / ghc equally

fommil:
This PR adds support for detecting which build tool is in use when `M-x haskell-compile` is used, which I recommend be bound to `C-c c` with

```lisp
(bind-key "C-c c" 'haskell-compile haskell-mode-map)
(bind-key "C-c e" 'next-error haskell-mode-map)
```

and to remember custom commands if called with `C-u`. To avoid breaking backwards compatibility, I can remove this feature from the `cabal` and `ghc` codepaths, but I think that would be very confusing.

Some decisions in this PR are unexpected and I'd like to explain them:

- why is `stack` detection first? Because stack projects often have `.cabal` files in submodules leading to incorrect detection of build tool.
- why not use `locate-dominating-file` for cabal? This is the existing behaviour and I didn't want to break anybody's workflow.
- why is the "last" command global and not buffer local? I am open to changing this... both have their merits, and it is a lot easier for somebody to make something be buffer local than to make something local become global.
- why have you written so much duplicating code, I want to rip out my eyes!? Pure incompetence, I'll try to clean up my lisp with follow ups...

### DIFF
--- a/haskell-compile.el
+++ b/haskell-compile.el
@@ -50,6 +50,20 @@ The `%s' placeholder is replaced by the cabal package top folder."
   :group 'haskell-compile
   :type 'string)
 
+(defcustom haskell-compile-stack-build-command
+  "cd %s && stack build --fast"
+  "Default build command to use for `haskell-stack-build' when a stack file is detected.
+The `%s' placeholder is replaced by the stack package top folder."
+  :group 'haskell-compile
+  :type 'string)
+
+(defcustom haskell-compile-stack-build-alt-command
+  "cd %s && stack clean && stack build --fast"
+  "Alternative build command to use when `haskell-stack-build' is called with a negative prefix argument.
+The `%s' placeholder is replaced by the stack package top folder."
+  :group 'haskell-compile
+  :type 'string)
+
 (defcustom haskell-compile-command
   "ghc -Wall -ferror-spans -fforce-recomp -c %s"
   "Default build command to use for `haskell-cabal-build' when no cabal file is detected.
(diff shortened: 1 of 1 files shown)


haskell-compile.el
```diff
@@ -143,19 +158,50 @@ node `(haskell-mode)compilation' for more details."
...
-    (when (and edit-command (not (eq edit-command '-)))
-      (setq command (compilation-read-command command)))
+  (if-let ((stackdir (locate-dominating-file buffer-file-name "stack.yaml")))
+      (haskell-compile-stack stackdir edit-command)
+    (if-let ((cabaldir (haskell-cabal-find-dir)))
+        (haskell-compile-cabal cabaldir edit-command)
+      (let ((srcfile (buffer-file-name)))
+        (haskell-compile-ghc srcfile edit-command)))))
+
+(setq haskell-compile-stack-last nil)
```
matthew-piziak: These should probably be `defvar`s.

haskell-compile.el
```diff
@@ -143,19 +158,50 @@ node `(haskell-mode)compilation' for more details."
...
+(defun haskell-compile-stack (dir edit-command)
+  (let* ((default (or haskell-compile-stack-last
+                      haskell-compile-stack-build-command))
```
matthew-piziak: I can't find defs for `haskell-compile-build-command` or `haskell-compile-build-alt-command`.

haskell-compile.el
```diff
@@ -143,19 +158,50 @@ node `(haskell-mode)compilation' for more details."
...
+                     ('nil default)
+                     ('-  haskell-compile-stack-build-alt-command)
+                     (_   (compilation-read-command default))))
+         (command (format template dir)))
+    (unless (eq edit-command '-)
+      (setq haskell-compile-stack-last template))
+    (compilation-start command 'haskell-compilation-mode)))
+
+(setq haskell-compile-cabal-last nil)
+(defun haskell-compile-cabal (dir edit-command)
```
matthew-piziak: I see nothing wrong with using macros here for `ghc/cabal/stack`.

haskell-compile.el
fommil: I have no idea why I didn't do that... :confused:

haskell-compile.el
fommil: I think I'd like to add a lot more functionality to this procedure, eventually (e.g. test/run/other support). Making it a macro will introduce tech debt that may be hard to remove. I'll look into passing the symbol of the `-last` to see if that's a possible way to update the state.

haskell-compile.el
fommil: good catch...

fommil:
@chrisdone you may like this :smile:

fommil:
CI is failing for the same reason it always fails...

```
The command '/bin/sh -c apt-get update     && apt-get install -y build-essential     && apt-get install -y emacs-nox     && apt-get install -y ghc     && apt-get install -y haskell-stack' returned a non-zero code: 100
```

I am using this code daily, so I can vouch for it. Could somebody else please try this out and confirm it doesn't make their world implode?

fommil:
@hvr argh, you're right, this is potentially a very annoying regression for people in that situation. It's an easy fix, I'll push something today.

fommil:
@hvr that should fix it...

fommil:
btw, looks like the best way to support more than one compile at a time is to rename the buffer, e.g. https://stackoverflow.com/questions/4367371

We could rename to `*haskell-compile-<build tool>-<directory>*` which I think would be good.

fommil:
ah! The docs are always better than stack overflow

```
If NAME-FUNCTION is non-nil, call it with one argument (the mode name)
to determine the buffer name.  Otherwise, the default is to
reuses the current buffer if it has the proper major mode,
else use or create a buffer with name based on the major mode.
```

This is too useful, I'll make the change now...

oscarvarto:
@hvr Is this PR useful, now that cabal is the preferred choice?

hvr:
@oscarvarto yes, it's better now; I did want to give it a test-run but didn't have time yet, hence why I didn't comment yet.

hvr:
lgtm :-)

PR #1620, opened 2018-10-19
Changes to Hoogle

tsoernes:


haskell-hoogle.el
```diff
@@ -94,6 +96,18 @@ is asked to show extra info for the items matching QUERY.."
               "hoogle" "server" "-p" (number-to-string haskell-hoogle-port-number))))
     (error "\"hoogle\" executable not found")))
 
+(defun haskell-hoogle-start-stack-server ()
+  "Start hoogle local server using the local Stack configuration."
+  (interactive)
+  (if (executable-find "stack")
```
fommil: I don't think you need this check, if the executable doesn't exist the user will get a suitable error anyway.

fommil:
Thanks! I had been meaning to add something like this.

tsoernes:
I think the previous build failed because of a repo being down. I removed the check you mentioned.

fommil (approved):

Issue #1615, opened 2018-09-24
':' doesn't get recognized as a char


neongreen:
![image](https://user-images.githubusercontent.com/1523306/45957821-3c670c80-c016-11e8-95e2-f7e71f778a04.png)

ivan-m:
`':'` seems to have `haskell-constructor-face` (with the quotes being `default`) applied to it still, instead of `font-lock-string-face` as in `'a'`.

This isn't the case with `":"` though.

Wild guess: it's part of what allows you to use `':` in type-signatures causing it.

alexpeits:
This issue was related: https://github.com/haskell/haskell-mode/issues/1508 and I think https://github.com/haskell/haskell-mode/pull/1578 causes the current issue

PR #1624, opened 2018-11-20
Fix lexer to read ':' as character

alexpeits:
This PR fixes issue https://github.com/haskell/haskell-mode/issues/1615. Another related issue (now resolved) is https://github.com/haskell/haskell-mode/issues/1615. The character lexeme regex does not take into account the single colon character, which caused haskell-mode to interpret it as the type operator `':` (enabled with `DataKinds` and `TypeOperators`). This generally did not cause any issues, however in my case (in a parsing function), if `':'` was followed by a closed paren it confused the lexer and caused the rest of the lines to be indented the wrong way (also notice the red closing quote):

screenshot from 2018-11-20 18-31-38

### DIFF
--- a/haskell-lexeme.el
+++ b/haskell-lexeme.el
@@ -218,7 +218,7 @@ only escape sequences defined in Haskell Report.")
 
 (defconst haskell-lexeme--char-literal-rx
   (rx-to-string `(: (group "'")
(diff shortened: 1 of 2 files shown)


purcell:
Thanks. I feel like this might only be a partial fix for the general issue of the regexp being overly specific, but it makes sense to merge it as-is. Sorry for the long delay.

PR #1625, opened 2018-11-20
Fix loading for GHC 8.4, suggested by gregorycollins

### LINKED ISSUE #1553: Parsing problem with ghc-8.2.1
It seems the messages when loading a file have changed with ghc-8.2.1.
I got the error "Unexpected response from haskell process".
After looking into it, I fixed it by changing the regex in haskell-load.el. My patch is :

--- Source/hs/haskell-load.el	2017-09-06 12:24:28.341913600 +0200
+++ .emacs.d/elpa/haskell-mode-20170903.948/haskell-load.el	2017-09-06 12:30:54.954209500 +0200
@@ -107,11 +107,11 @@
   (let* ((ok (cond
               ((haskell-process-consume
                 process
-                "Ok, modules loaded: \\(.+\\)\\.$")
+                "Ok, \\([[:digit:]]+ \\)?modules? loaded[.:]\\(.*\\)$")
                t)
               ((haskell-process-consume
                 process
-                "Failed, modules loaded: \\(.+\\)\\.$")
+                "Failed, \\([[:digit:]]+ \\)?modules? loaded[.:]\\(.*\\)$")
                nil)
               (t
                (error (message "Unexpected response from haskell process.")))))

I didn't check it's working on previous ghc versions, but it's meant to do it.

sjsch:
@gregorycollins suggested this on #1605, which fixes #1553 for the case where no modules are loaded.

### DIFF
--- a/haskell-load.el
+++ b/haskell-load.el
@@ -111,7 +111,7 @@ actual Emacs buffer of the module being loaded."
                t)
                ((haskell-process-consume
                 process
-                "Ok, \\(?:[a-z]+\\) module loaded\\.$") ;; for ghc 8.4
+                "Ok, \\(?:[a-z]+\\) modules? loaded\\.$") ;; for ghc 8.4
                t)
               ((haskell-process-consume
                 process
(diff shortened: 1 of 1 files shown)


gregorycollins:
I think there's some other stuff needed too -- I will submit a PR with my local patch (that fixes repl for me)

gregorycollins:
> I think there's some other stuff needed too -- I will submit a PR with my local patch (that fixes repl for me)

....actually, this patch is byte-identical with my local one and it looks like the CI failed for other reasons.

PR #1628, opened 2018-12-21
Fix for "Unexpected output.." when modules load failed on GHC 8.6.3

gzh:
Looks like there was similar fix for successful modules load with GHC 8.4 (commit e6169a6597c79e53aa3632c8455c5ba10fa2ba74). I've just upgraded to 8.6.3 and struggled with it's output. Found out that it sometimes says "Failed, five modules loaded.". Respective fix attached.

### DIFF
--- a/haskell-load.el
+++ b/haskell-load.el
@@ -119,6 +119,10 @@ actual Emacs buffer of the module being loaded."
                nil)
               ((haskell-process-consume
                 process
+                "Failed, \\(?:[a-z]+\\) modules? loaded\\.$") ;; ghc 8.6.3 says so
+               nil)
+              ((haskell-process-consume
+                process
                 "Ok, modules loaded: \\(.+\\)\\.$")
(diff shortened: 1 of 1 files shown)


bezirg:
I ran into the same problem with Emacs-26.1, GHC-8.6.3, and latest haskell/haskell-mode . The above PR indeed fixed the problem with the `interactive-haskell-mode` and ghc-8.6.3

peti:
I can confirm that this patch works. `ghc-8.6.4` sometimes responds with a spelled out number, e.g.

    Failed, two modules loaded.

when the number of loaded modules is sufficiently small. This used to break `haskell-mode` all the time, but applying this patch fixes the issue.

I don't know what these errors on Travis-CI are about, but I very much doubt that these failures are in any way related to this patch.

purcell:
Thanks for this. Sorry it languished so long.

PR #1633, opened 2019-01-18
Remove extra loading of haskell-indentation-mode

massysett:
haskell-indentation-mode is already loaded by default in the
haskell-mode-hook.  By forcing it to load every time haskell-mode loads,
the user can disable it only by toggling it off again; removing
haskell-indentation-mode from haskell-mode-hook will have no effect.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -845,8 +845,7 @@ Minor modes that work well with `haskell-mode':
   (add-hook 'completion-at-point-functions
             'haskell-completions-completion-at-point
             nil
-            t)
-  (haskell-indentation-mode))
+            t))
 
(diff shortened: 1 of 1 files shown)


purcell:
Makes sense, thanks.

fommil:
@purcell can you please revert? See linked ticket.

The startup logic for indentation is ... complicated... for reasons. People have gone to great pains to ensure that legacy indentation modes still work for the outliers who prefer the old code and this has resulted in some hacks to the startup hooks.

fommil:
umm, sorry, I just reverted locally and it doesn't fix it. Hmm, this is most perplexing.

fommil:
I think this might be Emacs 26.2 related. I just reverted to the melpa stable version and still see this.

PR #1635, opened 2019-01-24
haskell-cabal-mode: Add support for common-stanzas

### LINKED ISSUE #1627: Support common stanzas
Cabal 2.2 introduced `common` stanzas, but current haskell-cabal-mode does not support one.
Could you please support `common` stanzas?
Thank you for maintaining this great mode.

david-christiansen:
Fixes #1627.

This is mostly an update to the keywords used for highlighting, but there is also a command and a keybinding for consistency with the rest of the mode.

One thing to be aware of: <kbd>M-g c</kbd> is typically bound to `goto-char`, and this shadows that binding in `haskell-cabal-mode`. I think that the consistency with the other commands in this mode is more important that this very-rarely-used binding, but that's a judgment call. I'm happy to remove that default binding if you'd like.

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -153,7 +153,7 @@ it from list if one of the following conditions are hold:
     (define-key map (kbd "M-g l") 'haskell-cabal-goto-library-section)
     (define-key map (kbd "M-g e") 'haskell-cabal-goto-executable-section)
     (define-key map (kbd "M-g b") 'haskell-cabal-goto-benchmark-section)
-    (define-key map (kbd "M-g c") 'haskell-cabal-goto-common-section)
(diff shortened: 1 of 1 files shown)


purcell:
> One thing to be aware of: M-g c is typically bound to `goto-char`, and this shadows that binding in `haskell-cabal-mode`. I think that the consistency with the other commands in this mode is more important that this very-rarely-used binding, but that's a judgment call. I'm happy to remove that default binding if you'd like.

Looks good, but how about `M-g o` (cOmmon), which isn't bound by default

purcell:
Never mind, I went ahead and made that change directly and merged all of this. Thanks, and sorry for the long delay.

david-christiansen:
Thank you for your tireless work!

PR #1636, opened 2019-02-21
Switch to a better Hoogle

magthe:
This version of Hoogle is newer and has a larger index.


purcell:
Thanks, and sorry for the delay.

PR #1640, opened 2019-03-29
Allow user to provide custom hoogle server command

sevanspowell:
I use it like this:

```
  (setq haskell-hoogle-server-command (lambda (port) (apply haskell-process-wrapper-function (list (list "hoogle" "server" "--local" "-p" (number-to-string port))))))
```

to launch my hoogle in a nix-shell.

### DIFF
--- a/haskell-hoogle.el
+++ b/haskell-hoogle.el
@@ -48,6 +48,16 @@ If nil, use the Hoogle web-site."
           (const :tag "hayoo" "http://hayoo.fh-wedel.de/?query=%s")
           string))
 
+(defcustom haskell-hoogle-server-command (lambda (port)
+                                           (list "hoogle" "server"
+                                            "--local"
+                                            "-p"
+                                            (number-to-string port)))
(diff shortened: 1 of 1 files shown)


purcell:
This kinda makes sense in isolation, but I'm also concerned about the general approach of trying to patch every part of Emacs that runs a command in order to run it via `nix-shell`. After all, once packages are installed, nix-shell is primarily running plain UNIX commands with a carefully-modified PATH (and maybe other env vars).

A general approach similar to my `exec-path-from-shell` library could ultimately allow us to set up `process-environment` locally in the buffers inside a Nix-managed project, and then everything would work without needing patches like this (or `haskell-process-wrapper-function`). And the approach would then be extensible to `stack`, `npm`, `rbenv` and other tools that work similarly.

In any case, I guess I'll merge this for now and think about the above in parallel.

sevanspowell:
@purcell Yeh I agree, doing these kind of patches isn't desirable or sustainable.

PR #1641, opened 2019-04-02
introduce `anyclass`, `stock` and `via` keywords

haitlahcen:
This PR introduces three keywords used in deriving clauses


purcell:
Thanks!

PR #1645, opened 2019-04-16
Make {-# and #-} not be paragraph boundaries

juliapath:
{- and -} are recognised as paragraph boundaries. While this makes a lot
of sense, it does not make sense to also recognise {-# and #-} as
paragraph boundaries.

Every time I want to do something at the top of the file I go to the first line, but then if I move around as usual using paragraph-wise navigation I have to go over all the language pragmas one by one, which is super annoying.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -793,8 +793,8 @@ Minor modes that work well with `haskell-mode':
     (error "haskell-mode requires at least Emacs 24.3"))
 
   ;; paragraph-{start,separate} should treat comments as paragraphs as well.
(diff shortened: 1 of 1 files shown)


TikhonJelvis:
I ran into the same problem, but never realized it was because `{-` and `-}` are paragraph markers. Great to see a fix—navigating over language pragmas has been a small but consistent annoyance for me as well.

purcell:
I'm inclined to agree, so I'll merge this and wait for signs of dissent. :-)

PR #1650, opened 2019-05-13
Add support for passing arguments to the command specified by haskell…

IvanMalison:
…-mode-stylish-haskell-path


purcell:
Great, thanks

PR #1653, opened 2019-05-28
Improve manual description to obtain SVG rendering at the GHCi REPL

fedeinthemix:
A clarification in the manual to obtain SVG rendering at the GHCi REPL.


purcell:
Thanks, sorry for the delay.

PR #1655, opened 2019-06-06
Don’t show message when minibuffer is active

matthewbauer:
This messes up minibuffer editing when active. We only want a message
to show up when there is no active minibuffer window.

/cc @purcell


purcell:
Makes sense, thanks.

PR #1656, opened 2019-06-09
Fixed incorrect comment padding

walseb:
This way of including the padding in the `comment-start` variable instead of using `comment-padding` properly is bad and breaks `outline-mode` and `outshine-mode`.  I don't know if this change breaks other packages but haven't found anything yet.


purcell:
Thanks, and sorry for the delay.

PR #1658, opened 2019-06-17
Do not obsoletely redefine byte-compile-dest-file

scfc:
Since Emacs 23.2, redefining byte-compile-dest-file has been
deprecated in favour of setting byte-compile-dest-file-function.  As
haskell-mode only supports Emacs 24.3 or later,
byte-compile-dest-file-function's value will always be honoured.


purcell:
Thanks! I just fixed this independently in a branch I plan to merge, which will restore the CI build to a reasonably working state.

Issue #1666, opened 2019-07-13
C-a in interactive-mode should stop after prompt


andreas-roehler:
C-a should stop after prompt when on first line, but stop at BOL in following lines of a multiline-edit.

Ideally it should stop at second prompt then. But not leave that  line, (was bug #1611).

purcell:
Okay, I've committed a fix: <kbd>C-a</kbd> now consistently moves point to the beginning of the line _unless_ the current prompt is on the line, in which case it moves point to just after the prompt.

PR #1667, opened 2019-07-13
#1666, C-a in interactive-mode should stop after prompt

andreas-roehler:
fixed

Bind C-a to new command ‘haskell-interactive-beginning-of-line’


andreas-roehler:
This seems solved meanwhile, closing.

PR #1670, opened 2019-08-02
Fix haskell-program-name-with-args

renatofdds:
```haskell-program-name-with-args``` uses ```nconc``` with quoted lists, which leads to infinite lists on repeated calls [1]. 

[1] [Elisp manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Rearrangement.html)

### DIFF
--- a/inf-haskell.el
+++ b/inf-haskell.el
@@ -57,19 +57,14 @@
   "Return the command with the arguments to start the repl based on the
 directory structure."
   (cl-ecase (haskell-process-type)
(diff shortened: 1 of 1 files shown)


yiyunliu:
I encountered this issue when trying to use haskell in org-babel. It would be nice to have this pull request merged.

purcell:
Thanks for this, and sorry for the slow turnaround.

renatofdds:
Thank you!

PR #1675, opened 2019-08-23
added: stack ghci fallback when `(eq 'auto haskell-process-type)`

ncaq:
I have GHC installed only by Stack.
GHC alone is not installed.
I'm afraid of conflicts because I'm using Gentoo and it takes time to compile GHC.

In that environment, `switch-to-haskell` doesn't work in the global space.
This is because `haskell-process-type` only runs `stack ghci` in a Stack project.
Actually `stack ghci` can operate in global space.
This will not break the behavior of those who want to use a single ghci for the final fallback.


purcell:
LGTM, thanks!

PR #1677, opened 2019-09-25
Handle cabal indentation relative to first line

matthewbauer:
Previously, cabal-mode indented to the ":" column of a subsection in a
cabal file. This was not how most users indented their .cabal files
and was surprising. Instead of doing this, we indent relatively to the
first entry in a subsection list. When it doesn’t exist, use the old
behavior.

:data-indent-column is added to store this new indent level. It needs
to be separate from :data-start-column due to :data-start-colum also
being used in finding the start of a subsection heading.

/cc @purcell

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -460,9 +460,13 @@ OTHER-WINDOW use `find-file-other-window'."
       (list :name (match-string-no-properties 2)
             :beginning (match-end 0)
             :end (save-match-data (haskell-cabal-subsection-end))
-            :data-start-column (save-excursion (goto-char (match-end 0))
-                                               (current-column)
-                                               )))))
+	    :data-start-column (save-excursion (goto-char (match-end 0))
+					       (current-column)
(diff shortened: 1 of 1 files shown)


haskell-cabal.el
```diff
@@ -460,9 +460,13 @@ OTHER-WINDOW use `find-file-other-window'."
       (list :name (match-string-no-properties 2)
             :beginning (match-end 0)
             :end (save-match-data (haskell-cabal-subsection-end))
-            :data-start-column (save-excursion (goto-char (match-end 0))
-                                               (current-column)
-                                               )))))
+	    :data-start-column (save-excursion (goto-char (match-end 0))
+					       (current-column)
+                                               )
```
purcell: This guy is lonely.

purcell:
Great, thanks!

purcell:
I'm working on fixing up the CI build, btw - I realise it's not very helpful currently.

PR #1678, opened 2019-10-14
Enable haskell-cabal-mode in cabal-like files

cydparser:
This change adds syntax highlighting to ~/.cabal/config, cabal.project, cabal.project.local, and cabal.project.freeze.


purcell:
Thanks, merged with a minor fix to one of the regexps. :-)

PR #1680, opened 2019-10-26
Interactive mode backward kill word

r-zip:
This PR implements a fix to `backward-kill-word` in interactive-haskell-mode.

The way things are currently implemented, calling `backward-kill-word` in the REPL causes the point to move to the beginning of the line if there is nothing else to delete on the line. The user then gets a message that the prompt text is read-only.

This is annoying for two reasons:
1. The message about read-only prompt text clutters the minibuffer, and
2. The user has to move the point past the prompt again to continue typing.

The changes in this PR improve the ergonomics of interactive-haskell-mode by detecting when `backward-kill-word` would move past the prompt. In that case, the point is just moved to right after the prompt.

I've chosen to re-bind <kbd>M-\<backspace\></kbd> and <kbd>C-\<backspace\></kbd> in the `haskell-interactive-mode-map`, and from what I can tell, it works.

I've never contributed a PR to open-source, but I've read the manual. Please let me know if I need to add tests or anything else.

Old behavior:
https://asciinema.org/a/OYRtYHCAUSNZjWq2NoCn4UEdb

New behavior:
https://asciinema.org/a/fm3Gd6WV1Q7IMWMzAxwcNQTCV


purcell:
I feel like there must be a more general way to tackle this, since many other commands exist which would need to be overridden in this way, e.g. `backward-kill-sexp`, `kill-back-to-indentation` etc.

Issue #1691, opened 2019-12-19
wrong-number-of-arguments (3 . 3), 2


Vonfry:
It starts from commits 4d77a88a2cb14ff6b9672c1503c46086300f94e7.

Some details: [travis-ci](https://travis-ci.org/haskell/haskell-mode/jobs/626987871#L375-L376)

cc: @purcell

guygastineau:
I am experiencing this tonight as well.

guygastineau:
This is line 139 of `haskell-interactive-mode.el`:
``` elisp
(define-obsolete-face-alias 'haskell-interactive-face-prompt2 'haskell-interactive-face-prompt-cont)
```
Unfortunately, the macro `define-obsolete-face-alias` requires 3 arguments.
``` elisp
(defmacro define-obsolete-face-alias (obsolete-face current-face when)
  "Make OBSOLETE-FACE a face alias for CURRENT-FACE and mark it obsolete.                                             
If provided, WHEN should be a string indicating when the face                                                         
was first made obsolete, for example a date or a release number."
  `(progn
     (put ,obsolete-face 'face-alias ,current-face)
     ;; Used by M-x describe-face.                                                                                    
     (put ,obsolete-face 'obsolete-face (or (purecopy ,when) t))))
```

qcfu-bu:
I am having the same issue. From the way the comment in the macro is worded, it should be able to take either 2 or 3 arguments?

guygastineau:
@qcfu-bu thank you for that clarification.

I think I didn't really read the docstrings, because I was already thinking about what the debug messages told me.

I wonder why this is happening then. I am not used to elisp macros (I'm used to scheme macros). Do elisp macros need to state optional args as optional?

qcfu-bu:
I believe there should be an optional argument, though I may be wrong.

[https://www.gnu.org/software/emacs/manual/html_node/eintr/Optional-Arguments.html](url)

guygastineau:
Yeah, reading some documentation it seems like `&optional` can be used with `defmacro`, so perhaps that is the problem here.

I can't actually tell where `define-obsolete-face-alias` is defined (I got the source from the debugger, but it came from a byte code run and it didn't give me the source location).  If it is from another package, then maybe it needs to be fixed there.

I excluded the comment that preceded it in the source. That comment basically said, "the following macro doesn't do anything, and don't use it."

So I don't really know where this leads us, but maybe that information is helpful to someone.

[...]

PR #1692, opened 2019-12-19
Fix for #1691: disable-obsolete-face-alias is defined to require three args

jfb-ampli:
... so we need to supply a non-sense third argument here.

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -136,7 +136,7 @@ be nil.")
   :group 'haskell-interactive)
 
 ;;;###autoload
-(define-obsolete-face-alias 'haskell-interactive-face-prompt2 'haskell-interactive-face-prompt-cont)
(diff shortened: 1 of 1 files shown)


qcfu-bu:
I want to point out there's another instance of disable-obsolete-face-alias in haskell-customize.el

jfb-ampli:
> I want to point out there's another instance of disable-obsolete-face-alias in haskell-customize.el. I changed both instances of mine.

Is there? On head, I see `define-obsolete-variable-alias`, which is a different macro (and I checked, it has the `&optional` marker)

jfb-ampli:
`(defmacro define-obsolete-variable-alias (obsolete-name current-name
						 &optional when docstring)`

qcfu-bu:
My bad, you are right. The one in customize is correct.

purcell:
fixed separately in 38c512b, thanks.

PR #1694, opened 2020-01-13
Fix haskell-compile.el

pranaysashank:
It is currently checking for the prefix to be a negative integer, which throws a compiler error because we are checking for it to be a `'-` in other places. Looks like this error is introduced in b8892291c0e44f964dc1634351e62b1b242d55f7 commit. This PR reverts it to the earlier behaviour.


purcell:
Thanks, good catch. Sorry for the inconvenience.

PR #1698, opened 2020-02-21
Enable fontification for qualified quasiquoters

alexpeits:
This enables font-locking when the quasiquoter name is imported qualified, e.g.:

```
[SomeModule.sql|SELECT * from table|]
```

### DIFF
--- a/tests/haskell-lexeme-tests.el
+++ b/tests/haskell-lexeme-tests.el
@@ -292,6 +292,26 @@ buffer."
    "[xml| <xml />"
    '("[xml| <xml />")))
 
+(ert-deftest haskell-lexeme-quasi-quote-qual-1 ()
+  (check-lexemes
+   '("[Mod.xml| <xml /> |]")
+   '("[Mod.xml| <xml /> |]")))
+
+(ert-deftest haskell-lexeme-quasi-quote-qual-2 ()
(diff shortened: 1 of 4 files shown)


tests/haskell-lexeme-tests.el
```diff
@@ -292,6 +292,26 @@ buffer."
    "[xml| <xml />"
    '("[xml| <xml />")))
 
+(ert-deftest haskell-lexeme-quasi-quote-qual-1 ()
```
alexpeits: These tests pass even without my fix for some reason, not 100% sure if it makes sense to have them here

purcell:
LGTM, thanks!

purcell:
FWIW, I found that font-locking lots of quasi-quoted SQL made `haskell-mode` pretty sluggish.

alexpeits:
It does, probably because it uses a temp buffer and loads the relevant mode. I normally just make the quasiquote mapping alist `nil` so everything gets fontified as a string (but this didn't happen for qualified strings, and the contents were fontified as normal haskell code)

Thanks for the quick response!

purcell:
I end up just setting `haskell-font-lock-quasi-quote-modes` to `nil` in `.dir-locals.el` for some projects.

PR #1715, opened 2020-08-22
Fall back to 'haskell-mode-tag-find' when 'haskell-mode-find-def' fails in 'haskell-mode-jump-to-def-or-tag'

### LINKED ISSUE #351: haskell-mode-jump-to-def-or-tag sometimes doesn't jump
Sometimes when I'm editing haskell code and I hit `M-.`, which is bound to `haskell-mode-jump-to-def-or-tag`, instead of jumping to the location where that tag is defined, it just puts a message in my minibuffer "Defined in `Some.Module`.

Strangely enough, if I manually run `find-tag` with the same tag, it jumps to the file as expected. I wish I had a better repro case, I just notice it happen several times a day at work. Let me know if there's any more information I could provide that would help.

TOTBWF:
This PR changes the behavior of `haskell-mode-jump-to-def-or-tag` to fall back on `haskell-mode-tag-find`when `haskell-mode-find-def` fails. This fixes #351.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -376,7 +376,7 @@ If the definition or tag is found, the location from which you jumped
 will be pushed onto `xref--marker-ring', so you can return to that
 position with `xref-pop-marker-stack'."
   (interactive "P")
-  (if-let ((_ (haskell-session-maybe))
+  (if-let ((session (haskell-session-maybe))
            (initial-loc (point-marker))
            (loc (haskell-mode-find-def (haskell-ident-at-point))))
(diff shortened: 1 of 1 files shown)


purcell:
Sounds good, thanks, but there's a byte-compilation issue which is failing CI - could you take a look please?

purcell:
I fixed the issue that was breaking CI on master, so I'll assume everything is fine here and merge it, rather than figure out how to rebase this branch first. Thanks!

PR #1719, opened 2020-09-19
Make Hoogle work in Nix shell

### LINKED ISSUE #1718: Hoogle not called inside direnv environment
I use a Direnv and a Nix shell to provide my Hoogle (and `envrc.el` to integrate with Emacs). However, for reasons I can't figure out, `shell-command-to-string`, which `haskell-hoogle` [uses](https://github.com/haskell/haskell-mode/blob/master/haskell-hoogle.el#L84), does not run its commands inside the Nix shell, even though `shell-command` does.

galagora:
Fixes #1718.

Also, Hayoo should be removed, since it seems dead.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -192,29 +192,6 @@ MODULE-BUFFER is the actual Emacs buffer of the module being loaded."
 (defvar haskell-cabal-targets-history nil
   "History list for session targets.")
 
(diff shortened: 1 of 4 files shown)


purcell:
Do you want to just go ahead and remove the Hayoo references as part of this?

galagora:
Sure

purcell:
Super, thank you!!

PR #1721, opened 2020-09-25
Restore `haskell-hoogle-command` configurability

### LINKED ISSUE #1720: Hoogle results missing URL
Local `hoogle` query results were displaying URL, but that disappeared in one of last commits.

For example, before it was like this:
`Prelude class Eq a => Ord a -- https://hackage.haskell.org/package/base/docs/Prelude.html#t:Ord`

but now it is displayed as
`Prelude class Eq a => Ord a`

Url was very convenient to open a web page directly from `emacs`.

Was that removed on accident or is there some new setting that controls this behaviour or a new functionality that would allow opening `hoogle` result in browser?

galagora:
Fixes #1720.

### DIFF
--- a/haskell-hoogle.el
+++ b/haskell-hoogle.el
@@ -40,24 +40,15 @@
                        nil nil def)
           )))
 
-;;;###autoload
-(defun haskell-hoogle (query &optional info)
-  "Do a Hoogle search for QUERY.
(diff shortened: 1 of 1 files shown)


purcell:
Thanks!

PR #1722, opened 2020-09-29
Don't message "Tags generated" if tags command exited non-zero

unhammer:


haskell-mode.el
```diff
@@ -1205,8 +1205,9 @@ generated."
          (command (haskell-cabal--compose-hasktags-command dir)))
     (if (not command)
         (error "Unable to compose hasktags command")
-      (shell-command command)
-      (haskell-mode-message-line "Tags generated.")
+      (when (eq 0 (shell-command command))
+        ;; don't message if it fails – don't want to  overwrite shell-command's message
```
purcell, line 1209:
```suggestion
      (when (zerop (shell-command command))
```

purcell (approved):

purcell:
Thanks!

PR #1733, opened 2020-12-27
Always use buffer-file-name for ghc compilation

jacobilsoe:
Seems ghc compilation was broken recently. Invoking haskell-compile in a buffer with a single haskell source file calls ghc with the directory and not the filename.


purcell:
LGTM, thanks.

PR #1737, opened 2021-03-03
Add package prefix to `haskell-compilation-error-regexp-alist`

Jasagredo:
## Problem

When compiling a package with Stack that declares multiple packages on its `stack.yaml` file, the output of each package compilation is prepended with the string `package-name\s*>\s` like this:

```
testing> /home/javier/testing/src/Lib.hs:7:1: warning: [-Wmissing-signatures]
testing>     Top-level binding with no type signature: someFunc :: IO ()
```

Without this change, the string `testing> ` would be matched in the first part of the regexp and then emacs would fail to find the file as it would be searching for:
```
testing> /home/javier/testing/src/Lib.hs
```
as filename. The string before the `>` is also padded to the longest package name in the compilation.

## Proposed solution
The change modifies slightly the regexp to match and filter out this prefix.

## Before the change
Padded package:
image
Un-padded package:
image

## After the change:
Padded package:
image
Un-padded package:
image

Errors on simple stack packages still work as expected:
image
Errors on Cabal compilations work as expected:
image
Errors on plain GHC work as expected:
image

### DIFF
--- a/haskell-compile.el
+++ b/haskell-compile.el
@@ -92,7 +92,8 @@ The `%s' placeholder is replaced by the current buffer's filename."
 
 (defconst haskell-compilation-error-regexp-alist
   `((,(concat
-       "^ *\\(?1:[^\t\r\n]+?\\):"
+       "^ *\\([^\n\r\t>]*\s*> \\)?" ;; if using multi-package stack project, remove the package name that is prepended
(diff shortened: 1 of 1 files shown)


purcell:
LGTM, thanks!

PR #1739, opened 2021-03-25
Make sure that a local file name is used in compile commands

### LINKED ISSUE #1559: Support tramp
When you open project via tramp, if you launch `haskell-compile` you get something like:

```
cd /ssh:ubuntu@builder:/home/ubuntu/workspace/proj/ && stack build
/bin/sh: 2: cd: can't cd to /ssh:ubuntu@builder:/home/ubuntu/workspace/proj/

HsCompilation exited abnormally with code 2 at Tue Oct  3 13:13:07

```

Would be very handy if `haskell-compile` would work under tramp. Thank you!

juergenhoetzel:
This prevents passing TRAMP file names as arguments to external compile
commands.

Fixes #1559


purcell:
LGTM, thanks!

PR #1740, opened 2021-04-07
Use "!" class for literate comments instead of "<" (fixes #1594)

### LINKED ISSUE #1594: fill-paragraph is completely broken in literate-haskell-mode
Consider the following text in `test.lhs`:

```
Every expression in Haskell has a type which is determined at compile
time. All the types composed together by function application have to
match up. If they don't, the program will be rejected by the
compiler. Types become not only a form of guarantee, but a language
for expressing the construction of programs.
```

Running `emacs -Q`, I load the latest revision of `haskell-mode` (d71cca7f5ddc8178f5fe331563edcefe91cc0bcb) and press `M-q` with point on the second line of the file. Result:

```
Every expression in Haskell has a type which is determined at compile
 time. All the types composed together by function application have to
 match up. If they don't, the program will be rejected by the
 compiler. Types become not only a form of guarantee, but a language
 for expressing the construction of programs.
```

Pressing `M-q` multiple times gets me to something like this:

```
Every expression in Haskell has a type which is determined at compile
             time. All the types composed together by function
             application have to match up. If they don't, the program
             will be rejected by the compiler. Types become not only a
             form of guarantee, but a language for expressing the
             construction of programs.
```

What??

---

Emacs version: 26.1-rc1 installed from Homebrew
Operating system: macOS 10.11.6

goose121:
The `!` syntax class is intended for making arbitrary pieces of text into comments; thus, tools such as `fill-paragraph` and `auto-fill-mode` treat it differently than `<`, which is intended for specifying actual comment characters.

### DIFF
--- a/haskell-font-lock.el
+++ b/haskell-font-lock.el
@@ -586,10 +586,10 @@ on an uppercase identifier."
               (goto-char (1+ (point)))))))
       ;; must return nil here so that it is not fontified again as string
       nil))
-   ;; Detect literate comment lines starting with syntax class '<'
+   ;; Detect literate comment lines starting with syntax class '!'
    ((save-excursion
       (goto-char (nth 8 state))
-      (equal (string-to-syntax "<") (syntax-after (point))))
+      (equal (string-to-syntax "!") (syntax-after (point))))
     'haskell-literate-comment-face)
    ;; Detect pragmas. A pragma is enclosed in special comment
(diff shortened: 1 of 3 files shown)


purcell:
Thanks for this. I think that build failure in Emacs 27.1 is unrelated: have seen it a few times before. Will go ahead and merge.

PR #1745, opened 2021-04-24
Run hooks for haskell-error-mode

erickgnavar:
When `haskell-interactive-popup-error` is called, `haskell-error-mode`
hooks aren't running because `run-hooks` call is missed

### DIFF
--- a/haskell-interactive-mode.el
+++ b/haskell-interactive-mode.el
@@ -652,6 +652,7 @@ wrapped in compiler directive at the top of FILE."
         (with-current-buffer buf
 
           (haskell-error-mode)
+          (run-hooks 'haskell-error-mode-hook)
           (let ((inhibit-read-only t))
             (erase-buffer)
             (insert (propertize response


purcell:
Isn't this because the hook variable name produced by the `define-minor-mode` form is `haskell-error-mode-hook`, without a trailing `s`?

erickgnavar:
> Isn't this because the hook variable name produced by the `define-minor-mode` form is `haskell-error-mode-hook`, without a trailing `s`?

The code I'm adding is without the trailing `s` as you can see

```lisp
(run-hooks 'haskell-error-mode-hook)
```

Maybe there was a confusion with the PR's description 🤔

purcell:
Oh, sorry if I got confused. But in any case, the expanded definition of `haskell-error-mode` includes code to run that hook, so I'd like to understand why that didn't seem to be working for you.

erickgnavar:
I forgot I have this PR open, I haven't used `haskell-mode` in a while so I'm closing this

PR #1748, opened 2021-05-06
Update supported extensions and options to GHC 9.0.1

cydparser:
I updated the variables using GHC 9.0.1 and `emacs --batch --file haskell-ghc-support.el --eval "(progn (eval-buffer) (haskell-update-ghc-support \"ghc\") (save-buffer))"`.


purcell:
Nice, thanks for this!

PR #1749, opened 2021-05-07
Add font-lock support for ImportQualifiedPost

mrBliss:
Now the `qualified` keyword in

```haskell
import X qualified as Y
```

is correctly highlighted.


purcell:
Nice, thank you!

Issue #1751, opened 2021-06-19
Degrading scroll performance in terminal mode


dminuoso:
# The Problem #

Since I switched to emacs I noticed how scrolling sporadically became worse. At times, it would take 3-4s to redraw 40 lines of code.

# Apparent relevant factors #

After doing a lot of analyzing I narrowed it down to the following combination of things:

1) Emacs must run in terminal, in GUI either the problem is not as apparent or its non-existant.
2) gc-cons-threshold must be set to a high value. LSP users likely have something like `(setq gc-cons-threshold 100000000)`
3) `haskell-mode` and `font-lock-mode` must be enabled
4) Do some text editing

Watch drawing lines become slower. The presence of constructor names seems especially bad. Note, that the entire line is drawn much slower, so if you vertically tile your emacs to have some unrelated buffer on the right hand side, then the entire line with many data constructors on the left hand window will be drawn slowly.

5) If you then cause a `(garbage-collect)`, scrolling becomes fast again.

# Reproduce #

Locally I used emacs 27.1 installed through the emacs nix overlay:


<details>
<summary>Click to see/use my nixos module</summary>
<p>

```nix
{ config, pkgs, ... }:

let
  emacsOverlay = import (builtins.fetchTarball {
  # Descriptive name to make the store path easier to identify
    name = "emacs-overlay-2020-09-11";
    # Commit hash for nixos-unstable as of 2018-09-12
    url = https://github.com/nix-community/emacs-overlay/archive/a69588a3f7de6d68f20cea21562ab7f6f91a400a.tar.gz;
    # Hash obtained using `nix-prefetch-url --unpack <url>`
    sha256 = "180jp31zfrlhnz490x6hckz6z3bvfxwz1613cg4zmf0lrflrl3gp";
  });

  myEmacs = pkgs.emacs;
  emacsWithPackages = (pkgs.emacsPackagesGen myEmacs).emacsWithPackages;
  emacs = emacsWithPackages (epkgs: (with epkgs.melpaStablePackages; [
  # melpa stable
    ripgrep
    magit
    # dockerfile-mode
    # markdown-mode
    # yaml-mode
    # json-mode
    # go-mode
    # nix-mode
    use-package
    projectile
    # direnv
    # swiper
    # ox-gfm
    # pdf-tools
  ]) ++ (with epkgs.melpaPackages; [
  # melpa unstable
    solarized-theme
    gruvbox-theme
    company
    counsel
    counsel-projectile
    ivy
[...]

PR #1760, opened 2021-08-16
Reduce (over)use of markers (#1751)

mattiase:
The parser used by the font-lock code created a massive amount of
markers; try to reduce it as far as possible since they make some
operations expensive as long as they are in the buffer. Even orphaned
markers stay until next GC.

Read match data as integers instead of markers, and don't create new
markers when setting the match data. This should work just fine as
long as the buffer isn't changed during the match data lifetime, which
appears reasonable.


purcell:
Makes sense to me so will merge. I'm taking a short time to catch up on some PRs like this which have been languishing, thanks for your patience.

mattiase:
You are very diligent indeed, thank you!

PR #1761, opened 2021-09-04
Fix compilation command to handle whitespace in file path

### LINKED ISSUE #1673: haskell-compilation error when file path has spaces in Windows
The error message follows:
```
-*- mode: haskell-compilation; default-directory: "g:/My Drive/haskell/" -*-
HsCompilation started at Tue Aug 13 20:36:38

ghc -Wall -ferror-spans -fforce-recomp -c g:/My Drive/haskell/hello.hs
<command line>: does not exist: G:\My

HsCompilation exited abnormally with code 1 at Tue Aug 13 20:36:39
```

EuanParr:
Enclose the selected file's path in quotes when compiling a stand-alone Haskell file to prevent paths with spaces being parsed as multiple arguments.

Fixes #1673.

### DIFF
--- a/haskell-compile.el
+++ b/haskell-compile.el
@@ -69,7 +69,7 @@ For legacy compat, `%s' is replaced by the stack package top folder."
   :type 'string)
 
 (defcustom haskell-compile-command
-  "ghc -Wall -ferror-spans -fforce-recomp -c \"%s\""
+  "ghc -Wall -ferror-spans -fforce-recomp -c %s"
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. Even this isn't really safe, though, because the path could include quotes. It would likely be safer to (elsewhere) replace `%s` with `(shell-quote-argument file-path)` instead of `file-path`.

EuanParr:
Thanks for the advice. This new commit is hopefully an improvement.

purcell:
Great, thank you!

PR #1771, opened 2022-02-06
Add `capi` calling convention

amesgen:
See e.g. [this blogpost](https://www.haskell.org/ghc/blog/20210709-capi-usage.html) for what `capi` does. A quick example seems to indicate that `capi` is not supported in `foreign export` (which I think makes sense).


purcell:
Oh, hi @amesgen 😁

PR #1773, opened 2022-03-12
Remove autoloads from face and group definitions

minad:
Generally only functions and modes should be autoloaded. add-to-list can also be
used cautiously to register modes etc. Faces are usually not autoloaded. To make
faces, groups and custom variables available for customization, a package must
be loaded.


minad:
See also #1048

purcell:
Yep, makes sense, thanks!

PR #1774, opened 2022-03-25
Add interactive command to insert a LANGUAGE pragma

TikhonJelvis:
I need to manually add `{-# LANUGAGE _ #-}` pragmas to the top of my modules pretty often, so I wrote a function to do this. I've been using it for a few days and found it really convenient, and other people liked the idea when I mentioned it on Twitter.

Called interactively it uses `haskell-ghc-supported-extensions` for completions which means less typing and no more errors like using `RecordWildcards` instead of `RecordWildCards`.

I bound it to `C-c C-e` on my own setup, but not sure whether it makes sense to include a keybinding for it by default.

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -936,13 +936,13 @@ newlines and extra whitespace in signature before insertion."
         (insert sig "\n")
         (indent-to col)))))
 
-(defun haskell-mode-add-language-pragma (extension)
+(defun haskell-command-insert-language-pragma (extension)
   "Insert a {-# LANGUAGE _ #-} pragma at the top of the current
 buffer for the given extension."
   (interactive
    (list (completing-read "Extension: " haskell-ghc-supported-extensions)))
   (save-excursion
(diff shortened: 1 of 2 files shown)


purcell (changes requested):
Looks good, thanks! Noted a couple of minor points that should be fixed up please.

TikhonJelvis:
Looks like CI is failing with some kind of Nix issue now. Would it be worth pinning the version of Nixpkgs that you're using here?

purcell:
> Out of curiosity, what's the difference between (beginning-of-buffer) and (goto-char (point-min))

You'd have to look at the source (`M-.` will jump to definition) but from memory, the former pushes the mark first.

purcell:
> Looks like CI is failing with some kind of Nix issue now. Would it be worth pinning the version of Nixpkgs that you're using here?

Possibly... this needs a bit of investigation. But I'll go ahead and merge the changes in the meantime, thanks.

PR #1775, opened 2022-03-28
fix typos

askeblad:
haskell-c2hs.el
haskell-completions.el
haskell-decl-scan.el
haskell-font-lock.el
haskell-indent.el
haskell-indentation.el
haskell-mode.el
haskell-utils.el
tests/haskell-font-lock-tests.el

### DIFF
--- a/haskell-c2hs.el
+++ b/haskell-c2hs.el
@@ -111,7 +111,7 @@
                        (opt (+ ,ws)
                             (group-n 3
                                      "as")))
-                  ;; TODO: fun hook highlighting is incompelete
+                  ;; TODO: fun hook highlighting is incomplete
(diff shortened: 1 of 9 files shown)


askeblad:
@purcell

grammar looks off in this comment

```elisp
;; For major use function `haskell-completions-grab-prefix' is supposed, and
;; other prefix grabbing functions are used internally by it.  So, only this
;; function have prefix minimal length functionality and invokes predicate
;; function `haskell-completions-can-grab-prefix'.
```

askeblad:
guess the intention was `reset`


> tests/haskell-utils-tests.el
>
>```
>    ;; set some random value to flag to test that it will be reseted
>```

PR #1780, opened 2022-05-18
Fixed build errors.

### LINKED ISSUE #1779: Build fails on Emacs 28.1
```
$ make
EMACS check-emacs-version
Using EMACS = emacs, version = 28.1
mkdir build-28.1
EMACS build-28.1/ghc-core.elc
EMACS build-28.1/ghci-script-mode.elc
EMACS build-28.1/haskell-align-imports.elc
EMACS build-28.1/haskell-c2hs.elc
EMACS build-28.1/haskell-cabal.elc

In toplevel form:
haskell-cabal.el:362:8: Error:  docstring wider than 80 characters
make: *** [Makefile:80: build-28.1/haskell-cabal.elc] エラー 1
```

taiju:
rel: #1779 

Hello,

When building with Emacs 28.1, there were several places where an error occurred because the docstring had more than 80 characters.
I have adjusted the number of characters per line in the docstring where the error occurs.

If there are line breaks in the wrong place, please feel free to correct it.

Thank you.

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -360,7 +360,8 @@ OTHER-WINDOW use `find-file-other-window'."
 (defconst haskell-cabal-conditional-regexp "^[ \t]*\\(\\if\\|else\\|}\\)")
 
 (defun haskell-cabal-classify-line ()
-  "Classify the current line into 'section-header 'subsection-header 'section-data 'comment and 'empty '"
+  "Classify the current line into 'section-header 'subsection-header
+'section-data 'comment and 'empty '"
   (save-excursion
     (beginning-of-line)
     (cond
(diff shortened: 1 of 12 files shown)


purcell:
This is great, thanks. What really needs to be done here, though, is to keep `M-x checkdoc` happy with the reformatted docstrings. Generally that means keeping the first line as a full (but shorter) sentence, and then often moving other info from the docstring into subsequent lines of the docstring, e.g.
```
  "Enumerate .cabal targets. PROCESS-TYPE determines the format of the returned target."
```
to
```
  "Enumerate .cabal targets.
PROCESS-TYPE determines the format of the returned target."
```

or
```
  "Classify the current line into 'section-header 'subsection-header 'section-data 'comment and 'empty '"
```
to
```
  "Classify the current line's type.
Possible results are 'section-header, 'subsection-header,
'section-data, 'comment and 'empty."
```

So it's a slightly bigger job overall. Do you think you might be able to help with it? I haven't found time.

taiju:
I will try, but please don't expect much.
Because I am not familiar with Haskell just a learner, besides, I am not good at English 😅

taiju:
@purcell  I fixed it so that the first line is a complete sentence.

The checkdoc command also points out other things, but I'm sorry ,I did not fix those. I'm only the new points raised by the change for line breaking.

Commits are separated, but please 'squash' or 'fixup' when merging.

purcell:
Merged, thanks so much! 🙏

PR #1784, opened 2022-09-14
Correct parenthese matching error

VitalyAnkh:
The emacs upstream will raise an error now when `otherwise` or `t` clause is not the last clause in a `cl-case`. There is a such `cl-case` in `haskell-commands.el` which is apparently a typo and should be fixed.

Reference: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=6d8f5161ead689b7a2e44a7de0a695f0ab4c833b

### DIFF
--- a/haskell-commands.el
+++ b/haskell-commands.el
@@ -677,9 +677,9 @@ happened since function invocation)."
                             (cdr (reverse haskell-utils-async-post-command-flag))))
                ;; Present the result only when response is valid and not asked
                ;; to insert result
-               (haskell-command-echo-or-present response)))
+               (haskell-command-echo-or-present response))))
 
-            (haskell-utils-async-stop-watching-changes init-buffer))))))))
+          (haskell-utils-async-stop-watching-changes init-buffer)))))))
 
 (make-obsolete 'haskell-process-generate-tags
                'haskell-mode-generate-tags


Jasagredo:
I'm curious about why you ended up in this final commit. If I go through your series of pushes, it seems that at the beginning you did the same as I did in #1785. which is basically going from this:
```
(let* (...)
  (cl-case 
    ...
    (otherwise
      (if ...
        ...
        (haskell-command-echo-or-present response)))
    (haskell-utils-async-stop-watching-changes init-buffer)))
```
to this:
```
(let* (...)
  (cl-case 
    ...
    (otherwise
      (if ...
        ...
        (haskell-command-echo-or-present response))))
  (haskell-utils-async-stop-watching-changes init-buffer))
```

Instead now your PR results in this:
```
(let* (...)
  (cl-case 
    ...
    (otherwise
      (if ...
        ...
        (haskell-command-echo-or-present response))
      (haskell-utils-async-stop-watching-changes init-buffer))))
```

Is there any particular reason?

VitalyAnkh:
@Jasagredo Hi! I think you are right. The `haskell-utils-async-stop-watching-changes` should be in the same hierarchy as the `cl-case`. The code is updated now. Thanks!

purcell:
Merged, thanks

odanoburu:
It seems this is not released yet… can such a release be made? I just found and corrected this bug, only to find out it was already fixed :P

PR #1785, opened 2022-09-15
Fix misplaced parenthesis

Jasagredo:
I `doom/upgrade`d my Emacs today, and it started complaining about:

```
Misplaced t or ‘otherwise’ clause
```

It seems one of the parentheses was misplaced, creating a case branch after `otherwise` which should not be there.


Jasagredo:
Closing as a duplicate of 1784

PR #1786, opened 2022-09-17
Hoogle: colorize hoogle cli output using haskell-mode font-lock

jaor:
It's been a long time since hoogle command line interface lost the ability to
output ANSI colors (see https://github.com/ndmitchell/hoogle/issues/133).

This patch makes haskell-hoogle to use emacs font lock to fontify its output,
with a custom flag to opt out.

### DIFF
--- a/haskell-hoogle.el
+++ b/haskell-hoogle.el
@@ -27,6 +27,7 @@
 ;;; Code:
 
 (require 'ansi-color)
+(require 'view)
 (require 'haskell-mode)
 (require 'haskell-utils)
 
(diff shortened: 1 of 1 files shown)


purcell:
This makes sense to me, but if it actually works then I'd suggest we do without the flag. Or does it not work so well? My rationale here is: the fewer knobs and switches, the better.

jaor:
On Fri, Mar 03 2023, Steve Purcell wrote:

> This makes sense to me, but if it actually works then I'd suggest we
> do without the flag. Or does it not work so well? 

it's worked fine for my needs at least, the only reason for the flag was
to not surprise previous users.

> My rationale here is: the fewer knobs and switches, the better.

yes, i think (at least generally speaking) i agree with that.

purcell:
Okay, I removed the defcustom and merged the changes — thanks!

PR #1790, opened 2022-11-04
Remove needless quotes from keylist of case macro

syohex:
Development Emacs(29.0.50) generates following byte-compile warnings. This PR fixes warnings like below ones.

```
haskell-cabal.el:704:12: Warning: Case 'before will match ‘quote’.  If that’s
    intended, write (before quote) instead.  Otherwise, don’t quote ‘before’.
haskell-cabal.el:704:12: Warning: Case 'after will match ‘quote’.  If that’s
    intended, write (after quote) instead.  Otherwise, don’t quote ‘after’.
haskell-cabal.el:704:12: Warning: Case 'single will match ‘quote’.  If that’s
    intended, write (single quote) instead.  Otherwise, don’t quote ‘single’.
```

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -702,7 +702,7 @@ Respect the COMMA-STYLE, see
 `haskell-cabal-strip-list-and-detect-style' for the possible
 styles."
   (cl-case comma-style
-    ('before
+    (before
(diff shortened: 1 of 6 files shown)


purcell:
Super, thank you!

PR #1792, opened 2022-11-06
Escape single quote in docstring

syohex:
Development Emacs(version 29) warns not escaped quotes in docstring as below. This PR fixes this issue and makes `make` command success on development Emacs too.

```
In haskell-cabal-classify-line:
haskell-cabal.el:363:2: Warning: docstring has wrong usage of unescaped single
    quotes (use \= or different quoting)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,7 @@ $(AUTOLOADS): $(ELFILES)
 	$(BATCH) \
 		--eval '(setq make-backup-files nil)' \
 		--eval "(setq generated-autoload-file (concat command-line-default-directory \"/\" \"$@\"))" \
+		--eval "(require 'autoload)" \
 		-f batch-update-autoloads "."
 # check if autoloads will really load
 	$(BATCH) -l "$@"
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -150,6 +150,7 @@ $(AUTOLOADS): $(ELFILES)
 	$(BATCH) \
 		--eval '(setq make-backup-files nil)' \
 		--eval "(setq generated-autoload-file (concat command-line-default-directory \"/\" \"$@\"))" \
+		--eval "(require 'autoload)" \
```
syohex, line 153: `autoload.el` will be obsoleted since Emacs 29. So it is necessary to load explicitly on latest Emacs.

haskell-cabal.el
```diff
@@ -362,8 +362,8 @@ OTHER-WINDOW use `find-file-other-window'."
 
 (defun haskell-cabal-classify-line ()
   "Classify the current line's type.
-Possible results are 'section-header 'subsection-header 'section-data 'comment
-and 'empty '"
```
syohex, line 366: Remove needless quote

purcell:
Thank you!

PR #1793, opened 2022-11-10
Define haskell-mode-hook before haskell-mode

### LINKED ISSUE #1480: haskell-mode-hook customizations are lost
Customizations to the variable `haskell-mode-hook` are lost (reset to nil) after haskell-mode is first loaded.

1. Add, e.g., `'(haskell-mode-hook '(haskell-indent-mode interactive-haskell-mode intero-mode))` to `custom-set-variables`.
2. Restart Emacs.
3. `(find-file "something.hs")`
4. `(describe-variable 'haskell-mode-hook)` and its value will be nil.

blucoat:
If `define-derived-mode haskell-mode` comes before `defcustom haskell-mode-hook`, custom will report `haskell-mode-hook` as being changed outside of customize.  This prevents customizations from taking effect.  Swapping the order of definitions fixes this.

Fixes #1480.  Note that what @phil-s mentioned in that issue about clobbering the docstring is not a problem since Emacs 26.1, thanks to [this commit](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=3e7ebbe1bd5d33476190c789d17e4dd2d5f1bdfb), so I didn't bother with the workaround he suggested.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -743,6 +743,26 @@ Prefix ARG is handled as per `delete-indentation'."
       (funcall #'electric-pair-default-inhibit ch)))
 
 ;; The main mode functions
+(defcustom haskell-mode-hook '(haskell-indentation-mode interactive-haskell-mode)
+  "List of functions to run after `haskell-mode' is enabled.
+
+Use to enable minor modes coming with `haskell-mode' or run an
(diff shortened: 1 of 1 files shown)


purcell:
Thanks!

PR #1797, opened 2022-11-24
add optional single-line-p in haskell-process-show-repl-response

Inc0n:
With `SINGLE-LINE-P` as nil, it echos the response as is.  This allows the response to remain multi lined, which is better for error reponses.


purcell:
This isn't an interactive command, so it's not clear to me who would actually call this function with that new additional parameter. What am I missing?

purcell:
Great, thanks for confirming.

PR #1799, opened 2023-01-19
Fix permanent haskell-doc mode.

ErnestKz:
The `eldoc-documentation-function` is turned on regardless if `haskell-doc-mode` is on or off, and it is hard to find a way to disable it.

`haskell-doc-mode` still works as intended, and actually turns off the eldoc popup when the mode is disabled.

### DIFF
--- a/haskell-mode.el
+++ b/haskell-mode.el
@@ -832,8 +832,6 @@ Minor modes that work well with `haskell-mode':
   (setq-local parse-sexp-ignore-comments nil)
   (setq-local syntax-propertize-function #'haskell-syntax-propertize)
 
-  ;; Set things up for eldoc-mode.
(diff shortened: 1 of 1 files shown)


purcell:
Thanks @ErnestKz! `haskell-doc-mode` was a bit overengineered so I've put together an alternative approach in #1802 — would you mind taking a look?

purcell:
Replaced by #1802

ErnestKz:
Thank you for the tidy up. I had a look, but didn't have enough of an understanding to give an informed opinion. So I planned to have a deeper look later.

Though I'm sure you know what you are doing, so I'm happy that you merged your clean up sooner than later.

purcell:
Super, thanks for the feedback, and prompting these changes!

PR #1800, opened 2023-02-02
change lib-target in haskell-cabal-enum-targets to support multiple library components.

### LINKED ISSUE #1796: haskell-session-change-target unusable in cabal package with multiple library stanzas
I have a project in which my cabal file specifies multiple libraries (since I'm using backpack) [1].  If I try to fire up an repl I have to select the correct target using haskell-session-change-target. That works well for selecting benchmarks or testsuites,  however, 
for each library stanza I just get a copy of "lib:<package-name>"  (so "lib:hgeometry-vector" in my particular case), and unfortunately that does not allow me to actually switch to the right library.

Browsing through the source a bit, I think the culprit is the [haskell-cabal-enum-targets function](https://github.com/haskell/haskell-mode/blob/a34ccdc54be15043ff0d253c3c20087524255491/haskell-cabal.el#L498) in haskell-cabal.el. In particular, it seems that on [line 520](https://github.com/haskell/haskell-mode/blob/a34ccdc54be15043ff0d253c3c20087524255491/haskell-cabal.el#L520) it is indeed creating the "lib:<package-name>" instances rather than picking whatever the right session name is. While I (think) I can read enough elisp to figure out where this went wrong, I'm not really fluent enough to easily change it to do the right thing. I guess we want to keep the current behavior for the unnamed global library, but use the proper section-name (=name of the sub-library) when the section has a name).

[1] the particular package in question is https://github.com/noinia/hgeometry/tree/hgeom-1/hgeometry-vector but I don't think it should matter.

noinia:
Fixes #1796.

### DIFF
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -515,7 +515,9 @@ PROCESS-TYPE determines the format of the returned target."
                        (val (car (split-string
                                   (haskell-cabal-section-value section)))))
                   (if (equal (downcase component-type) "library")
-                      (let* ((lib-name (if (eq 'val "") package-name val))
+                      (let* ((lib-name (if (not val)
+                                           (if (eq 'stack-ghci process-type) "lib" package-name)
+                                           val))
                              (lib-target (concat package-name ":" lib-name))
                              )
                         (push lib-target matches))
--- a/haskell-cabal.el
+++ b/haskell-cabal.el
@@ -515,11 +515,8 @@ PROCESS-TYPE determines the format of the returned target."
                        (val (car (split-string
                                   (haskell-cabal-section-value section)))))
                   (if (equal (downcase component-type) "library")
-                      (let* ((lib-name (if (not val)
-                                           (if (eq 'stack-ghci process-type) "lib" package-name)
-                                           val))
-                             (lib-target (concat package-name ":" lib-name))
-                             )
+                      (let* ((lib-name (or val (if (eq 'stack-ghci process-type) "lib" package-name)))
+                             (lib-target (concat package-name ":" lib-name)))
                         (push lib-target matches))
                     (push (concat  (when (eq 'stack-ghci process-type)
                                      (concat package-name ":"))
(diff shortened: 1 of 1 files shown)


haskell-cabal.el
```diff
@@ -515,9 +515,11 @@ PROCESS-TYPE determines the format of the returned target."
...
                                   (haskell-cabal-section-value section)))))
                   (if (equal (downcase component-type) "library")
-                      (let ((lib-target (if (eq 'stack-ghci process-type)
-                                            (concat package-name ":lib")
-                                          (concat "lib:" package-name))))
+                      (let* ((lib-name (if (not val)
+                                           (if (eq 'stack-ghci process-type) "lib" package-name)
+                                           val))
+                             (lib-target (concat package-name ":" lib-name))
+                             )
```
purcell, line 522:
```suggestion
                      (let* ((lib-name (or val (if (eq 'stack-ghci process-type) "lib" package-name)))
                             (lib-target (concat package-name ":" lib-name)))
```

purcell:
Tests now fail with what looks like a legitimate mismatch — worth taking a look and/or adding new tests to reflect the correct behaviour.

purcell:
Closing and re-opening to re-trigger CI and get a fresh picture of how this looks.

purcell:
```
Test haskell-cabal-enum-targets-1 condition:
    (ert-test-failed
     ((should
       (equal
	'("lib:Test" "test:test-1" "bench:bench-1" "exe:bin-1")
	(haskell-cabal-enum-targets)))
      :form
      (equal
       ("lib:Test" "test:test-1" "bench:bench-1" "exe:bin-1")
       ("Test:Test" "test:test-1" "bench:bench-1" "exe:bin-1"))
      :value nil :explanation
      (list-elt 0
		(arrays-of-different-length 8 9 "lib:Test" "Test:Test" first-mismatch-at 0))))
   FAILED   62/540  haskell-cabal-enum-targets-1 (0.000657 sec) at tests/haskell-cabal-tests.el:16
```

Is this correct? Should the expected list of targets now be changed?

PR #1805, opened 2023-04-27
Add new cabal stanzas to font-lock

jasagredo:
Sublibraries and program-options were not contemplated


jasagredo:
I'm not sure who to ping. I don't want to create noise but I also think this can be merged without probably any discussion. Sorry if I should not be pinging you @purcell

PR #1806, opened 2023-06-03
Inferior-haskell: do not insert repeated prompts.

cnngimenez:
Hi!

I found that ob-haskell.el shows multiple prompts at any code block output. This was a little annoying for me, so I tried to fix it. Then, I realized it was from the inferior-haskell: It did not remove the prompts from the process output. 
For example, when starting an inferior Haskell with `M-x run-haskell`  and writing a multiple line function, the "\*haskell\*" buffer looks like this:

```
Loaded package environment from /home/poo/.ghc/x86_64-linux-9.0.2/environments/default
GHCi, version 9.0.2: https://www.haskell.org/ghc/  :? for help
ghci> ghci> :{
f x
  | x == 0 = 1
  | x < 0  = 1
  | otherwise = 2 * f (x - 1)
:}

ghci| ghci| ghci| ghci| ghci| ghci> ghci> 
```

Therefore, I wrote a function to remove all prompts except for the last one. All continuation prompts (ending with "| ") are completely removed too. After that, I added it to the hook `comint-preoutput-filter-outputs` buffer-locally. Now, the result in the "\*haskell\*" buffer is:

```
Loaded package environment from /home/poo/.ghc/x86_64-linux-9.0.2/environments/default
GHCi, version 9.0.2: https://www.haskell.org/ghc/  :? for help
ghci> :{
f x
  | x == 0 = 1
  | x < 0  = 1
  | otherwise = 2 * f (x - 1)
:}

ghci> 
```

I believe that this output is easier to read, and the ob-haskell.el (and maybe other libraries) process the buffer output without the need for more code.

Is it good to include this modification? 
Does this patch requires any other change?
Cheers.

### DIFF
--- a/inf-haskell.el
+++ b/inf-haskell.el
@@ -113,6 +113,25 @@ The format should be the same as for `compilation-error-regexp-alist'.")
 correctly interpret multi-line input even when modules are
 imported.")
 
+(defconst inferior-haskell-cont-prompt-regexp
+  "^[[:alnum:].*_() |λ]*| "
+  "Continuation prompt rexep.
+Used to remove them from the output by the comint preoutput filter.  See
+`inferior-haskell-remove-extra-prompts'.
+
+This should be a similar regexp as `haskell-prompt-regexp', but it usually
+ends with \"| \" instead of \"> \".")
+
+(defconst inferior-haskell-maybe-cont-prompt-regexp
+  "^[[:alnum:].*_() |λ]*[>|] "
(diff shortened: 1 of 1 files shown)


purcell:
Looking back through a couple of PRs, since nobody actively maintains haskell-mode these days, and this seems worth merging — thanks!

purcell:
Going to revert this due to #1867, but if you want to fix it up and try again, let me know.

Issue #1470, opened 2017-01-10
haskell debugger not responding


zclod:
I can start the debugger process and set or remove breakpoints but I can't use it properly.

If i start the debug and hit a breakpoint the interface shows the debugger's commands but I can't interact because the context is not updated and I see only the message "Context"

if I type "a" to abandon the context and refresh the interface I get "Finished debugging" and it shows the trace of the execution.

(i'm using cabal + ghc8)

gracjan:
`haskell-debug` could use a lot of love. We need volunteers to tackle the issues there.

zclod:
have you successfully reproduced the bug? Because I'am using spacemacs and I don't know for sure if is a spacemacs issue or an haskell-mode issue.

Since the abandon and the continue commands are working and then the interface is refreshed I think that is an haskell-mode bug.

I would happily contribute but I'm pretty new to emacs and my elisp is very basic. Anyway, when I have some free time I will try to fix it.

PR #1810, opened 2023-07-11
fixed debugger for GHC 8/9

mobid:
I have tried ghc 8 and 9 on NixOS and both put extra line at the end of output about current break location, eg:
```
λ> :show breaks
[0] Main /home/me/path/Main.hs:(30,8)-(31,65) enabled
[/home/me/path/Main.hs:85:49-56] λ> 
```

Now debugger parse sections without errors. This bugfix is backward compatible.

This extra string is also shown with `haskell-doc-show-type` and break completion (when ghc is on break). Should we trim it in `haskell-process-queue-sync-request`?

#1470 could be related to this

### DIFF
--- a/haskell-debug.el
+++ b/haskell-debug.el
@@ -253,11 +253,12 @@
    (let* ((breakpoints (haskell-debug-get-breakpoints))
           (context (haskell-debug-get-context))
           (string
-           (haskell-process-queue-sync-request
-            (haskell-debug-process)
-            (if expr
-                (concat ":step " expr)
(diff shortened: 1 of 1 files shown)


purcell:
Merging this, thanks.

> This extra string is also shown with haskell-doc-show-type and break completion (when ghc is on break). Should we trim it in haskell-process-queue-sync-request?

Not sure, but that looks like the wrong place to be trimming such strings routinely. Better to do it close to where they are used IMO.

PR #1814, opened 2023-08-02
Fix test-redefined errors with Emacs >29

xgqt:
There were 2 duplicate test unless I missed some small difference.


purcell:
Thanks — I just added Emacs 29 to the CI matrix too.

Issue #1755, opened 2021-08-07
Distributing haskell-mode on NonGNU ELPA


phikal:
As mentioned in a [commit comment](https://github.com/haskell/haskell-mode/commit/17357211a16394884a981fe0c062f94250b40300), I am preparing a few packages for inclusion in NonGNU ELPA. Among these is haskell-mode. 

The ELPA build-system assumes a new package is released, whenever the version tag is bumped, but haskell-mode does not include a header, as it is distributed by MELPA, that adds a version tag before packaging.

Currently, the plan is to manually tag the last release, but to avoid having to manually update the version tag for each release, it would be helpful if a version tag could be added to `haskell-mode.el` itself, when the next release is being prepared.

phikal:
Just as an update, the package has been added to NonGNU, with a manually tagged commit for now: https://elpa.nongnu.org/nongnu/haskell-mode.html

purcell:
While I'm not averse to distributing this on NonGNU ELPA in due course, I'm honestly not super happy that it has happened without any any feedback or buy-in from the Haskell Mode maintainers.

phikal:
I understand your complaint, and if you wish it can be removed for the time being, or at least until others weigh in.

phikal:
Sorry for the delay, but the package has been removed now: https://git.savannah.gnu.org/cgit/emacs/nongnu.git/commit/?id=6f642491e2cd5007c06c3d1ca2c11315185459b8.

purcell:
Thanks, appreciated. Let's leave this issue open to record the wish.

phikal:
No problem! I hope that it will be possible to install `haskell-mode` OOTB some day, so if there are any issues I'd be glad to help.

skangas:
ping @purcell?

monnier:
Pi..ing?

monnier:
FWIW, I just re-added `haskell-mode` to the `elpa-packages` file in NonGNU ELPA:
While it was removed from that file, it was never actually removed from NonGNU ELPA itself (because of a missing feature in `elpa-admin.el` which made it unable to remove a package from the `archive-contents`), so this doesn't make much difference in practice.
But the problem in `elpa-admin.el` file is about to be fixed, and in the mean time the `hcel` package was added to GNU ELPA, and it uses `haskell-mode` to highlight its code, so we'd rather not remove `haskell-mode` from NonGNU ELPA any more.

purcell:
[...]

PR #1815, opened 2023-08-03
minor fixes made while trying to get debian package working with emacs 29.1

barak:
These are some minor fixes, like stray version numbers and some spelling errors and such. Plus one more serious change, to make it easier to debug lexical test cases by actually returning a list of tokens. This also avoids some ugly code duplication.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@
 # We should have a script that changes it everywhere it is needed and
 # syncs it with current git tag.
 #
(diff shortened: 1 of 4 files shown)


purcell:
The previous versions were like this to indicate that the master version of the code was approaching 17.2, so the new equivalent would be 17.4-git.

But after consideration, I'll just merge this as-is, as it relates to #1755. Thanks!

PR #1829, opened 2023-10-24
Implement indentation support for TeX-style literate Haskell buffers

DataKinds:
This PR implements a new function `haskell-indentation-literate-outside-code-p` which may be used to detect when the point is outside a code block in both Bird-style and TeX-style literate buffers. Note that GHC accepts a mixture of both Bird-style and TeX-style delimiting in the same file, so changing the possible values of the `haskell-literate` variable might be in-scope here as well.

### DIFF
--- a/haskell-indentation.el
+++ b/haskell-indentation.el
@@ -126,6 +126,14 @@ set and deleted as if they were real tabs."
   "Return t if this is a literate Haskell buffer in bird style, NIL otherwise."
   (eq haskell-literate 'bird))
 
+(defun haskell-indentation-tex-p ()
+  "Return t if this is a literate Haskell buffer in tex style, NIL otherwise."
+  (eq haskell-literate 'tex))
+
+(defun haskell-indentation-literate-p ()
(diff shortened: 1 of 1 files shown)


purcell:
Thanks!

DataKinds:
> Hey @DataKinds, @smonnier [suggests](https://lists.gnu.org/archive/html/emacs-devel/2023-10/msg00612.html) a simpler formulation of the new function:
> 
> ```emacs-lisp
> (defun haskell-indentation-tex-outside-code-p ()
>   "Non-NIL if we are in tex literate mode, but outside of code."
>   (and (haskell-indentation-tex-p)
>        (not (and (save-excursion
>                    (re-search-backward "\\\\end{code}\\|\\\\begin{code}\\(\\)"
>                                        nil t))
>                  (match-end 1)))))
> ```
> 
> Would you mind testing it out to see if it works?

Tested both in and out of `evil-mode` on the same file I wrote my PR with. As far as I can tell the simpler formulation works well. The empty group at the end of the regex is a cool trick.

PR #1835, opened 2024-01-10
Fix #1834: remove bad regex that was supposed to recognize a LANGUAGE suggestion

### LINKED ISSUE #1834: Add {-# LANGUAGE Data #-} to the top of the file? (y or n)
When I load a hs file without project (cabal or stack) I get asked
> Add {-# LANGUAGE Data #-} to the top of the file? (y or n) 

This question can be repeated several times in a row, regardless whether I answer y or n.
Each time I answer y the `{-# LANGUAGE Data #-}` line will be added (which does not make any sense since LANGUAGE pragmas are idempotent).

The `Data` extension isn't even listed here: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/table.html
I don't know what purpose it should serve.

I just upgraded to the latest version of `haskell-mode` (v20231115) but the problem persists.

andreasabel:
Remove bad regex that was supposed to recognize a LANGUAGE suggestion.

This regex fires unfortunately on GHC 9.8's new `x-partial` warning,
destroying the UX with obnoxious questions whether to "add LANGUAGE Data".

Closes #1834.

### DIFF
--- a/haskell-load.el
+++ b/haskell-load.el
@@ -205,7 +205,13 @@ list of modules where missed IDENT was found."
             (string-match
              "Use \\([A-Z][A-Za-z]+\\) if you want to disable this"
              msg)
-            (string-match "use \\([A-Z][A-Za-z]+\\)" msg)
+            ;; Andreas Abel, 2024-01-10, https://github.com/haskell/haskell-mode/issues/1834
+            ;; The following regex also matches the new x-partial warning
+            ;; "Consider refactoring to use Data.List.NonEmpty"
+            ;; which would result in a suggestion to
+            ;; "Add {-# LANGUAGE Data #-} to the top of the file?"
+            ;; Thus, we disable it.
+            ;; (string-match "use \\([A-Z][A-Za-z]+\\)" msg)
             (string-match "You need \\([A-Z][A-Za-z]+\\)" msg)))
          (when haskell-process-suggest-language-pragmas
            (haskell-process-suggest-pragma
--- a/haskell-load.el
+++ b/haskell-load.el
@@ -205,13 +205,8 @@ list of modules where missed IDENT was found."
             (string-match
              "Use \\([A-Z][A-Za-z]+\\) if you want to disable this"
              msg)
-            ;; Andreas Abel, 2024-01-10, https://github.com/haskell/haskell-mode/issues/1834
-            ;; The following regex also matches the new x-partial warning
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. Better to remove this code completely: the comment in the commit message is sufficient. Even better would be to extend the existing message so that it matches its original targets but not the new "refactor" message. The tricky thing is finding examples of the former!

haskell-load.el
```diff
@@ -205,7 +205,13 @@ list of modules where missed IDENT was found."
...
              "Use \\([A-Z][A-Za-z]+\\) if you want to disable this"
              msg)
-            (string-match "use \\([A-Z][A-Za-z]+\\)" msg)
+            ;; Andreas Abel, 2024-01-10, https://github.com/haskell/haskell-mode/issues/1834
+            ;; The following regex also matches the new x-partial warning
+            ;; "Consider refactoring to use Data.List.NonEmpty"
+            ;; which would result in a suggestion to
+            ;; "Add {-# LANGUAGE Data #-} to the top of the file?"
+            ;; Thus, we disable it.
+            ;; (string-match "use \\([A-Z][A-Za-z]+\\)" msg)
```
purcell, line 214: Or, following the example a few lines above, how about:
```suggestion
            (and (string-match "use \\([A-Z][A-Za-z]+\\)" msg)
                 (not (string-match "refactoring to use" msg)))
```

Bodigrim:
(Just passing by) Is it difficult for `haskell-mode` to filter anything remotely suspicious to be a language extension through `ghc --supported-extensions`? This would be a more robust solution than adjusting regexps here and there.

andreasabel:
@purcell : Thanks for the quick review.
I will leave the decision to you how to fix the problem.
@Bodigrim 's advice is good; the current heuristics is just too brittle.

purcell:
Sounds reasonable. I think there may even be a list of supported extensions in the code which gets updated periodically — using that would be better than a fresh process invocation if so. I'm unlikely to work on this myself though, sorry.

andreasabel:
@purcell : You may at least merge my hotfix...

purcell:
> You may at least merge my hotfix...

Thanks! 😉

Alright, have applied my suggestion above and will merge once CI passes. Might have to close and re-open this PR briefly to trigger a rebase on top of changes I just made in `master` to fix CI.

andreasabel:
Thanks Steve @purcell!

PR #1842, opened 2024-04-28
Make dabbrev-expand detect underbar and prime as part of identifiers

kenta2:
The regex syntax `\s_` is documented at <https://www.gnu.org/software/emacs/manual/html_node/emacs/Regexp-Backslash.html>, and the syntax table entries for prime and underbar (underscore) are already set in `haskell-mode-syntax-table`.


purcell:
Looking back through a couple of PRs, since nobody actively maintains haskell-mode these days, and this seems worth merging — thanks!

PR #1846, opened 2024-05-20
Add GHC 9.10.1 extensions and options

amesgen:
GHC 9.10.1 has been [released](https://downloads.haskell.org/ghc/9.10.1/docs/users_guide/9.10.1-notes.html)


purcell:
Nice, thanks!

Issue #1642, opened 2019-04-05
haskell-process-load-file on a file with a relative import and a type error fails with "Unexpected response from haskell process" instead of showing the error


rntz:
**How to replicate:** Make a directory with two files, `A.hs` and `B.hs`, like so:
```
-- In A.hs:
module A where {}
-- In B.hs:
import A
x = unbound -- or any other code that shouldn't typecheck
```

Then visit `B.hs` and run `C-c C-l`, that is, `haskell-process-load-file`.

**Expected result:** I get a nice error message in `*haskell*` telling me `Variable not in scope: unbound`.

**Actual result:** I get a message in the minibuffer saying: `Haskell process command errored with: (error "Unexpected response from haskell process.")`, and no error message in `*haskell*`, so I am left not knowing what went wrong or how to fix it.

rntz:
Revisiting this after 5 years, this particular bug appears to be fixed for me. However, other very similar cases generate the same problem. For instance:

```haskell
import Data.Map
x = unbound
```

This may be the root cause of https://github.com/haskell/haskell-mode/issues/1621 as well.

rntz:
this may be fixed by https://github.com/haskell/haskell-mode/pull/1850

PR #1850, opened 2024-09-11
haskell-process-load-complete: handle ghc 9.10 message

rntz:
ghc 9.10 and later can respond to `:load` with "Failed, unloaded all modules." Unfortunately `haskell-process-load-complete` does not recognize this response. This PR fixes this. To simplify and future-proof the code I've adjusted `haskell-process-load-complete` to look for the regexes `Ok, .*$` or `Failed, .*$`. Alternatively, if you'd prefer to narrowly add the "Failed, unloaded all modules." message, [you could cherry-pick this commit.](https://github.com/haskell/haskell-mode/commit/337c41e146e47c5f92b28009c535fe9e72cfad76)

This may possibly fix https://github.com/haskell/haskell-mode/issues/1642.

### DIFF
--- a/haskell-load.el
+++ b/haskell-load.el
@@ -134,6 +134,10 @@ actual Emacs buffer of the module being loaded."
                 process
                 "Failed, no modules loaded\\.$") ;; for ghc 8.4
                nil)
+              ((haskell-process-consume
+                process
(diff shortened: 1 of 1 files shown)


purcell:
Looks good, thanks — merged!

PR #1851, opened 2024-09-11
ci: Test only the latest Emacs minor version

jcs090218:
It's generally enough to only test the latest minor version. 🤔 So we can lower the CI time.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,18 +14,14 @@ jobs:
     strategy:
       matrix:
         emacs_version:
-          - 25.1
           - 25.3
-          - 26.1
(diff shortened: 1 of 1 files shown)


purcell:
Why?

jcs090218:
I've added the reason in OP. :)

purcell:
Okay, so here's why it was the way it was:

- Firstly, it's always desirable to test the oldest compatible version
- Secondly, I like to test the oldest and newest minor release for each major version, but am happy to skip the ones in-between

So I have a couple of updates to make here, but I won't make exactly the changes proposed. Thanks for the poke though!

PR #1857, opened 2024-11-07
Optimize and simplify `haskell-ds-create-imenu-index` with hashtable

Hi-Angel:
The function has implemented a poor man's hash table by manually enlisting keys and then using `cl-case` on values. Just replace this with a hashtable and thus simplify the code considerably. This makes it faster as well because the `cl-case + setq` is O(n) whereas `puthash` is O(1).

The output of the function was tested on Pandoc's
`Writers/Powerpoint/Output.hs` file of 2841 lines and comparing the hashsum of the output before the changes and after. It didn't change.

The commit is also basically a backport of a similar change on purescript-mode that has been forked form haskell-mode at some point: https://github.com/purescript-emacs/purescript-mode/pull/27

### DIFF
--- a/haskell-decl-scan.el
+++ b/haskell-decl-scan.el
@@ -106,6 +106,7 @@
 (require 'haskell-mode)
 (require 'syntax)
 (require 'imenu)
+(require 'subr-x)
 
 (defgroup haskell-decl-scan nil
   "Haskell declaration scanning (`imenu' support)."
@@ -537,11 +538,7 @@ datatypes) in a Haskell file for the `imenu' package."
   ;; These lists are nested using `(INDEX-TITLE . INDEX-ALIST)'.
   (let* ((bird-literate (haskell-ds-bird-p))
          (index-alist '())
-         (index-class-alist '()) ;; Classes
-         (index-var-alist '())   ;; Variables
-         (index-imp-alist '())   ;; Imports
-         (index-inst-alist '())  ;; Instances
-         (index-type-alist '())  ;; Datatypes
+         (imenu (make-hash-table :test 'equal))
          ;; The result we wish to return.
          result)
     (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


Hi-Angel:
Just in case: CC: @cocreature as the last person who touched the function

purcell:
Thanks for this. I approved the CI and it's complaining about part of the change, not immediately obvious to me why. :)

purcell:
It's a warning that is promoted to an error, but the cause is `when-let*` not necessarily being defined. Adding `(require 'subr-x)` fixes it for me, at least in Emacs 28.x.

Hi-Angel:
Oh, thank you! I was comparing the `require` lines on the purescript and haskell "decl-scan" files, but they are the same, so I dismissed that thought. But perhaps in "purescript-mode" the `(require 'subr-x)` line comes from another import *(both haskell and purescript "decl-scan" files include the main mode file)*.

I added the line, let's see if CI will be okay with it

Hi-Angel:
Welp… 😊 I'll dig into it later, don't bother for now, thank you!

purcell:
> But perhaps in "purescript-mode" the (require 'subr-x) line comes from another import (both haskell and purescript "decl-scan" files include the main mode file).

Yeah, I don't really like that pattern of importing, partly for this reason!

The error with Emacs 25.1 is because `when-let*` was only added to `subr-x` in Emacs 26.1. Probably best to avoid that macro — `when-let` was available in Emacs 25.1, so you can just use that instead.

Hi-Angel:
> The error with Emacs 25.1 is because `when-let*` was only added to `subr-x` in Emacs 26.1. Probably best to avoid that macro — `when-let` was available in Emacs 25.1, so you can just use that instead.

Well, `when-let*` compiles just fine on 25.1 [with the same change on purescript-mode](https://github.com/purescript-emacs/purescript-mode/pull/27)

purcell:
Huh! 🤔

Hi-Angel:
> Well, `when-let*` compiles just fine on 25.1 [with the same change on purescript-mode](https://github.com/purescript-emacs/purescript-mode/pull/27)

Oh, actually, scratch that. I just looked at the job log, and apparently the purescript-mode CI, for some reason, doesn't compile that specific file at all. It compiles other files, but not this one. So most likely this is the reason I don't see the failure there.

purcell:
> The reason I attempted `when-let*` is because `when-let` was declared obsolete in preference of `when-let*` in current Emacs master…

Yeah, always tricky juggling these things when aiming for compatibility with a wide range of Emacsen. It's possible that the `compat` package helps here, but I'd rather not add it as a dependency just for this case.

Merging now, thanks!

PR #1861, opened 2025-02-02
ensure run-expr uses *haskell* buffer

zaz:
Previously, it would run `(goto-char (point-max)) (insert "\n")` in whatever buffer you are currently in.

This was [causing issues for some people](https://stackoverflow.com/a/71212327/405550).

### DIFF
--- a/haskell-repl.el
+++ b/haskell-repl.el
@@ -70,34 +70,35 @@
   "Run the given expression."
   (let ((session (haskell-interactive-session))
         (process (haskell-interactive-process)))
-    (haskell-process-queue-command
-     process
-     (make-haskell-command
-      :state (list session process expr 0)
-      :go (lambda (state)
-            (goto-char (point-max))
-            (insert "\n")
-            (setq haskell-interactive-mode-result-end
-                  (point-max))
-            (haskell-process-send-string (cadr state)
-                                         (haskell-interactive-mode-multi-line (cl-caddr state)))
-            (haskell-process-set-evaluating (cadr state) t))
-      :live (lambda (state buffer)
-              (unless (and (string-prefix-p ":q" (cl-caddr state))
-                           (string-prefix-p (cl-caddr state) ":quit"))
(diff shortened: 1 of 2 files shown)


purcell:
Thanks, I cherry-picked and applied the main commit, but will skip the other one. Thanks!

zaz:
@purcell Thank you! Do you have time to explain why? Was there something wrong with the code?

As you can see from [StackOverflow](https://stackoverflow.com/q/13213717/405550), this feature is desired. Is there a better way to achieve this goal?

purcell:
> Do you have time to explain why? Was there something wrong with the code?
>

Primarily, I wanted to avoid conflating unrelated changes in one PR.

But my reservation is that `main` is only an obvious entry point in certain files, while in others it might be `test`. `main` might also expect command line args. So I don't think this is enough to be very widely useful, though I can imagine it might be a handy shortcut for some people, who could define such a command locally.

zaz:
@purcell Thanks for explaining! Would you accept a PR for a function that prompts the user for which function to run and command line args and then remembers that for future invocations?

Or is there a better way to achieve the same thing? How are most people running their Haskell files from within Emacs?

purcell:
Yeah, that could make sense. Maybe just add a function name to the signature of your `haskell-load-and-run` command:

You can do something like this to remember the last select expr, so that an extra `RET` will let you quickly re-run the last expression sent:

```el
(defvar haskell-load-and-run-expr-history nil)

(defun haskell-load-and-run (expr)
  "Load the current buffer and run EXPR, e.g. \"main\"."
  (interactive (list (read-string "Run function: "
                                  (car haskell-load-and-run-function-history)
                                  'haskell-load-and-run-function-history)))
  (haskell-process-load-file)
  (haskell-interactive-mode-run-expr expr))
```

Perhaps give that a test-drive and see how it is for you in practice?

purcell:
(Edited the above on Github to fix typos, just in case you're relying on email)

PR #1863, opened 2025-02-08
Load and run expr

zaz:
_following from PR #1861_ 

Thanks a lot! This works well for me.

In the future it could be [modified to use completing-read](https://github.com/zaz/haskell-mode/compare/load-and-run-expr..completing-read?expand=1) or similar, but I find that pops up a buffer that is annoyingly large and what you wrote already works great.


purcell:
Nice, thanks!

Issue #1826, opened 2023-10-01
haskell-cabal: bad indentation with redundant leading comma


Commelina:
As [the Cabal document](https://cabal.readthedocs.io/en/latest/file-format-changelog.html#cabal-version-2-2) says, Cabal allows redundant leading (and trailing) commas since 2.2. And some tools such as [cabal-fmt](https://github.com/phadej/cabal-fmt) use leading commas by default. However, the indentation of such leading commas does not work very well. For example,

```
build-depends:
        , bar
, foo
```

When indenting the last line, we may expect 

```
build-depends:
        , bar
        , foo
```

However, `haskell-cabal-indent-line` gives

```
build-depends:
        , bar
      , foo
```

This is a little annoying. I am not sure if this is caused by my incorrect usage. However, I tried [checking](https://github.com/Commelina/haskell-mode/commit/0f398ef3dd249a96db01ef991d695647afe2d8be) if there was a leading comma before indenting and taking different indent strategies, which worked as expected. (Of course, there are still corner cases such as multiple blank lines. But the origin code does not handle them very well either...) I will make a PR if it is OK.

Commelina:
I attempted to fix the indentation for subsections with leading commas or multiple newlines. It should work as expected on the cases:

```
  build-depends:
         , base
         , text
```

```
  build-depends:


       base
     , text 
```

Do you have any idea? @andreasabel

andreasabel:
This looks good.

> Do you have any idea? [@andreasabel](https://github.com/andreasabel)

What kind of input do you want from me?  (Sorry, I fail to comprehend your question.)

Commelina:
Uh, I was just saying if the patch works for you. Sorry for my ambiguity!

andreasabel:
I am currently using the `haskell-mode` from MELPA so I cannot easily test a PR.

PR #1869, opened 2025-02-26
haskell-cabal: fix indentation with leading commas and newlines

Commelina:
Fixes [#1826](https://github.com/haskell/haskell-mode/issues/1826).


purcell:
Thanks, and sorry for the delay. I think this is safe enough to merge and we'll see if any issues emerge.

Issue #1743, opened 2021-04-18
toggling a comment does not work in Haskell-Cabal mode


FabArd:
When I want to uncomment a line in the Haskell-Cabal mode it doesn't work if the line starts with a space.

Exemple : if I want to uncomment the above line with C-x C-;
  -- other-modules:
will result in 
  -- -- other-modules:

It seems that uncommenting a line that starts with a space is impossible.
No problem when the line do not start with space : toggling is correct.

bernhard-herzog:
Is this still a problem? It seems to work fine for me in a current version haskell-mode.

Hi-Angel:
@bernhard-herzog it's because it works when the comment is at the beginning of a line, but not when there's whitespace before it. For details see PR I just linked to the issue.

PR #1870, opened 2025-03-04
haskell-cabal: fix (comment-dwim) by correcting comment-start-skip

Hi-Angel:
Prior to this commit using (comment-dwim) over a comment in .cabal file that doesn't start at the beginning of a line resulted in it adding even more comments. E.g. this:

    -- comment not at the beginning of a line

resulted in:

    -- -- comment not at the beginning of a line

This is because `comment-start-skip` was including whitespace that isn't part of the comment syntax. Fix this by only limiting the regexp to the comment part.

The resulting regexp is basically same as the one in haskell-mode, barring that that one also takes into account `{-` comment starter, which isn't a thing in .cabal files.

Fixes: https://github.com/haskell/haskell-mode/issues/1743


Hi-Angel:
Just realized there was a bugreport for this, added to "fixes".

purcell:
Nice, thanks.

PR #1872, opened 2025-03-04
Fix incorrect face usage and undeclared list elements type

Hi-Angel:
This fixes a few warnings. Couldn't fix more because it complains `if-let` being an obsolete macro and that `if-let*` should be preferred, but I'm not sure what's the best course of action here. `if-let*` appeared since 26.1, so it's either upping the minimal version 26.1 or adding a wrapper for older Emacses. Emacs 26.1 was released in 2018.


purcell:
Yes, `if-let` has had a complicated history. An option is requiring the `compat` package, but I'd prefer to avoid pulling in external dependencies for something trivial like this. Having a separate `let` and `if`/`when` is perfectly fine, so rewriting to that might be the best option.

PR #1874, opened 2025-03-11
Let Projectile choose haskell-build-type if possible

webdevred:
Overview:
This patch enhances Haskell Mode by integrating Projectile to determine whether Stack, Cabal, or system GHC should be used for a project. Previously, GHC selection relied on auto-detection, which could mistakenly pick up unintended project files (e.g., a stray stack.yaml in the home directory). This change gives users explicit control over the selection based on their project's context.

What This Patch Does:

- Projectile Integration: Uses Projectile to determine the project root and select the appropriate GHC environment.
- Avoids Misconfiguration: Prevents Haskell Mode from mistakenly using the wrong compiler due to stray configuration files outside the intended project.

Benefits:

- More Predictable Behavior: Ensures the correct compiler is used based on project settings.
- Better Workflow with Projectile: Works seamlessly with existing Emacs project management tools.
- Prevents Confusion: Eliminates cases where users might unknowingly use an unintended Stack or Cabal project.

Testing:

- Verified that the correct GHC environment is selected based on the project type.
- Ensured that fallback behavior works as expected when no specific configuration is found.

Conclusion:
This update makes Haskell Mode more robust and user-friendly by giving users direct control over how GHC is selected per project. It prevents unintended behavior while integrating well with the existing Emacs tooling.

Let me know if any refinements are needed!

### DIFF
--- a/haskell-customize.el
+++ b/haskell-customize.el
@@ -410,6 +410,24 @@ imports."
   "The path which is considered as project root, this is determined by the
 presence of a *.cabal file or stack.yaml file or something similar.")
 
+(defun haskell-projectile-build-type ()
+  (if (and (fboundp 'projectile-project-root) (bound-and-true-p projectile-mode))
+      (let* ((d (projectile-project-root))
+            (cabal (cl-find-if
+                    (lambda (f) (string-match-p ".\\.cabal\\'" f))
+                    (directory-files d))))
+        (cond
+         ((and (file-exists-p (concat d "cabal.project")) (executable-find "cabal"))
+          (cons 'cabal-project d))
+         ((and (file-exists-p (concat d "cabal.sandbox")) (executable-find "cabal"))
+          (cons 'cabal-sandbox d))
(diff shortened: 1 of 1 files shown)


purcell:
I don't think `haskell-mode` should be deferring to `projectile` to find the project root:

- It introduces quite different paths by which Haskell build roots might be found, according to what's installed and enabled
- There's `project.el` these days as standard anyway, and `projectile` can use that under the hood

It's not obvious to me that `haskell-build-type` can be reconciled with the idea of a single project root anyway. You can have a project repo with sub-components that each build differently, potentially. What we're talking about here is finding how to compile from the closest root for the current buffer, which often wouldn't be the project root. Rather than trying to drive everything magically off a single project root returned by an external library returned by projectile, wouldn't we really want a way to more finely control the directory used for `haskell-compile`, which is what `haskell-build-type` is predominantly used for?

(Certainly the existing `haskell-compiler-type` var and `haskell-build-type` machinery is a bit clunky.)

PR #1880, opened 2025-09-30
Correct Cabal comment syntax highlighting

sellout:
Previously it highlighted the entire line as a comment. This fixes it to ignore the leading whitespace, highlight the `--` as a comment delimiter, then highlight the remainder as the actual comment.


webdevred:
LGTM!

PR #1885, opened 2026-02-05
Do not redefine xref-prompt-for-identifier

bcc32:
The correct way to declare a variable as dynamically bound but defined elsewhere is to use defvar with a variable name but no value.  Otherwise, this causes the following unwanted side-effect when pressing `C-h v xref-prompt-for-identifier`:

```
xref-prompt-for-identifier is a variable defined in ‘haskell.el’.
```


purcell:
Yup, thanks!

PR #1893, opened 2026-08-19
Support GHC 9.12 MultilineStrings

### LINKED ISSUE #1877: Add support for MultilineStrings extension
GHC 9.12 introduced multiline string literals feature.
Triple double quote denotes a String.
After such token indentation breaks a bit.

jmcarthur:
Fixes https://github.com/haskell/haskell-mode/issues/1877.

This fixes the syntax highlighting and the indentation. Recognition is not conditional on the extension being used, the same way quasi-quotes are treated.

When it recognizes either `"""` or `"`, `haskell-lexeme-looking-at-string-literal` dispatches into a general scanner, `haskell-lexeme--scan-string-literal`, parameterized by the number of quotes (1 or 3) and by whether an unescaped newline should terminate the literal. `haskell-syntax-propertize` marks the extra delimiter quotes and any quotes inside the literal as punctuation, so the syntax parser sees one ordinary string spanning lines.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -8,6 +8,8 @@ This file uses Org mode. Some useful (default) key-bindings:
 
 * Changes in 17.5
 
+- Support multiline string literals
+
 - See [[https://github.com/haskell/haskell-mode/compare/v17.4...v17.5][Git history]]
 
 * Changes in 17.4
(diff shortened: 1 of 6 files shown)


jmcarthur:
I just realized I didn't include any tests to support the indentation claim. I'll fix that shortly.

webdevred:
Looks good.

purcell:
Thanks, merged! 🎉

